PR #17, opened 2019-08-05
Add support for default branch and build skipping

juliangrosshauser:
Thank you for your work on this provider plugin, @jradtilbrook! Really appreciate it.
Made my job a lot easier 🙂

I noticed that setting the default branch wasn't supported yet, so I implemented that. Empty default branch means all branches, so I tested `default_branch = ""` (which is also the default) and that works without problems.

Build skipping fields were recently added to the GraphQL API (https://forum.buildkite.community/t/feature-parity-for-pipeline-creation-between-graphql-and-rest-apis/600/3), so I implemented them too.

### DIFF
--- a/buildkite/resource_pipeline.go
+++ b/buildkite/resource_pipeline.go
@@ -12,23 +12,23 @@ import (
 
 // PipelineNode represents a pipeline as returned from the GraphQL API
 type PipelineNode struct {
-	DefaultBranch graphql.String
-	Description   graphql.String
-	Id            graphql.String
-	Name          graphql.String
-	Repository    struct {
-		Url graphql.String
-	}
-	Slug  graphql.String
-	Steps struct {
-		Yaml graphql.String
-	}
-	Uuid                                 graphql.String
-	WebhookURL                           graphql.String `graphql:"webhookURL"`
-	SkipIntermediateBuilds               graphql.Boolean
-	SkipIntermediateBuildsBranchFilter   graphql.String
 	CancelIntermediateBuilds             graphql.Boolean
 	CancelIntermediateBuildsBranchFilter graphql.String
+	DefaultBranch                        graphql.String
+	Description                          graphql.String
+	Id                                   graphql.String
+	Name                                 graphql.String
+	Repository                           struct {
+		Url graphql.String
+	}
+	SkipIntermediateBuilds             graphql.Boolean
+	SkipIntermediateBuildsBranchFilter graphql.String
+	Slug                               graphql.String
+	Steps                              struct {
+		Yaml graphql.String
+	}
+	Uuid       graphql.String
+	WebhookURL graphql.String `graphql:"webhookURL"`
 }
 
 func resourcePipeline() *schema.Resource {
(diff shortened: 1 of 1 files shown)


juliangrosshauser:
Travis fails because `go fmt $(go list ./...)` fails for `go1.10`. Formatting is different for versions > 1.10. Does it make sense to test version 1.10?

jradtilbrook:
Awesome, thank you for contributing! :tada:  
All those additions sound great, as you might be able to tell, I've just been adding to them as I personally need them.

I think you are right; I'll update the travis config to remove 1.10 and then you can rebase onto that. I'll let you know when it's ready

jradtilbrook:
Alright that's ready to go now

buildkite/resource_pipeline.go
```diff
@@ -23,8 +23,12 @@ type PipelineNode struct {
...
 		Yaml graphql.String
 	}
-	Uuid       graphql.String
-	WebhookURL graphql.String `graphql:"webhookURL"`
+	Uuid                                 graphql.String
+	WebhookURL                           graphql.String `graphql:"webhookURL"`
+	SkipIntermediateBuilds               graphql.Boolean
+	SkipIntermediateBuildsBranchFilter   graphql.String
+	CancelIntermediateBuilds             graphql.Boolean
+	CancelIntermediateBuildsBranchFilter graphql.String
```
jradtilbrook: Can you sort all the parameters alphabetically

buildkite/resource_pipeline.go
```diff
@@ -62,6 +66,26 @@ func resourcePipeline() *schema.Resource {
...
+				Type:     schema.TypeString,
+			},
+			"cancel_intermediate_builds": &schema.Schema{
+				Optional: true,
+				Type:     schema.TypeBool,
+			},
+			"cancel_intermediate_builds_branch_filter": &schema.Schema{
+				Optional: true,
+				Type:     schema.TypeString,
+			},
```
jradtilbrook: Can you also sort these ones too?

buildkite/resource_pipeline.go
```diff
@@ -77,15 +101,20 @@ func CreatePipeline(d *schema.ResourceData, m interface{}) error {
 	var mutation struct {
 		PipelineCreate struct {
 			Pipeline PipelineNode
-		} `graphql:"pipelineCreate(input: {organizationId: $org, name: $name, description: $desc, repository: {url: $repository_url}, steps: {yaml: $steps}})"`
+		} `graphql:"pipelineCreate(input: {organizationId: $org, name: $name, description: $desc, repository: {url: $repository_url}, steps: {yaml: $steps}, defaultBranch: $default_branch, skipIntermediateBuilds: $skip_intermediate_builds, skipIntermediateBuildsBranchFilter: $skip_intermediate_builds_branch_filter, cancelIntermediateBuilds: $cancel_intermediate_builds, cancelIntermediateBuildsBranchFilter: $cancel_intermediate_builds_branch_filter})"`
```
jradtilbrook: Do you think this line should be split? I was in two minds before this change, but now its much longer!

juliangrosshauser: I also wanted to split it, but that's unfortunately not possible 😔 https://github.com/shurcooL/graphql/issues/21

jradtilbrook: Ah damn!

juliangrosshauser:
@jradtilbrook I also sorted the GraphQL arguments in the tags alphabetically.

jradtilbrook:
Awesome! Nice work and thanks again for contributing. I'd like to get this listed on the official Terraform 3rd party providers page at some point so a star would be appreciated if you're up for it.  
I'll cut a new minor release as well

juliangrosshauser:
> I'd like to get this listed on the official Terraform 3rd party providers page at some point so a star would be appreciated if you're up for it.

Oh, that would be awesome 👏 Starred the repo now 🌟

I'll definitely keep working on this provider. Pipeline schedules are the next thing I want to tackle, I think.

jradtilbrook:
That would be excellent. I've already created 2 issues for the next things I'm going to focus on; adding tests and user related things like SSO/teams/etc.

Your additions will be gladly welcomed

PR #51, opened 2020-07-09
Make pipeline teams optional

### LINKED ISSUE #48: "team": required field is not set
This looks like a great update of the terraform provider!

But I noticed team support was added, and is now required. I get:

```
Error: "team": required field is not set

  on ops.tf line 14, in resource "buildkite_pipeline" "ops":
  14: resource "buildkite_pipeline" "ops" {

```

But teams is an optional feature, and not required. Is there a reason it's required by the terraform resources?

darrylb-github:
As issue #48 points out, teams are only a thing on paid plans. This makes the team block optional and updates the readme including a warning about non-admin users creating a pipeline without a team. I'd imagine this doesn't error on a free plan but does error on a paid plan.

Resolves #48

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,7 +129,7 @@ resource "buildkite_pipeline" "repo2" {
 - `description` - (Optional) A description of the pipeline.
 - `repository` - (Required) The git URL of the repository.
 - `steps` - (Required) The string YAML steps to run the pipeline.
-- `team` - (Optional) Set team access for the pipeline. Can be specified multiple times for each team. Note that non-admin users may receive a buildkite permission error if trying to create a pipeline without at least one team. 
+- `team` - (Optional) Set team access for the pipeline. Can be specified multiple times for each team.  
 
 The `team` block supports:
 
--- a/README.md
+++ b/README.md
@@ -115,6 +115,11 @@ resource "buildkite_pipeline" "repo2" {
     name = "repo2"
     repository = "git@github.com:org/repo2"
     steps = file("./steps.yml")
+
+    team {
+        slug = "everyone"
+        access_level = "READ_ONLY"
+    }
 }
 ```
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -124,6 +129,12 @@ resource "buildkite_pipeline" "repo2" {
 - `description` - (Optional) A description of the pipeline.
 - `repository` - (Required) The git URL of the repository.
 - `steps` - (Required) The string YAML steps to run the pipeline.
+- `team` - (Optional) Set team access for the pipeline. Can be specified multiple times for each team. Note that non-admin users may receive a buildkite permission error if trying to create a pipeline without at least one team.
```
sj26, line 132: To clarify how this works on the backend:

- If teams is not enabled on the organization (default for all new organizations)
  - There cannot be teams
- If teams is enabled on the organization (via https://buildkite.com/organizations/-/teams)
  - And the user is an admin
    - No teams are required
  - And the user is not an admin
    - And the user is only in a single team
      - No teams are required — it will default to the single team (usually the default "Everyone" team)
    - And the user is in multiple teams
      - Teams are required — otherwise there is ambiguity

sj26: The simplest approach for the terraform provider is probably to make teams optional.

There is a small potential write/read inconsistency, though — if teams is enabled, the user is not an admin, and they are in a single team, and they do not provide teams, then the provider might read the resource after creation and see teams, and thus a diff.

A more complex approach would be to read the organization to see if teams is enabled and then dynamically decide if teams are required or not. I don't think we currently expose whether teams is enabled or not — I'll see if I can change that.

sj26:
Here's the teams page for a new organization:

image

It's an optional feature because it subtly changes how permissions work and we didn't want to change that for all historical organizations. :-)

README.md
darrylb-github: > There is a small potential write/read inconsistency, though — if teams is enabled, the user is not an admin, and they are in a single team, and they do not provide teams, then the provider might read the resource after creation and see teams, and thus a diff.

In that scenario I don't think it would create the pipeline at all and would return`Error: Please ensure at least one team is added to this pipeline`. That was my experience anyway. And thinking more about it, that error is perfectly clear in this scenario and fixable now by the user as they can specify a team, so I don't think I need the disclaimer bit anymore.

sj26:
FYI I've added a property to organizations via graphql which indicated whether teams is enabled or not:

```graphql
query {
  organization(slug: "...") {
    isTeamsEnabled # true/false
  }
}
```

I'm not sure it makes sense to use here, but it's available nonetheless.

Issue #53, opened 2020-07-29
Q: How to set branch filtering?


runlevel5:
I could not find any reference to branch filter in the source code. Wondering if it is possible to set branch filter? cc @jradtilbrook 

![Screenshot_2020-07-29 Settings · admin-ui](https://user-images.githubusercontent.com/135605/88753999-ec70a800-d1a0-11ea-9658-8db48abef1a8.png)

jradtilbrook:
Hi @runlevel5, this provider doesn't currently support that setting. I'm happy to review and accept a PR though if you wanted to add it!

runlevel5:
@jradtilbrook I am new to Golang ..however I could try if you are okay with giving me feedback on my PR.

jradtilbrook:
> @jradtilbrook I am new to Golang ..however I could try if you are okay with giving me feedback on my PR.

Absolutely! If you're willing to give it a go I'll give as much feedback as I can :+1:

runlevel5:
@jradtilbrook https://github.com/jradtilbrook/terraform-provider-buildkite/pull/54

PR #54, opened 2020-08-03
Add branch_configuration arg

runlevel5:
### Context

This PR would allow users to configure pipeline-level branch configuration

### Consideration

Because there is no support to update this attribute in GraphQL API. The implementation is done via the RESTful API. It comes with a cost... that is an extra HTTP call after the GraphQL API call just for updating one attribute.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,7 @@ resource "buildkite_pipeline" "repo2" {
 - `description` - (Optional) A description of the pipeline.
 - `repository` - (Required) The git URL of the repository.
 - `default_branch` - (Optional) The default branch to prefill when new builds are created or triggered.
+- `branch_configuration` - (Optional) Limit which branches and tags cause new builds to be created, either via a code push or via the Builds REST API.
 - `steps` - (Required) The string YAML steps to run the pipeline.
 - `team` - (Optional) Set team access for the pipeline. Can be specified multiple times for each team.  
 
--- a/README.md
+++ b/README.md
@@ -128,6 +128,7 @@ resource "buildkite_pipeline" "repo2" {
 - `name` - (Required) The name of the pipeline.
 - `description` - (Optional) A description of the pipeline.
 - `repository` - (Required) The git URL of the repository.
+- `default_branch` - (Optional) The default branch to prefill when new builds are created or triggered.
 - `steps` - (Required) The string YAML steps to run the pipeline.
 - `team` - (Optional) Set team access for the pipeline. Can be specified multiple times for each team.  
 
--- a/buildkite/resource_pipeline.go
+++ b/buildkite/resource_pipeline.go
@@ -3,10 +3,11 @@ package buildkite
 import (
 	"context"
 	"fmt"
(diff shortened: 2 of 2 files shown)


buildkite/resource_pipeline.go
```diff
@@ -3,17 +3,19 @@ package buildkite
...
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/shurcooL/graphql"
 )
 
 // PipelineNode represents a pipeline as returned from the GraphQL API
 type PipelineNode struct {
 	CancelIntermediateBuilds             graphql.Boolean
 	CancelIntermediateBuildsBranchFilter graphql.String
 	DefaultBranch                        graphql.String
+	BranchConfiguration                  graphql.String
```
runlevel5, line 18: @jradtilbrook despite of having this new field for this struct, I still get the following error when using the plugin:

```
Error: Field 'branchConfiguration' doesn't exist on type 'Pipeline'
```

Wondering if I miss anything here. Many thanks

jradtilbrook: It looks to me like this isn't available in the GraphQL API. You can explore the GraphQL API and run queries from https://graphql.buildkite.com/explorer.

You can see I have gotten around this for deleting a pipeline which at the time of development wasn't available - but I now see it is :tada:! So if you want to use the http client that is available to update the branch configuration when creating or updating a pipeline you could do that in the meantime. See https://github.com/jradtilbrook/terraform-provider-buildkite/blob/master/buildkite/resource_pipeline.go#L252-L259 for how I've done it previously.

Pinging @sj26 to see if they can confirm.

runlevel5: You are right. There is no `branch_configuration` under `pipeline` in GraphQL API. 

I am going to convert the code to http api instead.

runlevel5:
@jradtilbrook the PR is ready for review

README.md
```diff
@@ -128,6 +128,8 @@ resource "buildkite_pipeline" "repo2" {
 - `name` - (Required) The name of the pipeline.
 - `description` - (Optional) A description of the pipeline.
 - `repository` - (Required) The git URL of the repository.
+- `default_branch` - (Optional) The default branch to prefill when new builds are created or triggered.
```
jradtilbrook, line 131: Thanks for adding this as well! :+1:

jradtilbrook:
Thanks for the contribution @runlevel5 and nice work.  
What is your opinion of Go now that you've done more?

Closes #53

runlevel5:
>  What is your opinion of Go now that you've done more?
I like it though I am not big fan of its syntax.

buildkite/resource_pipeline.go
sj26: Confirmed, there's no `Pipeline#branchConfiguration` property at the moment, sorry. I'll raise it internally and see if it can be added.

jradtilbrook: No problems, thanks for checking and getting back to us

PR #59, opened 2020-08-20
Update DeletePipeline() with GraphQL implementation

### LINKED ISSUE #56: Delete a pipeline using GraphQL
When the pipeline resource was developed in this provider, the GraphQL API did not support deleting a pipeline so the rest API had to be used. There is now a delete mutation so this could be refactored to use GraphQL

runlevel5:
## Context

Pipeline deletion was implemented with Buildkite REST API instead of GraphQL because of lacking of deletion mutation for pipeline. Buildkite has since added this missing feature.

## Changes

* Use GraphQL to perform pipeline deletion

Fixes #56

### DIFF
--- a/buildkite/resource_pipeline.go
+++ b/buildkite/resource_pipeline.go
@@ -37,7 +37,11 @@ type PipelineNode struct {
 	UUID       graphql.String
 	WebhookURL graphql.String `graphql:"webhookURL"`
 }
+
+// PipelineAccessLevels represents a pipeline access levels as returned from the GraphQL API
 type PipelineAccessLevels graphql.String
+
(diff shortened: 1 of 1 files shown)


runlevel5:
@jradtilbrook I have manually tested this change and confirmed that it works. I wish I can write a test for the change but my Golang level is not there yet...

jradtilbrook:
> @jradtilbrook I have manually tested this change and confirmed that it works. I wish I can write a test for the change but my Golang level is not there yet...

That's okay for now, there isn't much automated testing here yet but it's on the list to add.

jradtilbrook (approved):
Awesome. Fast work on all sides to implement this :+1:

PR #63, opened 2020-10-06
README.md: added download instructions

nitrocode:
### Changes

* Added download instructions for OSX.
* Removed one example since it did not have the `team` argument which is required for the `buildkite_pipeline` resource.


jradtilbrook (approved):
Looks good. Thanks!

PR #69, opened 2020-11-27
Adds missing pipeline arguments to documentation

leungtimothy:
I noticed that the pipeline resource supports the following variables but wasn't documented.


jradtilbrook (approved):
Nice find. Thanks!

PR #76, opened 2020-12-14
Add a Makefile for common tasks

yob:
This is a super simple modern go project that uses gomodules, so the commands to run tests and compile the provider are easy to remember.

Still, it's nice to wrap them in make targets so folks less familiar with go can get up and running with development. I also expected we might end up with some more complex make targets when building a github release pipeline.


yob:
The red build is because I haven't finished setting up the pipeline, but this doesn't change any code so I'll merge

PR #77, opened 2020-12-17
add the beginnings of a Buildkite pipeline

yob:
nothing fancy yet, just checking go fmt, go test and confirming that it all compiles.


yob:
I think this is a good start, so I'll merge and iterate further (like creating GH releases) in future PRs

PR #89, opened 2021-01-16
Upgrade to terraform-plugin-sdk v2

vgrigoruk:
As the plugin explicitly states that it supports only Terraform v13+, I took the liberty to do an upgrade of terraform-plugin-sdk to v2 (which supports Terraform v12+ only.

I have tested it locally with all available resources and everything works fine.

Also: 
- bumped go version requiremenent as per https://github.com/hashicorp/terraform-plugin-sdk#go-compatibility


yob (approved):
thanks!

I had a read through the [v2 upgrade guide](https://www.terraform.io/docs/extend/guides/v2-upgrade-guide.html) and given the small surface of this provider (so far!) nothing jumped out as a problem.

Dropping support for terraform <= 0.11 in a non-major release of the provider is a bit cheeky, but we're in super early days and we can probably get away with sneaking it in before semver is required :+1:

PR #92, opened 2021-02-01
Build for PowerPC 64 LE

runlevel5:
So that I could use this provider on Linux ppc64le :)

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -23,6 +23,7 @@ builds:
     - '386'
     - arm
     - arm64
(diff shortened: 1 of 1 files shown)


yob (approved):
I have no easy way to test this, but I assume it'll just work because GOARCH 🤷‍♂️

yob:
@runlevel5 I've released this in v0.1.0 and would love to hear if it all just works for you

runlevel5:
@yob it works perfectly. Great thanks

PR #95, opened 2021-02-07
Add acceptance tests for the agent_token resource

yob:
These are the first acceptance tests for the provider, so they come with a bunch of supporting infrastructure:

* a new Make target
* Some instructions in the README
* a new pipeline step that can only run on the main branch
  * if a third party PR removes this restriction, the step will run but fail because it'll run on an agent with no access to the required credentials in SSM ParameterStore

The CI setup here is imperfect. Pull requests won't run the acceptance tests and there's a risk we'll only notice a failure in the acceptance tests after merging. For now we'll have to live with that (maybe by manually running the tests ourselves prior to merging?) while we discuss a better approach.


yob:
The following pages were useful resources while I worked on this:

* https://www.terraform.io/docs/extend/testing/acceptance-tests/testcase.html
* https://www.terraform.io/docs/extend/testing/acceptance-tests/teststep.html
* https://github.com/dnsimple/terraform-provider-dnsimple/blob/02564f2b6620760c1fe3e6002f5a28b8b029da6f/dnsimple/resource_dnsimple_record_test.go

Issue #84, opened 2020-12-19
Error adding simple pipeline. Asks for team when team is already specified.


apanzerj:
```
terraform {
  required_providers {
    buildkite = {
      source  = "buildkite/buildkite"
      version = "0.0.17"
    }
  }
}

resource "buildkite_pipeline" "test" {
  name       = "Panzer Test 1"
  repository = "github.com:pagerduty/web"

  steps = file("steps.yml")

  team {
    slug         = "sustainability"
    access_level = "MANAGE_BUILD_AND_READ"
  }
}

provider "buildkite" {
  # Configuration options
  api_token    = "redacted" # can also be set from env: BUILDKITE_API_TOKEN
  organization = "pagerduty"                                # can also be set from env: BUILDKITE_ORGANIZATION
}
```

Output:

```
$ terraform apply

An execution plan has been generated and is shown below.
Resource actions are indicated with the following symbols:
  + create

Terraform will perform the following actions:

  # buildkite_pipeline.test will be created
  + resource "buildkite_pipeline" "test" {
      + cancel_intermediate_builds = false
      + id                         = (known after apply)
      + name                       = "Panzer Test 1"
      + repository                 = "github.com:pagerduty/web"
      + skip_intermediate_builds   = false
      + slug                       = (known after apply)
      + steps                      = <<~EOT

            steps:
              - label: ':pipeline:'
                command: buildkite-agent upload
        EOT
      + team                       = [
          + {
              + access_level = "MANAGE_BUILD_AND_READ"
              + slug         = "sustainability"
            },
        ]
      + webhook_url                = (known after apply)
    }

Plan: 1 to add, 0 to change, 0 to destroy.

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

buildkite_pipeline.test: Creating...

Error: Please ensure at least one team is added to this pipeline

  on main2.tf line 16, in resource "buildkite_pipeline" "test":
  16: resource "buildkite_pipeline" "test" {
```

Note: My token has full access to all the things but is not administrator in the org. I can validate that I have... enough... permissions cause I can make the same pipeline manually in the UI. I'm a team member of the Sustainability team and that is the correct slug. 

```
$ terraform version
Terraform v0.13.5
+ provider registry.terraform.io/buildkite/buildkite v0.0.17
```

yob:
I've dug into this a bit and have some notes, but no solid plan yet.

[...]

PR #99, opened 2021-02-16
Add some aceptance tests for the pipeline resource

yob:
This resource had no acceptance tests, and we were nervous about the risk of regressions.
    
This adds a few initial test cases:
    
* creating a pipeline with only the required attributes set
* creating a pipeline with all attributes set
* updating a pipeline
* importing an existing pipeline
    
Pipeline deletion is tested implicitly when the terraform testing sdk cleans up after a test.
    
We assume there's potential for more tests that exercise specific combinations (for example, exposing this teams related bug in pipeline creation #84), but this is a good start

### DIFF
--- a/buildkite/resource_pipeline_test.go
+++ b/buildkite/resource_pipeline_test.go
@@ -1,0 +1,219 @@
+package buildkite
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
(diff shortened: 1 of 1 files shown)


yob:
I've run the full acceptance test suitre locally and confirmed it is green :+1:

chloeruka (approved):

PR #100, opened 2021-02-17
Add acceptance tests for pipeline schedule

chloeruka:
This resource also lacked acceptance tests, so in order to guard against regressions we have added the following simple acceptance tests:

* creating a schedule (with only the required attributes)
* updating a schedule (e.g. changing its cronline)
* importing a pipeline schedule

This work required the addition of UUID to the pipeline schedule resource schema. The UUID is used to assemble the slug of an existing pipeline schedule.

### DIFF
--- a/buildkite/resource_pipeline_schedule_test.go
+++ b/buildkite/resource_pipeline_schedule_test.go
@@ -1,0 +1,201 @@
+package buildkite
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+)
+
+// Confirm that we can add a new pipeline schedule to a pipeline
+func TestAccPipelineSchedule_add_remove(t *testing.T) {
+	var resourcePipeline PipelineNode
+	var resourceSchedule PipelineScheduleNode
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckPipelineScheduleResourceDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccPipelineScheduleConfigBasic("foo", "0 * * * *"),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					// Schedules need a pipeline
+					testAccCheckPipelineExists("buildkite_pipeline.foobar", &resourcePipeline),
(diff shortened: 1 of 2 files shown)


buildkite/resource_pipeline_schedule_test.go
```diff
@@ -0,0 +1,201 @@
...
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckPipelineScheduleResourceDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccPipelineScheduleConfigBasic("foo", "0 * * * *"),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					// Schedules need a pipeline
+					testAccCheckPipelineExists("buildkite_pipeline.foobar", &resourcePipeline),
```
yob, line 26: do we need to check the pipeline in these tests, or can we assume it works thanks to the specific tests for the pipeline resource?

buildkite/resource_pipeline_schedule_test.go
```diff
@@ -0,0 +1,201 @@
...
+					// Confirm the schedule exists in the buildkite API
+					testAccCheckPipelineScheduleExists("buildkite_pipeline_schedule.foobar", &resourceSchedule),
+					// Confirm the schedule has the correct values in Buildkite's system
+					testAccCheckPipelineScheduleRemoteValues(&resourcePipeline, &resourceSchedule, "main", "Test Schedule foo", "0 * * * *"),
+					// Confirm the pipeline has the correct values in terraform state
+					resource.TestCheckResourceAttr("buildkite_pipeline_schedule.foobar", "label", "Test Schedule foo"),
```
yob, line 32: is it worth checking the state has other properties set correctly as well (like cronline and branch)

yob (approved):
lovely :heart: 

I had a couple of non-blocking questions.

buildkite/resource_pipeline_schedule_test.go
chloeruka: Ahh, yeah this is the rub; we don't really need to know that it exists, but if we want to validate we put a schedule on the correct pipeline, we memref it to `testAccCheckPipelineExists` which populates the (until now nil) value of `resourcePipeline`

yob: oooohhh, of course. I keep forgetting this methods are also about grabbing an in-memory copy of the data :+1:

yob (approved):
looks good to me, I've confirmed the new tests are green when run locally for me :+1:

PR #101, opened 2021-02-17
update ci/build environment to go 1.16

yob:
1.16 was released today, and it includes support for m1 mac systems. It'll be interesting to see if the build passes, and if providing a
native version of the provider for m1 systems is possible.


yob:
build passed!

/cc @chloeruka

PR #103, opened 2021-02-19
Add acceptance tests for teams

chloeruka:
This is one of the resources that does not yet have acceptance tests associated with it. In order to guard against regressions, I have added the following tests:

* creating a team (with only the required attributes)
* updating a team (changing its name, description and visibility)
* importing a team

During this work I discovered an error in importing teams. In order to resolve this, I added the DefaultMemberRole value which was present in the schema to the TeamNode struct, and added a mapping to updateTeam, so this PR doubles as a bugfix for importing teams.

### DIFF
--- a/buildkite/resource_team_test.go
+++ b/buildkite/resource_team_test.go
@@ -140,10 +140,6 @@ func testAccCheckTeamExists(resourceName string, resourceTeam *TeamNode) resourc
 			return fmt.Errorf("No team found with graphql id: %s", resourceState.Primary.ID)
 		}
 
-		if string(query.Node.Team.Name) != resourceState.Primary.Attributes["name"] {
-			return fmt.Errorf("Team name in state doesn't match remote name")
-		}
-
 		*resourceTeam = query.Node.Team
 
 		return nil
--- a/buildkite/resource_team_test.go
+++ b/buildkite/resource_team_test.go
@@ -1,0 +1,193 @@
+package buildkite
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+)
+
+// Confirm that we can create a team, and then delete it without error
+func TestAccTeam_add_remove(t *testing.T) {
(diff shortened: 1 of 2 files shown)


buildkite/resource_team_test.go
```diff
@@ -0,0 +1,193 @@
...
+			return fmt.Errorf("Error fetching team from graphql API: %v", err)
+		}
+
+		if string(query.Node.Team.ID) == "" {
+			return fmt.Errorf("No team found with graphql id: %s", resourceState.Primary.ID)
+		}
+
+		if string(query.Node.Team.Name) != resourceState.Primary.Attributes["name"] {
+			return fmt.Errorf("Team name in state doesn't match remote name")
+		}
```
yob, line 145: In our other resource tests I think we only checked computer properties in this function, and I also note that name is checked in `testAccCheckTeamRemoteValues`. Do we definitely want to check it here too?

yob:
> During this work I discovered an error in importing teams. In order to resolve this, I added the DefaultMemberRole value which was present in the schema to the TeamNode struct, and added a mapping to updateTeam, so this PR doubles as a bugfix for importing teams.

Nice, tests work!

Purely out of interesting, could you describe this bug in a bit more detail? Did it manifest as an error when updating a team during `terraform apply`?

buildkite/resource_team_test.go
chloeruka: Probably excessive, I'll remove it!

chloeruka:
> Purely out of interesting, could you describe this bug in a bit more detail? Did it manifest as an error when updating a team during `terraform apply`?

image

When importing a resource, it was failing on `ImportStateVerify: true` because the returned resource did not have the field `DefaultMemberState` present in the definition of the resource:

```HCL
resource "buildkite_team" "test" {
  name = "example"
  description = "a cool team of example"
  privacy = "VISIBLE"
  default_team = true
  default_member_role = "MEMBER" # this field
}
```

yob:
Oh, right. So the schema had `default_member_role` and the provider would include it in team updates/creates, but it would never read the value back from buildkite and update the state?

Cool, nice find :sparkles:

yob (approved):
ship it :+1:

PR #104, opened 2021-02-19
update goreleaser from v0.149.0 to v0.157.0

### LINKED ISSUE #94: Add darwin/arm64 target once golang 1.16 is released
Once golang 1.16 is released with support for m1 systems from Apple, it'd be neat to add it as a compilation target for this provider.

I haven't checked if terraform is officially shipping support for m1 system yet though, we should probably check that first.

yob:
[v0.156.x adds support for compiling a native binary for Apple Silicon systems](https://github.com/goreleaser/goreleaser/releases/tag/v0.156.2):

> GOOS=darwin GOARCH=arm64 was ignored before this release, since it was
> a build for iOS that most people didn't care about and required more
> than just go to build.
>
> On Go 1.16, this target is now the macOS on Apple Silicon. GoReleaser
> v0.156.0+ is built for Go 1.16+ and will try to build the darwin arm64
> target if the config allows it

This is hard to test, we might just have to wait until we're ready to make an actual release.

Closes #94 (probably).

### DIFF
--- a/.buildkite/steps/release.sh
+++ b/.buildkite/steps/release.sh
@@ -45,9 +45,9 @@ fi
 
 echo "--- installing goreleaser"
 
-curl -L -o /tmp/goreleaser_Linux_x86_64.tar.gz https://github.com/goreleaser/goreleaser/releases/download/v0.149.0/goreleaser_Linux_x86_64.tar.gz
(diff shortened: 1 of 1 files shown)


yob:
Extra context: our `.goreleaser.yml` already has darwin/arm64 in the build matrix, but it was ignored before v0.156.0:

```yaml
  goos:
    - freebsd
    - windows
    - linux
    - darwin
  goarch:
    - amd64
    - '386'
    - arm
    - arm64
    - ppc64le

```

chloeruka (approved):
Yep!

PR #106, opened 2021-02-19
Add a pipeline data source

yob:
This adds a data source for fetching the properties of a single pipeline. The pipeline is specified using a slug, available in the
pipeline URL on buildkite.com.

This could be particularly useful for folks who want to lookup the webhook URL and set it on a GitHub repo (using the  GitHub terraform provider).

Closes #93

PR #107, opened 2021-02-21
stop using deprecated part of the terraform sdk API

yob:
Blocked by #100 and #103. I'm not keen to merge this until we have basic acceptance tests in place to detect regressions.

While working on #106 I noticed that v2 of the terraform sdk has deprecated the Create/Read/Update/Delete properties of the schema.Resource struct in favour of new properties that support passing context.

I thought I'd bite the bullet and just make the change everywhere so all current resources are a good model for adding new resources in the future.

Note that we only updated to v2 of the sdk in #89 ~3 weeks ago, so this deprecation only appeared for us recently.

I haven't made an effort to use the new context instances to improve the way timeouts are handled when we query the graphql API. That could be worth exploring in the future.

### DIFF
--- a/buildkite/resource_agent_token.go
+++ b/buildkite/resource_agent_token.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/shurcooL/graphql"
 )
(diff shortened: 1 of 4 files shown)


chloeruka (approved):
Good find!

yob:
Cool. I'll merge once #100 and #103 land :sparkles:

chloeruka:
:eyes:

yob:
I've rebased onto main and force pushed

chloeruka (approved):

PR #108, opened 2021-02-22
default acceptance tests to verbose

yob:
We don't have **that** many tests, and I find it useful to see the list of what's run along with timing.


chloeruka (approved):
Yeah, definitely!

PR #111, opened 2021-02-25
Ensure all resources are destroyed after tests

### LINKED ISSUE #110: Ensure resources are destroyed after acceptance tests
As part of ongoing maintenance of this repository, we need to ensure that all resources created by Terraform during acceptance tests are tidied up so that we don't end up with an inconsistent environment state. Terraform does its best, and _usually_ eliminates the resources, but if a runtime exception occurs, orphaned resources can be left behind.

- [x] [buildkite/resource_team_test.go](https://github.com/buildkite/terraform-provider-buildkite/blob/main/buildkite/resource_team_test.go)
- [x] [buildkite/resource_agent_token_test.go](https://github.com/buildkite/terraform-provider-buildkite/blob/main/buildkite/resource_agent_token_test.go)
- [x] [buildkite/resource_pipeline_test.go](https://github.com/buildkite/terraform-provider-buildkite/blob/main/buildkite/resource_pipeline_test.go)
- [x] [buildkite/resource_pipeline_schedule_test.go](https://github.com/buildkite/terraform-provider-buildkite/blob/main/buildkite/resource_pipeline_schedule_test.go)

chloeruka:
Resolve #110 by adding concrete cleanup checks for resource destroy and acceptance tests to all resources to ensure that Terraform can generate a plan to remove a resource (except agent tokens, as they are revoked rather than destroyed).

`testAccCheck*ResourceDestroy` unfortunately can't easily be tested unless Terraform hadn't cleaned up the resource. Hopefully, all the same, this gives us some good meaningful logging if something does break in the future. Plus we shedded TODOs.

### DIFF
--- a/buildkite/provider_test.go
+++ b/buildkite/provider_test.go
@@ -1,9 +1,13 @@
 package buildkite
 
 import (
+	"context"
+	"fmt"
 	"os"
 	"testing"
 
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
 )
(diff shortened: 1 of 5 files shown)


yob (commented):
Nice!

I was surprised to see the new CheckDestroy methods work by enumerating resources in the state and confirm they don't exist remotely anymore. I naively assumed these methods would do index/list API calls to Buildkite and confirm there was no data hanging around. Then I [read the docs properly and found the pattern is exactly as hashicorp recommend](https://www.terraform.io/docs/extend/testing/acceptance-tests/testcase.html). Woo!

I'm curious about the new tests that use `testAccCheckResourceDisappears` - what are they asserting? Are they calling the resource `DeleteContext` manually to confirm it works as expected?

chloeruka:
> I was surprised to see the new CheckDestroy methods work by enumerating resources in the state and confirm they don't exist remotely anymore. I naively assumed these methods would do index/list API calls to Buildkite and confirm there was no data hanging around. Then I [read the docs properly and found the pattern is exactly as hashicorp recommend](https://www.terraform.io/docs/extend/testing/acceptance-tests/testcase.html). Woo!

Yep, it feels strange to me too, but I aligned it to the pattern they use in the AWS terraform provider.

> I'm curious about the new tests that use `testAccCheckResourceDisappears` - what are they asserting? Are they calling the resource `DeleteContext` manually to confirm it works as expected?

They are, but the main thing is they also delete it from state and confirm the query plan isn't empty (i.e. there is a negative changeset consistent with a destroyed resource). This is another pattern well represented in the AWS provider tests, so I thought it would be good to add. This is the method doing all the work (in provider_test.go)
https://github.com/buildkite/terraform-provider-buildkite/blob/6a29ffc47cfb3a471ca62f6593979f56a71a00a9/buildkite/provider_test.go#L46-L72

yob:
> They are, but the main thing is they also delete it from state and confirm the query plan isn't empty (i.e. there is a negative changeset consistent with a destroyed resource).

Oooh.

So they're testing that if the resource is removed from the state and deleted on buildkite.com but the terraform config doesn't change, then a subsequent terraform plan should want to re-create the resource?

chloeruka:
> So they're testing that if the resource is removed from the state and deleted on buildkite.com but the terraform config doesn't change, then a subsequent terraform plan should want to re-create the resource?

I think so, yea.

yob (approved):
nice :+1:

PR #112, opened 2021-02-28
Adjust pipeline creation so it works for non-admins

### LINKED ISSUE #84: Error adding simple pipeline. Asks for team when team is already specified.

yob:
For Buildkite organizations with teams enabled, it's possible for non-admins to create pipelines within their own teams.

However, it only works if the team is included in the `pipelineCreate()` mutation. For a non-admin graphql token, `pipelineCreate() `will return an error if the input includes no teams.

The existing code was calling `pipelineCreate() `with no teams, and then following up with a `teamPipelineCreate()` mutation to apply the teams. We never got to the follow up though, because the initial `pipelineCreate()` was failing.

This merges the two steps so the initial `pipelineCreate()` includes teams. Once the pipeline has been created, team management is still done independently of the rest of the pipeline. That should be fine though, it's only `pipelineCreate()` that requires teams to be include2. Non-admin graphql tokens can call `pipelineUpdate()` without teams just fine.

Still needs a fair bit of polish:

* [x] confirm createPipeline() still works on orgs without teams enabled
* [x] confirm a non-admin graphql token can call updatePipeline() with no teams included
* [ ] acceptance tests would be really nice, if we can workout how to use a different API token in a subset of tests
 
Fixes #84

### DIFF
--- a/buildkite/resource_pipeline.go
+++ b/buildkite/resource_pipeline.go
@@ -48,6 +48,11 @@ type TeamPipelineNode struct {
 	Team        TeamNode
 }
 
+type PipelineTeamAssignmentInput struct {
+	Id          graphql.ID     `json:"id"`
+	AccessLevel graphql.String `json:"accessLevel"`
+}
+
 // resourcePipeline represents the terraform pipeline resource schema
 func resourcePipeline() *schema.Resource {
 	return &schema.Resource{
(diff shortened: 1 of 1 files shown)


yob:
The permutations here are a real headache, and I'd love to get some acceptance tests in place. Without tests, I'm sure we'll inadvertently break the experience for non-admin users who want to use the terraform provider.

Interestingly, in the current provider there's very few things a non-admin user can do beyond managing pipelines and pipeline schedules. The other resources (teams, agent tokens) require admin privileges, and future resources (like user) would presumably also be admin-only. 

This PR was 10% working out the graphql query to include teams into `pipelinesCreate()`, 90% wrestling with `github.com/shurcooL/graphql` to get it to generate the query shape I wanted. I am very not-fond of it as a result, and would be open to replacing it without something easier to use. Or maybe someone can just show the right way to hold it and it'll all be fine.

buildkite/resource_pipeline.go
```diff
@@ -153,11 +158,27 @@ func CreatePipeline(ctx context.Context, d *schema.ResourceData, m interface{})
...
 			Pipeline PipelineNode
-		} `graphql:"pipelineCreate(input: {cancelIntermediateBuilds: $cancel_intermediate_builds, cancelIntermediateBuildsBranchFilter: $cancel_intermediate_builds_branch_filter, defaultBranch: $default_branch, description: $desc, name: $name, organizationId: $org, repository: {url: $repository_url}, skipIntermediateBuilds: $skip_intermediate_builds, skipIntermediateBuildsBranchFilter: $skip_intermediate_builds_branch_filter, steps: {yaml: $steps}})"`
+		} `graphql:"pipelineCreate(input: {cancelIntermediateBuilds: $cancel_intermediate_builds, cancelIntermediateBuildsBranchFilter: $cancel_intermediate_builds_branch_filter, defaultBranch: $default_branch, description: $desc, name: $name, organizationId: $org, repository: {url: $repository_url}, skipIntermediateBuilds: $skip_intermediate_builds, skipIntermediateBuildsBranchFilter: $skip_intermediate_builds_branch_filter, steps: {yaml: $steps}, teams: $teams})"`
+	}
+
+	teamsData := make([]PipelineTeamAssignmentInput, 0)
+	for _, team := range teamPipelines {
+		log.Printf("converting team slug '%s' to an ID", string(team.Team.Slug))
+		teamID, err := GetTeamID(string(team.Team.Slug), client)
```
yob, line 171: this is a bit gross - for each team slug listed in the pipeline we do a graphql query to convert it into an ID. It's no worse than what we have on main though - `reconcileTeamPipelines()` also calls `GetTeamID()`

chloeruka (approved):
Paired with @yob on this, where we confirmed that the pipeline can still be created under the correct conditions (teams not enabled, teams enabled but non-admin)

yob:
@chloeruka and I have manually confirm pipeline creating and updating work with both admin and non-admin API tokens.

We also have a plan for acceptance testing with non-admin tokens, which we'll tackle in a follow up PR.

PR #114, opened 2021-02-28
adjust cron line we use in pipeline_schedule update test

yob:
pipeline schedules are not permitted to run more frequently than every 10 minutes.

This test was updating the cronline to `* 0 * * *`, which is "every minute during hour 0". Apparently our server side validation only
detects that as an error when the update runs during UTC hour 0, so [this test fails when run during UTC hour 0](https://buildkite.com/buildkite/terraform-provider-buildkite-main/builds/43#c6ca85d2-9311-43c8-91a0-9290460242c0).

The test now updates the cronline to "daily at 1:00 UTC".

### DIFF
--- a/buildkite/resource_pipeline_schedule_test.go
+++ b/buildkite/resource_pipeline_schedule_test.go
@@ -59,7 +59,7 @@ func TestAccPipelineSchedule_update(t *testing.T) {
 				),
 			},
 			{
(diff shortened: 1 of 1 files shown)


yob:
sneaky merge 'cos this seems so uncontroversial 🙈

chloeruka:
I know I've come across this before, but it's pretty easy to forget 😳 – we should improve the serverside validation so it knocks these invalid cronlines's back.

PR #123, opened 2021-03-10
Add provider_settings to pipeline resource

vgrigoruk:
## provider_settings

1. `provider_settings` block is `Optional` for backward-compatibility reasons. Also it is `Computed` (so default values are set by buildkite backend, while it is still possible to override the values by specifying custom values for each property in terraform resource configuration).

1. all `provider_settings` are provider-agnostic (the user should know which attributes are supported by the provider they use)

## Few examples:

- to disable build triggers:

  ```hcl
  provider_settings {
    trigger_mode = "none"
  }
  ```
- trigger builds for tags:
  ```hcl
  provider_settings {
    trigger_mode = "code" # default value, so could be omitted (not recommended though) 
    build_tags = true
  }
  ```

## Other changes

1. Removed `ReadPipeline` calls at the end of `CreatePipeline` and `UpdatePipeline` methods, as we already set attribute values for underlying terraform resource directly in these methods (based on a result of GraphQL mutation and response from  PATCH request). I know that according to [best practices](https://www.terraform.io/docs/extend/best-practices/detecting-drift.html#update-state-after-modification) we should call `READ` at the end of `CREATE` and `UPDATE`, but it feels like we're doing unnecessary API calls to Buildkite backend here.  I can drop this change from this PR if you want, or create a new PR with this change only if you think this is the right thing to do.

## TODO

- [x] update docs
- [x] acceptance tests

### DIFF
--- a/buildkite/client.go
+++ b/buildkite/client.go
@@ -1,7 +1,10 @@
 package buildkite
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
+	"io/ioutil"
 	"net/http"
 
 	"github.com/shurcooL/graphql"
@@ -27,4 +30,33 @@ func NewClient(org, apiToken string) *Client {
 		http:         httpClient,
 		organization: org,
 	}
+}
+
+func (client *Client) makeRequest(method string, url string, postData interface{}, responseObject interface{}) error {
+	jsonPayload, err := json.Marshal(postData)
+	if err != nil {
+		return err
+	}
+
+	req, err := http.NewRequest(method, url, bytes.NewBuffer(jsonPayload))
+	if err != nil {
+		return err
+	}
+
+	resp, err := client.http.Do(req)
+	if err != nil && resp.StatusCode >= 400 {
+		return err
+	}
+
+	defer resp.Body.Close()
+	responseBody, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return err
+	}
+
+	if err := json.Unmarshal(responseBody, responseObject); err != nil {
+		return err
+	}
+
+	return nil
 }
--- a/buildkite/client.go
+++ b/buildkite/client.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"net/http"
 
@@ -44,8 +45,11 @@ func (client *Client) makeRequest(method string, url string, postData interface{
 	}
 
 	resp, err := client.http.Do(req)
-	if err != nil && resp.StatusCode >= 400 {
+	if err != nil {
 		return err
+	}
+	if resp.StatusCode >= 400 {
+		return fmt.Errorf("Buildkite API request failed: %s %s (status: %d)", method, url, resp.StatusCode)
 	}
 
 	defer resp.Body.Close()
--- a/buildkite/resource_pipeline.go
+++ b/buildkite/resource_pipeline.go
@@ -1,12 +1,9 @@
 package buildkite
 
 import (
-	"bytes"
 	"context"
-	"encoding/json"
 	"fmt"
 	"log"
-	"net/http"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
@@ -61,7 +58,7 @@ func resourcePipeline() *schema.Resource {
 		UpdateContext: UpdatePipeline,
 		DeleteContext: DeletePipeline,
 		Importer: &schema.ResourceImporter{
-			State: schema.ImportStatePassthrough,
+			StateContext: schema.ImportStatePassthroughContext,
 		},
 
 		Schema: map[string]*schema.Schema{
@@ -145,6 +142,104 @@ func resourcePipeline() *schema.Resource {
 					},
 				},
 			},
+			"provider_settings": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Computed: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"trigger_mode": {
+							Computed: true,
+							Optional: true,
+							Type:     schema.TypeString,
+							ValidateFunc: func(val interface{}, key string) (warns []string, errs []error) {
+								switch v := val.(string); v {
+								case "code":
+								case "deployment":
+								case "fork":
+								case "none":
+									return
+								default:
+									errs = append(errs, fmt.Errorf("%q must be one of code, deployment, fork or none, got: %s", key, v))
+									return
+								}
+								return
+							},
+						},
+						"build_pull_requests": {
+							Computed: true,
+							Optional: true,
+							Type:     schema.TypeBool,
+						},
+						"pull_request_branch_filter_enabled": {
+							Computed: true,
+							Optional: true,
+							Type:     schema.TypeBool,
+						},
+						"pull_request_branch_filter_configuration": {
+							Computed: true,
+							Optional: true,
+							Type:     schema.TypeString,
+						},
+						"skip_pull_request_builds_for_existing_commits": {
+							Computed: true,
+							Optional: true,
+							Type:     schema.TypeBool,
+						},
+						"build_pull_request_ready_for_review": {
+							Computed: true,
+							Optional: true,
+							Type:     schema.TypeBool,
+						},
+						"build_pull_request_forks": {
+							Computed: true,
+							Optional: true,
+							Type:     schema.TypeBool,
+						},
+						"prefix_pull_request_fork_branch_names": {
+							Computed: true,
+							Optional: true,
+							Type:     schema.TypeBool,
+						},
+						"build_branches": {
+							Computed: true,
+							Optional: true,
+							Type:     schema.TypeBool,
+						},
+						"build_tags": {
+							Computed: true,
+							Optional: true,
+							Type:     schema.TypeBool,
+						},
+						"cancel_deleted_branch_builds": {
+							Computed: true,
+							Optional: true,
+							Type:     schema.TypeBool,
+						},
+						"publish_commit_status": {
+							Computed: true,
+							Optional: true,
+							Type:     schema.TypeBool,
+						},
+						"publish_blocked_as_pending": {
+							Computed: true,
+							Optional: true,
+							Type:     schema.TypeBool,
+						},
+						"publish_commit_status_per_step": {
+							Computed: true,
+							Optional: true,
+							Type:     schema.TypeBool,
+						},
+						"separate_pull_request_statuses": {
+							Computed: true,
+							Optional: true,
+							Type:     schema.TypeBool,
+						},
+					},
+				},
+			},
 			"webhook_url": {
 				Computed: true,
 				Type:     schema.TypeString,
@@ -205,7 +300,12 @@ func CreatePipeline(ctx context.Context, d *schema.ResourceData, m interface{}) 
 	log.Printf("Successfully created pipeline with id '%s'.", mutation.PipelineCreate.Pipeline.ID)
 
 	updatePipelineResource(d, &mutation.PipelineCreate.Pipeline)
-	updatePipelineWithRESTfulAPI(d, client)
+
+	pipelineExtraInfo, err := updatePipelineExtraInfo(d, client)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	updatePipelineResourceExtraInfo(d, &pipelineExtraInfo)
 
 	return ReadPipeline(ctx, d, m)
 }
@@ -228,6 +328,14 @@ func ReadPipeline(ctx context.Context, d *schema.ResourceData, m interface{}) di
 	}
 
 	updatePipelineResource(d, &query.Node.Pipeline)
+
+	if slug, pipelineExists := d.GetOk("slug"); pipelineExists {
+		pipelineExtraInfo, err := getPipelineExtraInfo(d, m, slug.(string))
+		if err != nil {
+			return diag.FromErr(err)
+		}
+		updatePipelineResourceExtraInfo(d, pipelineExtraInfo)
+	}
 
 	return nil
 }
@@ -268,7 +376,13 @@ func UpdatePipeline(ctx context.Context, d *schema.ResourceData, m interface{}) 
 	}
 
 	updatePipelineResource(d, &mutation.PipelineUpdate.Pipeline)
-	updatePipelineWithRESTfulAPI(d, client)
+
+	pipelineExtraInfo, err := updatePipelineExtraInfo(d, client)
+	if err != nil {
+		log.Print("Unable to update pipeline attributes via REST API")
+		return diag.FromErr(err)
+	}
+	updatePipelineResourceExtraInfo(d, &pipelineExtraInfo)
 
 	return ReadPipeline(ctx, d, m)
 }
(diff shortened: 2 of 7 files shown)


buildkite/resource_pipeline.go
```diff
@@ -295,37 +417,60 @@ func DeletePipeline(ctx context.Context, d *schema.ResourceData, m interface{})
...
+			SeparatePullRequestStatuses             bool   `json:"separate_pull_request_statuses"`
+		} `json:"settings"`
+	} `json:"provider"`
+}
 
-	jsonStr, err := json.Marshal(payload)
+func getPipelineExtraInfo(d *schema.ResourceData, m interface{}, slug string) (*PipelineExtraInfo, error) {
+	client := m.(*Client)
+	pipelineExtraInfo := PipelineExtraInfo{}
+	err := client.makeRequest("GET", fmt.Sprintf("https://api.buildkite.com/v2/organizations/%s/pipelines/%s", client.organization, slug), nil, &pipelineExtraInfo)
```
vgrigoruk, line 452: maybe we should move `https://api.buildkite.com/v2/organizations/%s/` part of the URL to `client.makeRequest` method, so we don't need to call `client.organization` here and duplicate the URL twice in this file.

vgrigoruk: Or maybe just use https://github.com/buildkite/go-buildkite ? I'm wondering if there are any plans internally for adding `provider_settings` and `branch_configuration` to GraphQL API?

yob:
Sorry for taking a few days to get to this @vgrigoruk!

This is looking super interesting, and I'm keen to help get it across the line. A few thoughts in no particular order.

> Added Compute: true to all pipeline resource attributes that has default values set by Buildkite (this fixes a lot of warnings that you can see running terraform apply for a with TF_LOG=debug env variable (maybe it should be done in a separate PR).

A separate PR for this would be great. It sounds uncontroversial and fairly easy to review, so we can commit to reviewing it quickly.

> Removed ReadPipeline calls at the end of CreatePipeline and UpdatePipeline methods, as we already set attribute values for underlying terraform resource directly in these methods (based on a result of GraphQL mutation and response from PATCH request). I know that according to best practices we should call READ at the end of CREATE and UPDATE, but it feels like we're doing unnecessary API calls to Buildkite backend here. I can drop this change from this PR if you want, or create a new PR with this change only if you think this is the right thing to do.

A separate PR for this would be great too. I'm not across what the implications of this would be, so I think we'd want to run some extra manual tests to ensure we understand it. It sounds like a sensible optimisation, but then why do hashicorp recommend doing it the less-optimal way? It also sounds like the PR wouldn't be a blocker for this one, does that match your udnerstanding?

> I'm wondering if there are any plans internally for adding provider_settings and branch_configuration to GraphQL API?

It's definitely our goal to make the graphql API a superset of the REST one, so those options should eventually make it into the graphql API. I'd hate to see this PR blocked waiting for that though, so falling back to REST is reasonable for now.

> Or maybe just use https://github.com/buildkite/go-buildkite ?

We could use go-buildkite, but I think we're happy with keeping it manual. **Hopefully** it's a temporary fallback, and the fallback shouldn't be needed in too many places.

> Few examples

There's a lot of permutations of the settings this adds, and the useful patterns may not be obvious. It'd be great if the final PR includes a few common configurations in `docs/resources/pipeline.md`.

vgrigoruk:
>> Removed ReadPipeline calls at the end of CreatePipeline and UpdatePipeline methods, as we already set attribute values for underlying terraform resource directly in these methods (based on a result of GraphQL mutation and response from PATCH request). I know that according to best practices we should call READ at the end of CREATE and UPDATE, but it feels like we're doing unnecessary API calls to Buildkite backend here. I can drop this change from this PR if you want, or create a new PR with this change only if you think this is the right thing to do.

> A separate PR for this would be great too. I'm not across what the implications of this would be, so I think we'd want to run some extra manual tests to ensure we understand it. It sounds like a sensible optimisation, but then why do hashicorp recommend doing it the less-optimal way? It also sounds like the PR wouldn't be a blocker for this one, does that match your udnerstanding?

In  [an example from hashicorp](https://www.terraform.io/docs/extend/best-practices/detecting-drift.html#update-state-after-modification), a `resourceExampleSimpleCreate` methods performs an API request that doesn't return any data: 
```
func resourceExampleSimpleRead(d *schema.ResourceData, meta interface{}) error {
   client := meta.(*ProviderApi).client
   resource, _ := client.GetResource(d.Id())
   d.Set("name", resource.Name)
   d.Set("type", resource.Type)
   return nil
}

func resourceExampleSimpleCreate(d *schema.ResourceData, meta interface{}) error {
   client := meta.(*ProviderApi).client
   name := d.Get("name").(string)
   client.CreateResource(name) # no data is returned from the backend here
   d.SetId(name)
   return resourceExampleSimpleRead(d, meta)
}
``` 

However, in the case of `buildkite.resource_pipeline`, Buildkite API (both GraphQL and REST) return pipeline attributes in a mutation result (for GraphQL) and inside a response to HTTP PATCH request.
Taking this into account, replacing `return ReadPipeline` with `return nil` at the end of `CreatePipeline` and `UpdatePipeline` should reduce the number of requests that are sent to Buildkite API by `2` each time pipeline resource is created/updated.

yob:
> However, in the case of buildkite.resource_pipeline, Buildkite API (both GraphQL and REST) return pipeline attributes in a mutation result (for GraphQL) and inside a response to HTTP PATCH request.
> Taking this into account, replacing return ReadPipeline with return nil at the end of CreatePipeline and UpdatePipeline should reduce the number of requests that are sent to Buildkite API by 2 each time pipeline resource is created/updated.

This makes a lot of sense, cheers. Are you up for making that change in a separate PR as well? We love small PRs :heart:

vgrigoruk:
@yob I believe this is ready for a final review now. I've updated the docs, however I'm not sure what to do about acceptance tests, as the tests should depend on a specific source control provider that is configured for test organization.
Also, I'll create a separate PR for https://github.com/buildkite/terraform-provider-buildkite/pull/123#issuecomment-800694522 after this one is merged.

yob:
Thanks @vgrigoruk, this is really neat. 

> I'm not sure what to do about acceptance tests, as the tests should depend on a specific source control provider that is configured for test organization.

I think we can mostly ignore how the provider settings behave differently with different source control providers, and just confirm that the new resource properties flow through to the server side correctly. If need be, we *could* set the repository ID to match a different provider (tests with github.com, gitlab.com and bitbucket.org repo URLs), but even that may not be necessary. There's no easy way to test GitHub Enterprise - the URLs for that are custom.

Can we get away with just adding all of the new provider settings into the config within `testAccPipelineConfigComplex()`?

I tried running the acceptance test suite and I'm getting a consistent failure, do you see the same?

```
$ BUILDKITE_ORGANIZATION=yyy BUILDKITE_API_TOKEN=xxx make testacc
TF_ACC=1 go test -v ./...
?       github.com/buildkite/terraform-provider-buildkite       [no test files]
=== RUN   TestAccDataPipeline_read
--- PASS: TestAccDataPipeline_read (6.65s)
=== RUN   TestProvider
--- PASS: TestProvider (0.00s)
=== RUN   TestProvider_impl
--- PASS: TestProvider_impl (0.00s)
=== RUN   TestAccAgentToken_add_remove
--- PASS: TestAccAgentToken_add_remove (2.74s)
=== RUN   TestAccAgentToken_update
--- PASS: TestAccAgentToken_update (5.24s)
=== RUN   TestAccAgentToken_import
--- PASS: TestAccAgentToken_import (3.00s)
=== RUN   TestAccPipelineSchedule_add_remove
--- PASS: TestAccPipelineSchedule_add_remove (5.79s)
=== RUN   TestAccPipelineSchedule_add_remove_withteams
--- PASS: TestAccPipelineSchedule_add_remove_withteams (5.79s)
=== RUN   TestAccPipelineSchedule_update
--- PASS: TestAccPipelineSchedule_update (11.19s)
=== RUN   TestAccPipelineSchedule_update_withteams
--- PASS: TestAccPipelineSchedule_update_withteams (15.84s)
=== RUN   TestAccPipelineSchedule_import
--- PASS: TestAccPipelineSchedule_import (6.62s)
=== RUN   TestAccPipelineSchedule_disappears
--- PASS: TestAccPipelineSchedule_disappears (6.03s)
=== RUN   TestAccPipeline_add_remove
--- PASS: TestAccPipeline_add_remove (4.49s)
=== RUN   TestAccPipeline_add_remove_complex
    resource_pipeline_test.go:40: Step 1/1 error: Check failed: 1 error occurred:
                * Check 8/12 error: buildkite_pipeline.foobar: Attribute 'branch_configuration' expected "main", got ""
         
--- FAIL: TestAccPipeline_add_remove_complex (2.82s)
=== RUN   TestAccPipeline_add_remove_withteams
--- PASS: TestAccPipeline_add_remove_withteams (9.89s)
=== RUN   TestAccPipeline_update
--- PASS: TestAccPipeline_update (8.13s)
=== RUN   TestAccPipeline_update_withteams
--- PASS: TestAccPipeline_update_withteams (7.33s)
=== RUN   TestAccPipeline_import
--- PASS: TestAccPipeline_import (4.42s)
=== RUN   TestAccPipeline_disappears
--- PASS: TestAccPipeline_disappears (11.19s)
=== RUN   TestAccTeam_add_remove
--- PASS: TestAccTeam_add_remove (3.68s)
=== RUN   TestAccTeam_update
--- PASS: TestAccTeam_update (6.10s)
=== RUN   TestAccTeam_import
--- PASS: TestAccTeam_import (3.27s)
=== RUN   TestAccTeam_disappears
--- PASS: TestAccTeam_disappears (2.20s)
FAIL
FAIL    github.com/buildkite/terraform-provider-buildkite/buildkite     132.422s
FAIL
make: *** [Makefile:21: testacc] Error 1
```

I also tried a manual test with the following config:

```hcl
terraform {
  required_providers {
    buildkite = {
      version = "0.0.18"
    }
  }
}

provider "buildkite" {
}

resource "buildkite_pipeline" "test" {
  name       = "Test 1"
  repository = "https://github.com/buildkite/terraform-provider-buildkite.git"

  cancel_intermediate_builds = true
  steps = ""

  provider_settings {
    trigger_mode = "none"
  }
}

output "pipeline-cancel-intermediate-builds" {
  value = buildkite_pipeline.test.cancel_intermediate_builds
}
```

Creating the pipeline works, and checking the UI on buildkite.com shows the pipeline was created with a trigger mode of none. However, subsequent `terraform apply` runs with no changes to the config continue to show a plan that wants to re-set the trigger mode. Do you see the same behaviour?

I think an acceptance spec that sets the provider settings would pick this case up.

vgrigoruk:
Hi @yob, thanks for taking a time to test the PR. However, I can not reproduce the issues you're describing (neither with automated acceptance tests, nor with manual testing).

Here is how the output of automated acceptance tests looks like for the latest commit I've just added:
```
❯ BUILDKITE_ORGANIZATION=xxx BUILDKITE_API_TOKEN=yyy make testacc
TF_ACC=1 go test -v ./...
?       github.com/buildkite/terraform-provider-buildkite       [no test files]
=== RUN   TestAccDataPipeline_read
--- PASS: TestAccDataPipeline_read (6.65s)
=== RUN   TestProvider
--- PASS: TestProvider (0.00s)
=== RUN   TestProvider_impl
--- PASS: TestProvider_impl (0.00s)
=== RUN   TestAccAgentToken_add_remove
--- PASS: TestAccAgentToken_add_remove (2.62s)
=== RUN   TestAccAgentToken_update
--- PASS: TestAccAgentToken_update (4.80s)
=== RUN   TestAccAgentToken_import
--- PASS: TestAccAgentToken_import (3.22s)
=== RUN   TestAccPipelineSchedule_add_remove
--- PASS: TestAccPipelineSchedule_add_remove (4.55s)
=== RUN   TestAccPipelineSchedule_add_remove_withteams
--- PASS: TestAccPipelineSchedule_add_remove_withteams (5.47s)
=== RUN   TestAccPipelineSchedule_update
--- PASS: TestAccPipelineSchedule_update (8.38s)
=== RUN   TestAccPipelineSchedule_update_withteams
--- PASS: TestAccPipelineSchedule_update_withteams (16.26s)
=== RUN   TestAccPipelineSchedule_import
--- PASS: TestAccPipelineSchedule_import (5.10s)
=== RUN   TestAccPipelineSchedule_disappears
--- PASS: TestAccPipelineSchedule_disappears (4.54s)
=== RUN   TestAccPipeline_add_remove
--- PASS: TestAccPipeline_add_remove (4.04s)
=== RUN   TestAccPipeline_add_remove_complex
--- PASS: TestAccPipeline_add_remove_complex (6.93s)
=== RUN   TestAccPipeline_add_remove_withteams
--- PASS: TestAccPipeline_add_remove_withteams (3.99s)
=== RUN   TestAccPipeline_update
--- PASS: TestAccPipeline_update (6.82s)
=== RUN   TestAccPipeline_update_withteams
--- PASS: TestAccPipeline_update_withteams (7.59s)
=== RUN   TestAccPipeline_import
--- PASS: TestAccPipeline_import (8.15s)
=== RUN   TestAccPipeline_disappears
--- PASS: TestAccPipeline_disappears (3.45s)
=== RUN   TestAccTeam_add_remove
--- PASS: TestAccTeam_add_remove (3.01s)
=== RUN   TestAccTeam_update
--- PASS: TestAccTeam_update (6.91s)
=== RUN   TestAccTeam_import
--- PASS: TestAccTeam_import (3.26s)
=== RUN   TestAccTeam_disappears
--- PASS: TestAccTeam_disappears (2.42s)
PASS
ok      github.com/buildkite/terraform-provider-buildkite/buildkite     118.549s
```

Also, I did manual tests with the config you provided, and `import, plan, apply, refresh` behave as expected. 
I've been rebasing PR branch and force-pushing it to remote yesterday. Could you check if your local copy was up-to-date with https://github.com/peakon/terraform-provider-buildkite/tree/vgrigoruk/pipeline-github-provider-settings ?

yob:
I tried again with commit fa66a22 and got the same failures. I then checked the permissions on my API token and found:

1. `graphl`
2. `write pipelines`

If I add the `read pipelines` scope, then the acceptance tests and my manual test pass. Do you have the `read pipelines` scope, and if you remove it do you get the same failures?

If so, that raises two questions:

1. It'd be good to avoid adding the need for an extra scope, given we hope to require only graphql access eventually. If we do need the new scope though, then it'd be good to document that somewhere. I think the existing README and docs list the required permissions for the token
2. If a missing scope was my issue, it's curious that I didn't get an error and it silently failed. I wonder if we can pinpoint why that is and if we can make it error more loudly?

buildkite/client.go
```diff
@@ -28,3 +31,32 @@ func NewClient(org, apiToken string) *Client {
...
+
+	req, err := http.NewRequest(method, url, bytes.NewBuffer(jsonPayload))
+	if err != nil {
+		return err
+	}
+
+	resp, err := client.http.Do(req)
+	if err != nil && resp.StatusCode >= 400 {
+		return err
+	}
```
yob, line 49: my spacing here is stuffed, but I think if we check for non-200 status code independently of the err, then my scope failure will be fatal.

```suggestion
  if err != nil {
    return err
  }
  if resp.StatusCode >= 400 {
    return fmt.Errorf("buildkite API request failed: %s %s (status: %d)", method, url, resp.StatusCode)
  }
```

vgrigoruk:
> If I add the `read pipelines` scope, then the acceptance tests and my manual test pass. Do you have the `read pipelines` scope, and if you remove it do you get the same failures?

good catch, I have added an `read pipelines` scope to my API token some time ago when I started working on this PR and completely forgot about it.

> If so, that raises two questions:
> 
> 1. It'd be good to avoid adding the need for an extra scope, given we hope to require only graphql access eventually. If we do need the new scope though, then it'd be good to document that somewhere. I think the existing README and docs list the required permissions for the token

Unfortunately, there `read pipelines` scope is a requirement for making an API call here: https://github.com/peakon/terraform-provider-buildkite/blob/38fecd614a607087f8c1c54b50ad0c24c7a0d77a/buildkite/resource_pipeline.go#L447

The code in `main` branch doesn't require API token to have `read pipelines` scope. This causes a bug, that `branch_configuration` is only written to remote (in create/update methods), but it is never being read from remote (as we don't make `GET` API call). 
This causes a configuration drift as:
- `terraform import` never imports the value of `branch_configuration` into the state.
- `branch_configuration` changes made on the remote are never synced back to terraform state;
-  `terraform apply` doesn't update the remote state for `branch_configuration` due to ^. 

This PR fixes the above bug for `branch_configuration` and also for `provider_settings`. So I will update the docs as you suggested.

> 2. If a missing scope was my issue, it's curious that I didn't get an error and it silently failed. I wonder if we can pinpoint why that is and if we can make it error more loudly?

I'll fix the error handling for this case

buildkite/resource_pipeline_test.go
```diff
@@ -302,6 +317,23 @@ func testAccPipelineConfigComplex(name string, steps string) string {
...
+				cancel_deleted_branch_builds = true
+				prefix_pull_request_fork_branch_names = true
+				publish_blocked_as_pending = true
+				publish_commit_status = true
+				publish_commit_status_per_step = true
+				pull_request_branch_filter_configuration = "features/*"
+				pull_request_branch_filter_enabled = true
+				separate_pull_request_statuses = true
+				skip_pull_request_builds_for_existing_commits = false
+			}
```
yob, line 336: I'm stoked this worked :+1:

yob (approved):
Perfect, thanks for all your work on this  :sparkles: 

I've confirmed the acceptance tests are still green (once I add the read pipelines scope), and that I get a fatal error when the scope is missing.

We'll ensure the changelog mentions the token requirements when we next do a release.

vgrigoruk:
Looking forward to the next release 🚀

PR #125, opened 2021-03-16
Use 'Computed: true' for attributes that are initialized on backend

vgrigoruk:
Added `Compute: true` to all pipeline resource attributes that has default values set by Buildkite (this fixes a lot of warnings that you can see running terraform apply for a with TF_LOG=debug env variable, like:

```
2021/03/09 17:41:15 [WARN] Provider "registry.terraform.io/buildkite/buildkite" produced an unexpected new value for buildkite_pipeline.test, but we are tolerating it because it is using the legacy plugin SDK.
    The following problems may be the cause of any confusing errors from downstream operations:
      - .branch_configuration: was null, but now cty.StringVal("")
```

This is a spin off from https://github.com/buildkite/terraform-provider-buildkite/pull/123

### DIFF
--- a/buildkite/resource_pipeline.go
+++ b/buildkite/resource_pipeline.go
@@ -66,24 +66,27 @@ func resourcePipeline() *schema.Resource {
 
 		Schema: map[string]*schema.Schema{
 			"cancel_intermediate_builds": {
+				Computed: true,
 				Optional: true,
-				Default:  false,
 				Type:     schema.TypeBool,
 			},
 			"cancel_intermediate_builds_branch_filter": {
+				Computed: true,
 				Optional: true,
-				Default:  "",
 				Type:     schema.TypeString,
 			},
 			"branch_configuration": {
-				Type:     schema.TypeString,
+				Computed: true,
 				Optional: true,
+				Type:     schema.TypeString,
 			},
 			"default_branch": {
+				Computed: true,
 				Optional: true,
 				Type:     schema.TypeString,
 			},
 			"description": {
+				Computed: true,
 				Optional: true,
 				Type:     schema.TypeString,
 			},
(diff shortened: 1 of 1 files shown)


yob:
I was vaguely aware of the `Computed` option, but I assumed it was for properties that are provided by the server and can't be controlled by the client (like `webhook_url` and `slug`).

I assume I'm wrong, although to research before testing this PR out I [read the docs](https://www.terraform.io/docs/extend/schemas/schema-behaviors.html). The Computed example is for a non-user controllable property (UUID), and the text says:

> Computed is often used to represent values that are not user configurable or can not be known at time of terraform plan or apply, such as date of creation or a service specific UUID.

However, the text also says:

> Computed can be combined with other attributes to achieve specific behaviors, and can be used as output for interpolation into other resources

I tried this branch out and I was able to set and change `cancel_intermediate_builds` just fine.  I noticed all the properties that have gained `Computed: True` in this PR also have `Optional: True`. Maybe combining `Optional: True` and `Computed: True` allows the use to control the value, but also allows the default to be controlled by the server when the user omits it?

Does `branch_configuration` need `Computed: True` as well?

yob:
In interesting tidbit from [here](https://www.terraform.io/docs/extend/terraform-0.12-compatibility.html):

> The original intent of Computed: true in a schema was to say that a particular attribute has a default value but that default value won't be known until after the object is created.

So using `Computed: True` as proposed by this PR seems totally in line with the original intent by hashicorp :sparkles:

vgrigoruk:
> Maybe combining Optional: True and Computed: True allows the use to control the value, but also allows the default to be controlled by the server when the user omits it?

Exactly, here is where I got this idea from: https://github.com/hashicorp/terraform/issues/21278#issuecomment-491547661

> Does branch_configuration need Computed: True as well?

Good point, I will update my PR

vgrigoruk:
I believe it is ready for review now, @yob

yob:
> Exactly! Here is where I got this idea from: hashicorp/terraform#21278 (comment)

this was super helpful extra context, cheers.

yob (approved):

PR #129, opened 2021-03-30
update docs for the release process

yob:
@chloeruka and I just released v0.3.0 and discovered these docs are slightlywrong.


chloeruka (approved):
Nice! Thanks for fixing that.

PR #136, opened 2021-04-27
Add buildkite_ip_ranges data source

### LINKED ISSUE #105: Data Provider for Egress IPs
Buildkite publishes the IP addresses we use for egress traffic (like webhooks and commit statuses) at this unauthenticated API: https://api.buildkite.com/v2/meta

It'd be neat if the provider included a data source that exposed these IPs so folks could automatically configure firewall rules.

yob:
This adds a new data source for listing the IP addresses that Buildkite uses to send webhooks and other external requests (like commit status API calls to GitHub Enterprise and BitBucket Server).

It takes no parameters, and can be used like this:

```hcl
    data "buildkite_ip_ranges" "ips" { }

    output "cidrs" {
      value = data.buildkite_ip_ranges.ips.webhook_ips
    }
```

This is intended to be used in other terraform projects to dynamically set firewall and ingress rules to allow traffic from Buildkite. For customers that use AWS, that might look like this:

```hcl
    data "buildkite_ip_ranges" "ips" { }

    resource "aws_security_group" "from_buildkite" {
      name = "from_buildkite"

      ingress {
        from_port        = "*"
        to_port          = "443"
        protocol         = "tcp"
        cidr_blocks      = data.buildkite_ip_ranges.ips.webhook_ips
      }
    }
```

Closes #105

### DIFF
--- a/buildkite/data_source_meta.go
+++ b/buildkite/data_source_meta.go
@@ -1,0 +1,51 @@
+package buildkite
+
+import (
+	"context"
+	"sort"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+type MetaResponse struct {
+	WebhookIps []string `json:"webhook_ips"`
+}
+
+func dataSourceMeta() *schema.Resource {
+	return &schema.Resource{
+		ReadContext: dataSourceMetaRead,
+		Schema: map[string]*schema.Schema{
+			"webhook_ips": {
+				Computed: true,
+				Type:     schema.TypeList,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
+		},
+	}
+}
+
+func dataSourceMetaRead(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
+	client := m.(*Client)
+	meta := MetaResponse{}
+	err := client.makeRequest("GET", "https://api.buildkite.com/v2/meta", nil, &meta)
+
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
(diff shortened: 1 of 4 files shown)


buildkite/data_source_ip_ranges_test.go
```diff
@@ -0,0 +1,64 @@
...
+			return err
+		}
+
+		if webhookIpsSize == 0 {
+			return fmt.Errorf("webhook_ips attribute should have more than 0 items (len: %d)", webhookIpsSize)
+		}
+
+		return nil
+	}
+}
```
yob, line 56: I don't want the test to fail if we change the IPs returned by the API, so I'm just doing some basic sanity checks:

1. is there an ID set
2. are there 1 or more strings returned

sj26:
Should the naming be aligned, should we call this resource `buildkite_meta`?

buildkite/data_source_ip_ranges.go
```diff
@@ -0,0 +1,51 @@
...
+	meta := MetaResponse{}
+	err := client.makeRequest("GET", "https://api.buildkite.com/v2/meta", nil, &meta)
+
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	// a consistent order will ensure a change in ordering from the server won't trigger
+	// changes in a terraform plan
+	sort.Strings(meta.WebhookIps)
```
sj26, line 41: Nice catch. Perhaps our API responses should be doing this as well.

yob:
> Should the naming be aligned, should we call this resource buildkite_meta? The intention was that this endpoint might return more meta information about the Buildkite platform over time.

I was following the pattern used in the AWS provider, but aligning the name with our own API makes sense. I'll do the thing.

yob:
I think this is ready for another look @sj26

sj26 (approved):
Looks good!

chloeruka (approved):
Looks good to me!

PR #146, opened 2021-05-12
Document build_branches provider setting

jlisee:
This property is present in the GitHub example earlier in the document
but not explicitly documented.

It's commonly used when setting up pipelines used to validate a specific
branch and pull request based workflows.


yob (approved):
Thanks!

Issue #148, opened 2021-05-14
Is it possible to include the buildbadge .svg file name as part of the buildkite_pipeline resource output


JPScutt:
Hi,

Could the output properties on the "buildkite_pipeline" been extended to include the svg file name? It will enable me to automate setting the build badge in GitHub repo readme files.

Thanks

yob:
@JPScutt this is a nice idea.

It seems the badge URL isn't available via the graphql unfortunately, but it is in the [REST API](https://buildkite.com/docs/apis/rest-api/pipelines).

Ideally we're keen for this provider to only use graphql, however the pipeline resource already falls back to the REST API for a couple of properties, so it wouldn't be an outrageous amount of work.

We'd be happy to help you land a PR if you're up for it?

It'd involve adding a new property to the pipeline schema, possibly near webhook_url (which is also a read-only property): https://github.com/buildkite/terraform-provider-buildkite/blob/e52f6c4610b907bbef63eee44ede9894fd70517f/buildkite/resource_pipeline.go#L243-L246

The `PipelineExtraInfo` would also need to have a new item added for BuildBadge (at the same nesting level as `BranchConfiguration` I think): https://github.com/buildkite/terraform-provider-buildkite/blob/e52f6c4610b907bbef63eee44ede9894fd70517f/buildkite/resource_pipeline.go#L421-L442 

Finally, I think `updatePipelineResourceExtraInfo` would need a new d.Set() that copies the badge url from `PipelineExtraInfo` struct into the schema: https://github.com/buildkite/terraform-provider-buildkite/blob/e52f6c4610b907bbef63eee44ede9894fd70517f/buildkite/resource_pipeline.go#L664

JPScutt:
Okay, I will try and look into this week. Thanks for the pointers.

JPScutt:
@yob I have raised [PR](https://github.com/buildkite/terraform-provider-buildkite/pull/151), I look forward to your feedback.

PR #151, opened 2021-05-21
feat: extended the Pipeline Extra Info to include the badge_url as an output

JPScutt:
Extended the PipelineExtraInfo so that it includes the badge_url as a readonly property. So that it can be returned as an output attribute when a new pipeline is created. 

Resolves [Issue](https://github.com/buildkite/terraform-provider-buildkite/issues/148)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -64,3 +64,4 @@ terraform.rc
 tf-proj/terraform.d/*
 tf-proj/*.tfstate
 tf-proj/*.tfstate.backup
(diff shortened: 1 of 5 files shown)


yob:
lovely :heart:

I've merged with a small addition to the tests added on top: 6333ea132436d476be3f618abebd7e9c3450fb36

yob:
Thanks again for this contribution - we've just released v0.5.0 which includes this PR

Issue #117, opened 2021-03-10
Support specifying pipeline team membership using team id


yob:
Currently we support placing pipelines in a team like this:

```hcl
resource "buildkite_pipeline" "repo2" {
    name = "repo2"
    repository = "git@github.com:org/repo2"
    steps = file("./steps.yml")

    team {
        slug = "everyone"
        access_level = "READ_ONLY"
    }
}
```

Specifying the team via a slug is nice and readable, but it requires an API call to map it to a team graphql ID. We'd like to deprecate the slug, and instead recommend this approach:

```hcl
resource "buildkite_team" "everyone" {
    name = "everyone"
}

resource "buildkite_pipeline" "repo2" {
    name = "repo2"
    repository = "git@github.com:org/repo2"
    steps = file("./steps.yml")

    team {
        id = buildkite_team.everyone.id
        access_level = "READ_ONLY"
    }
}
```

Adding a data source for buildkite_team would be a nice bonus too. We'd need to continue supporting slug for backwards compatibility, but it could be deprecated.

keith:
I hit this but the frustrating part was it seems the error is quite poor, when I try to apply I just see:

```
buildkite_pipeline.iosstaging: Creating...

Error: No ID was supplied
```

vgrigoruk:
> Adding a data source for buildkite_team would be a nice bonus too. We'd need to continue supporting slug for backwards compatibility, but it could be deprecated

Alternatively, you can just make a breaking change and release a new major version of the provider...

yob:
> Error: No ID was supplied

rats, that's really not a very useful error :thinking: 

It's not even useful as a pointer to where in the coed it might be coming from. `No ID was supplied` isn't in this code base, so I'm assuming it might be generated server side and included in a graphql response. Did it occur when the slug was empty or not-valid? Maybe:

```hcl
resource "buildkite_pipeline" "repo2" {
    name = "repo2"
    repository = "git@github.com:org/repo2"
    steps = file("./steps.yml")

    team {
        slug = ""
        access_level = "READ_ONLY"
    }
}
```

> And it doesn't look like ID is shown in the docs https://registry.terraform.io/providers/buildkite/buildkite/latest/docs/resources/team

Oops, thanks for picking this up. I've opened a PR to fix the docs.

> And this example shows using the slug directly https://registry.terraform.io/providers/buildkite/buildkite/latest/docs/resources/pipeline#example-usage

[...]

PR #155, opened 2021-05-25
Return a better user visible error when pipeline references an unrecognised team slug

yob:
In #117 @keith reported that we return a particularly unhelpful error to users when a pipeline references a team slug that doesn't exist:

    resource "buildkite_pipeline" "repo2" {
        name = "repo2"
        repository = "git@github.com:org/repo2"
        steps = file("./steps.yml")

        team {
            slug = "not valid"
            access_level = "READ_ONLY"
        }
    }

The error was:

> Error: No ID was supplied

I'm not 100% sure this error handling is where the error @keith saw was from, but it seems pretty likely. Even if it's not, this seems like a good change to make.

The Printf() is only visible to users when debug mode is on, which will almost never be true. Moving that message to an error that will be displayed to the user (and can be searched in this codebase while debugging) seems like a win.


eleanorakh (dismissed):
Looks like a sensible change! 👍

PR #173, opened 2021-12-16
SUP-106 Add `buildkite_team_member` resource

jradtilbrook:
This PR adds the `buildkite_team_member` resource to manage connecting organisation users to teams.

The basic stuff should work since I've tested it manually although I haven't written unit tests just yet.  
Since teams have a notion of "default role", the role attribute on this resource is optional. However, if you specify it, apply, then remove it and apply it will break. I think we need to store the teams default role or look it up for instances where the role is not supplied.

That leads to the following to do items:

- [x] unit tests
- [x] maybe even integration test in `tf-proj`
- [x] fix bug when specifying then removing role attribute

### DIFF
--- a/buildkite/resource_team_member_test.go
+++ b/buildkite/resource_team_member_test.go
@@ -149,7 +149,7 @@ func testAccTeamMemberConfigBasic(role string) string {
 		resource "buildkite_team_member" "test" {
 		    role = "%s"
 			team_id = buildkite_team.test.id
-			user_id = "VXNlci0tLWRlOTdmMjBiLWJkZmMtNGNjOC1hOTcwLTY1ODNiZTk2ZGEyYQ=="
+			user_id = "VXNlci0tLThkYjI5MjBlLTNjNjAtNDhhNy1hM2Y4LTI1ODRiZTM3NGJhYw=="
 		}
 	`
 	return fmt.Sprintf(config, role)
--- a/buildkite/resource_team_member_test.go
+++ b/buildkite/resource_team_member_test.go
@@ -155,7 +155,7 @@ func testAccTeamMemberConfigBasic(role string) string {
 	return fmt.Sprintf(config, role)
 }
 
-func testAccChecKTeamMemberExists(resourceName string, resourceTeamMember *TeamMemberNode) resource.TestCheckFunc {
+func testAccCheckTeamMemberExists(resourceName string, resourceTeamMember *TeamMemberNode) resource.TestCheckFunc {
 	return func(s *terraform.State) error {
 		resourceState, ok := s.RootModule().Resources[resourceName]
 
--- a/buildkite/resource_team_member_test.go
+++ b/buildkite/resource_team_member_test.go
@@ -1,0 +1,134 @@
+package buildkite
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+)
+
+// Confirm we can add and remove a team member
+func TestAccTeamMember_add_remove(t *testing.T) {
+	var resourceTeamMember TeamMemberNode
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testCheckTeamMemberResourceRemoved,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccTeamMemberConfigBasic("MEMBER"),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					// Confirm the team member exists in the buildkite API
+					testAccChecKTeamMemberExists("buildkite_team_member.test", &resourceTeamMember),
+					// Confirm the team has the correct values in Buildkite's system
+					testAccCheckTeamMemberRemoteValues(&resourceTeamMember, "MEMBER"),
+					// Confirm the team member has the correct values in terraform state
+					resource.TestCheckResourceAttr("buildkite_team_member.test", "role", "MEMBER"),
+				),
+			},
+		},
+	})
+}
+
+func testAccTeamMemberConfigBasic(role string) string {
+	config := `
+		resource "buildkite_team" "test" {
+			name = "acceptance testing"
+			description = "a cool team for testing"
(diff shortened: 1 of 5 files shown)


jradtilbrook:
Turns out there's a bug in the GraphQL mutation that always creates members with a MEMBER role. So we have to work around that.

I've also had to make `role` required otherwise it doesn't work if you want to specify it, then remove it to default it back - it just stays the same.

buildkite/resource_team_member_test.go
```diff
@@ -0,0 +1,238 @@
...
+			name = "acceptance testing"
+			description = "a cool team for testing"
+			privacy = "VISIBLE"
+			default_team = true
+			default_member_role = "MEMBER"
+		}
+		resource "buildkite_team_member" "test" {
+		    role = "%s"
+			team_id = buildkite_team.test.id
+			user_id = "VXNlci0tLWRlOTdmMjBiLWJkZmMtNGNjOC1hOTcwLTY1ODNiZTk2ZGEyYQ=="
```
jradtilbrook, line 152: @yob @jeremybumsted I could do with some ideas on how to handle this.
This ID is my email used for these acceptance tests which is a bit flaky. Should we create a machine account for this or do you have other ideas?

yob: How would a user of the provider find the correct user graphql ID to include here?

Maybe if we know that, we can find a way to replicate it in the acceptance tests?

If the answer is "look in up on buildkite.com" somewhere, that probably won't help in the tests. Would it be possible to add a data source for looking up a users graphql ID?

buildkite/resource_team_member_test.go
```diff
@@ -0,0 +1,238 @@
...
+		}
+	`
+	return fmt.Sprintf(config, role)
+}
+
+func testAccChecKTeamMemberExists(resourceName string, resourceTeamMember *TeamMemberNode) resource.TestCheckFunc {
```
yob, line 158:
```suggestion
func testAccCheckTeamMemberExists(resourceName string, resourceTeamMember *TeamMemberNode) resource.TestCheckFunc {
```

buildkite/resource_team_member_test.go
```diff
@@ -0,0 +1,238 @@
...
+					// Confirm the team has the correct values in Buildkite's system
+					testAccCheckTeamMemberRemoteValues(&resourceTeamMember, "MAINTAINER"),
+				),
+			},
+		},
+	})
+}
+
+// Confirm that this resource can be imported
+func TestAccTeamMember_import(t *testing.T) {
```
yob, line 90: I assume when importing that the user should provide the graphql ID of the team membership record? If so, how would they find that to paste into the import cli command?

yob (commented):
Neat!

I checked this out and tried it. I was able to create a team member record via terraform, see it on buildkite.com, and then delete it via terraform.

The code is nice and easy to read too, and follows the conventions of other resources.

An interesting thing: while testing this PR out I forgot that this new resource can only create team member records for users who are already in this organization. I started by trying to add a user who wasn't in my test org to a team, and got an error: `Error: No user found`. I reckon it's worth making that limitation clear in the docs. This isn't (yet) support for inviting users to an org via terraform.

Speaking of which, is it worth including some really basic docs for the new resource in this PR as well?

buildkite/resource_team_member_test.go
jradtilbrook: Yeah I think a datasource would be good but should be followed up in another PR. I'll create a ticket for that

yob: yer, staying focused in this PR is wise :+1:

buildkite/resource_team_member_test.go
jradtilbrook: Yeah thats right, it's not displayed in the web anywhere like some other things. It'll need to be found through a GraphQL query. I'll create a ticket to follow up with adding an example query to the docs for people to use when importing resources

buildkite/resource_team_member_test.go
jradtilbrook: What should we do for the integration tests here until then do you think? Use a real person a machine user? I think even if we waited for a datasource, we would still need one of these types of users added to the organization in order for the test to work.

buildkite/resource_team_member_test.go
jradtilbrook: The docs for this resource contain an example graphql query to help with this

docs/resources/team_member.md
```diff
@@ -0,0 +1,71 @@
+# Resource: team_member
+
+This resource allows manage team membership for existing organization users.
+
+The user must already be part of the organization to which you are managing team membership. This will not invite a new user to the organization.
```
yob, line 5: 👌

docs/resources/team_member.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Import
+
+Team members can be imported using the GraphQL ID of the membership. Note this is different to the ID of the user.
+
+```
+$ terraform import buildkite_team_member.a_smith VGVhbU1lbWJlci0tLTVlZDEyMmY2LTM2NjQtNDI1MS04YzMwLTc4NjRiMDdiZDQ4Zg==
+```
+
+To find the ID of a team member you are trying to import you can use the GraphQL snippet below. A link to this snippet can also be found at https://buildkite.com/user/graphql/console/c6a2cc65-dc59-49df-95c6-7167b68dbd5d.
```
yob, line 49: Looks good. Ideally we can replace this with a "organization user" data source in the future, and these instructions will become much easier.

jradtilbrook: Yeah definitely!

yob (approved):
I haven't re-run this since my last review, but the added docs and the integration spec change look good :+1:

buildkite/resource_team_member_test.go
dhess: How does one get GraphQL IDs of users in their organization? The example GraphQL query given in the docs can extract an existing team membership GraphQL ID for the purpose of importing existing team members into a team resource, but the docs also note that this is different than the user's GraphQL ID, and I can't find any way to get the latter.

> Yeah I think a datasource would be good but should be followed up in another PR. I'll create a ticket for that

Did you ever create that ticket?

dhess: Answering myself here, this did the trick for us: https://buildkite.com/docs/apis/graphql/graphql-cookbook#search-for-an-organization-member-by-email-address

PR #174, opened 2021-12-20
Bump the terraform sdk to the latest v2 release

yob:
I haven't tested this or read the release notes, so careful testing is advised!

I noticed we have a GitHub security alert about github.com/ulikunitz/xz. That's in our dependency tree via github.com/hashicorp/terraform-plugin-sdk/v2, and updating the sdk will resovle the alert.

Occasionally bumping the TF SDK feels like soemthing we should do anyway, but I'm not sure if there are other consequences of doing so. Does it change the supported versions of terraform for example?

The number of changes to go.sum is impressive, but I think that's OK? The diff is the result of these commands:

    $ go get -u github.com/hashicorp/terraform-plugin-sdk/v
    $ go mod tidy


yob:
This is well out of date now, closing in favour of dependabot occasionally bumping dependencies #233

Issue #179, opened 2022-02-02
[Feature request] Support cluster in buildkite_pipeline resource


kate-syberspace:
The graphql `pipelineCreate` mutation already accepts `cluster`. 

It would be handy to support this in the `buildkite_pipeline` terraform resource as well.

jradtilbrook:
This is something we will definitely be adding to the provider in the future. However, we don't have a timeline on this feature at the moment.

If you are willing to have a go at it and submit a PR we would be more than happy to review and work with you on it

kate-syberspace:
Thanks @jradtilbrook , I've opened a [PR](https://github.com/buildkite/terraform-provider-buildkite/pull/181).

PR #181, opened 2022-02-09
Add cluster_id argument to pipeline resource

kate-syberspace:
Add the `cluster_id` as optional argument to the pipeline resource.

The graphql `pipelineCreateInput` already takes `clusterId` as an input field, and `Cluster` is returned in the `PipelineNode` as well.

Note I was not able to run the acceptance tests locally, even off of the `main` `buildkite/terraform-provider-buildkite` branch. Local unit tests, build and `terraform plan` were successful, however.

Closes [issue 179](https://github.com/buildkite/terraform-provider-buildkite/issues/179).

🙇🏻‍♂️

### DIFF
--- a/buildkite/resource_pipeline.go
+++ b/buildkite/resource_pipeline.go
@@ -287,7 +287,7 @@ func CreatePipeline(ctx context.Context, d *schema.ResourceData, m interface{}) 
 	var mutation struct {
 		PipelineCreate struct {
 			Pipeline PipelineNode
-		} `graphql:"pipelineCreate(input: {cancelIntermediateBuilds: $cancel_intermediate_builds, cancelIntermediateBuildsBranchFilter: $cancel_intermediate_builds_branch_filter, clusterId: $cluster_id, defaultBranch: $default_branch, description: $desc, name: $name, organizationId: $org, repository: {url: $repository_url}, skipIntermediateBuilds: $skip_intermediate_builds, skipIntermediateBuildsBranchFilter: $skip_intermediate_builds_branch_filter, steps: {yaml: $steps}, teams: $teams})"`
+		} `graphql:"pipelineCreate(input: {cancelIntermediateBuilds: $cancel_intermediate_builds, cancelIntermediateBuildsBranchFilter: $cancel_intermediate_builds_branch_filter, defaultBranch: $default_branch, description: $desc, name: $name, organizationId: $org, repository: {url: $repository_url}, skipIntermediateBuilds: $skip_intermediate_builds, skipIntermediateBuildsBranchFilter: $skip_intermediate_builds_branch_filter, steps: {yaml: $steps}, teams: $teams})"`
 	}
 
 	teamsData := make([]PipelineTeamAssignmentInput, 0)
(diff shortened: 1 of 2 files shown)


buildkite/resource_pipeline.go
```diff
@@ -298,15 +306,16 @@ func CreatePipeline(ctx context.Context, d *schema.ResourceData, m interface{})
...
-		"default_branch":                           graphql.String(d.Get("default_branch").(string)),
-		"desc":                                     graphql.String(d.Get("description").(string)),
-		"name":                                     graphql.String(d.Get("name").(string)),
-		"org":                                      orgID,
-		"repository_url":                           graphql.String(d.Get("repository").(string)),
-		"skip_intermediate_builds":                 graphql.Boolean(d.Get("skip_intermediate_builds").(bool)),
-		"skip_intermediate_builds_branch_filter":   graphql.String(d.Get("skip_intermediate_builds_branch_filter").(string)),
-		"steps":                                    graphql.String(d.Get("steps").(string)),
-		"teams":                                    teamsData,
+		"cluster_id":                             graphql.ID(d.Get("cluster_id").(string)),
```
jradtilbrook, line 309: The integration tests are failing when this is present and empty. Our GraphQL API doesn't handle this case. If the `cluster_id` key is present in the mutation input, it expects it to be a valid ID.

Are you able to confirm this behaviour if you run a manual test with and without this field in your terraform?

I'm afraid the quickest solution is probably to handle this case in the provider here somehow - possibly duplicating the mutation code and only including `cluster_id` when it its not empty? Or do you have another idea?

jradtilbrook (changes requested):
Looks good except for one small issue with the GraphQL 👇

buildkite/resource_pipeline.go
kate-syberspace: Ah, thanks for pointing that out.

I ran the manual tests as you suggested (following directions [here](https://github.com/buildkite/terraform-provider-buildkite#manual-testing), and the `terraform plan` works both with and without `cluster_id`. 

I did play with the actual graphql mutation, though and saw what the issue is with passing `cluster_id` when it doesn't exist.

Just pushed up a commit to address that. Lemme know what you think. 🙏

jradtilbrook: That looks like it should do the trick @kate-syberspace. I'll try get it checked thoroughly today.

However, I think we might even be able to fix the mutation from our end to accept null properly (like it should IMO) this week. If we could get a release out with simpler code here would you be happy to wait until early next week?

kate-syberspace: Thanks @jradtilbrook . No rush from my side.  😀

buildkite/resource_pipeline.go
```diff
@@ -364,19 +386,20 @@ func UpdatePipeline(ctx context.Context, d *schema.ResourceData, m interface{})
 	var mutation struct {
 		PipelineUpdate struct {
 			Pipeline PipelineNode
-		} `graphql:"pipelineUpdate(input: {cancelIntermediateBuilds: $cancel_intermediate_builds, cancelIntermediateBuildsBranchFilter: $cancel_intermediate_builds_branch_filter, defaultBranch: $default_branch, description: $desc, id: $id, name: $name, repository: {url: $repository_url}, skipIntermediateBuilds: $skip_intermediate_builds, skipIntermediateBuildsBranchFilter: $skip_intermediate_builds_branch_filter, steps: {yaml: $steps}})"`
+		} `graphql:"pipelineUpdate(input: {cancelIntermediateBuilds: $cancel_intermediate_builds, cancelIntermediateBuildsBranchFilter: $cancel_intermediate_builds_branch_filter, clusterId: $cluster_id, defaultBranch: $default_branch, description: $desc, id: $id, name: $name, repository: {url: $repository_url}, skipIntermediateBuilds: $skip_intermediate_builds, skipIntermediateBuildsBranchFilter: $skip_intermediate_builds_branch_filter, steps: {yaml: $steps}})"`
```
jradtilbrook, line 389: The upstream API changes for this are going to take a while so it needs to be done here.

You've already fixed the create method, but update also needs it. I wasn't able to push to your PR so here is a patch you can hopefully apply to update it

```diff
diff --git a/buildkite/resource_pipeline.go b/buildkite/resource_pipeline.go
index 4789344..0e958fb 100644
--- a/buildkite/resource_pipeline.go
+++ b/buildkite/resource_pipeline.go
@@ -383,27 +383,41 @@ func ReadPipeline(ctx context.Context, d *schema.ResourceData, m interface{}) di
 // UpdatePipeline updates a Buildkite pipeline
 func UpdatePipeline(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
 	client := m.(*Client)
+	var err error
 	var mutation struct {
 		PipelineUpdate struct {
 			Pipeline PipelineNode
-		} `graphql:"pipelineUpdate(input: {cancelIntermediateBuilds: $cancel_intermediate_builds, cancelIntermediateBuildsBranchFilter: $cancel_intermediate_builds_branch_filter, clusterId: $cluster_id, defaultBranch: $default_branch, description: $desc, id: $id, name: $name, repository: {url: $repository_url}, skipIntermediateBuilds: $skip_intermediate_builds, skipIntermediateBuildsBranchFilter: $skip_intermediate_builds_branch_filter, steps: {yaml: $steps}})"`
+		} `graphql:"pipelineUpdate(input: {cancelIntermediateBuilds: $cancel_intermediate_builds, cancelIntermediateBuildsBranchFilter: $cancel_intermediate_builds_branch_filter, defaultBranch: $default_branch, description: $desc, id: $id, name: $name, repository: {url: $repository_url}, skipIntermediateBuilds: $skip_intermediate_builds, skipIntermediateBuildsBranchFilter: $skip_intermediate_builds_branch_filter, steps: {yaml: $steps}})"`
 	}
 	vars := map[string]interface{}{
 		"cancel_intermediate_builds":               graphql.Boolean(d.Get("cancel_intermediate_builds").(bool)),
 		"cancel_intermediate_builds_branch_filter": graphql.String(d.Get("cancel_intermediate_builds_branch_filter").(string)),
-		"cluster_id":                             graphql.ID(d.Get("cluster_id").(string)),
-		"default_branch":                         graphql.String(d.Get("default_branch").(string)),
-		"desc":                                   graphql.String(d.Get("description").(string)),
-		"id":                                     graphql.ID(d.Id()),
-		"name":                                   graphql.String(d.Get("name").(string)),
-		"repository_url":                         graphql.String(d.Get("repository").(string)),
-		"skip_intermediate_builds":               graphql.Boolean(d.Get("skip_intermediate_builds").(bool)),
-		"skip_intermediate_builds_branch_filter": graphql.String(d.Get("skip_intermediate_builds_branch_filter").(string)),
-		"steps":                                  graphql.String(d.Get("steps").(string)),
+		"default_branch":                           graphql.String(d.Get("default_branch").(string)),
+		"desc":                                     graphql.String(d.Get("description").(string)),
+		"id":                                       graphql.ID(d.Id()),
+		"name":                                     graphql.String(d.Get("name").(string)),
+		"repository_url":                           graphql.String(d.Get("repository").(string)),
+		"skip_intermediate_builds":                 graphql.Boolean(d.Get("skip_intermediate_builds").(bool)),
+		"skip_intermediate_builds_branch_filter":   graphql.String(d.Get("skip_intermediate_builds_branch_filter").(string)),
+		"steps":                                    graphql.String(d.Get("steps").(string)),
 	}
 
 	log.Printf("Updating pipeline %s ...", vars["name"])
-	err := client.graphql.Mutate(context.Background(), &mutation, vars)
+
+	// If the cluster_id key is present in the mutation, GraphQL expects a valid ID.
+	// Check if cluster_id exists in the configuration before adding to mutation.
+	if clusterID, ok := d.GetOk("cluster_id"); ok {
+		var mutationWithClusterID struct {
+			PipelineCreate struct {
+				Pipeline PipelineNode
+			} `graphql:"pipelineUpdate(input: {cancelIntermediateBuilds: $cancel_intermediate_builds, cancelIntermediateBuildsBranchFilter: $cancel_intermediate_builds_branch_filter, clusterId: $cluster_id, defaultBranch: $default_branch, description: $desc, id: $id, name: $name, repository: {url: $repository_url}, skipIntermediateBuilds: $skip_intermediate_builds, skipIntermediateBuildsBranchFilter: $skip_intermediate_builds_branch_filter, steps: {yaml: $steps}})"`
+		}
+		vars["cluster_id"] = graphql.ID(clusterID.(string))
+		err = client.graphql.Mutate(context.Background(), &mutationWithClusterID, vars)
+	} else {
+		err = client.graphql.Mutate(context.Background(), &mutation, vars)
+	}
+
 	if err != nil {
 		log.Printf("Unable to update pipeline %s", d.Get("name"))
 		return diag.FromErr(err)
```

jradtilbrook (changes requested):

mhornbacher:
any progress here?

buildkite/resource_pipeline.go
kate-syberspace: Sorry, missed the notification for this. I've just added those changes.

jradtilbrook (approved):
LGTM 🎉

PR #193, opened 2022-04-19
Add allow rebuilds to pipeline resource

### LINKED ISSUE #187: Add support for allow_rebuilds resource
In Buildkite UI  under pipeline Settings --> Builds there is a button called Allow Rebuilds

```
Rebuilds
Allow Rebuilds

Uncheck this option to disable rebuilds entirely.
```

It currently cannot be set in terraform, while all other  variables can be found in https://registry.terraform.io/providers/buildkite/buildkite/latest/docs/resources/pipeline

Ideally just like we can set cancel_intermediate_builds = false/true
we should be able to set 
allow_rebuilds = false/true

jradtilbrook:
This adds an `allow_rebuilds` property to the schema.  
It resolves #187

### DIFF
--- a/buildkite/resource_pipeline.go
+++ b/buildkite/resource_pipeline.go
@@ -12,6 +12,7 @@ import (
 
 // PipelineNode represents a pipeline as returned from the GraphQL API
 type PipelineNode struct {
+	AllowRebuilds                        graphql.Boolean
 	CancelIntermediateBuilds             graphql.Boolean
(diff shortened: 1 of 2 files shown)


jradtilbrook:
This includes commits from #192 so that it doesn't complain about the formatting. That PR should be merged first and then this one probably

jeremybumsted (approved):

Issue #199, opened 2022-05-13
Builds are failing after upgrading to version 0.9.0 with error - "Variable $allow_rebuilds is declared by  but not used"


akshayhacholli:
We are seeing failures after we upgraded to version 0.9.0, downgrading to 0.8.0 is working fine.

Error - 
```
│ Error: Variable $allow_rebuilds is declared by  but not used
│
│   with buildkite_pipeline.custom-job,
│   on main.tf line 1544, in resource "buildkite_pipeline" "custom-job":
│  1544: resource "buildkite_pipeline" "custom-job" {
```

jradtilbrook:
Yep that looks right!

PR #200, opened 2022-05-19
Add allowRebuilds to cluster mutations

jradtilbrook:
This fixes the issue with missing mutation arguments when also using clusters

### DIFF
--- a/buildkite/resource_pipeline.go
+++ b/buildkite/resource_pipeline.go
@@ -347,7 +347,7 @@ func CreatePipeline(ctx context.Context, d *schema.ResourceData, m interface{}) 
 		var mutationWithClusterID struct {
 			PipelineCreate struct {
 				Pipeline PipelineNode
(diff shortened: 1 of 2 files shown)


jayco (commented):
LGTM - have you tested it?

- It won't break things for folks using the current/older version
- It does the thing as expected

jayco (approved):

jradtilbrook:
Closes #199

jradtilbrook:
I think there is an issue with the REST API around clusters that is blocking this. I'm confirming and following up on that before adding tests here

jradtilbrook:
@jayco turns out the error was my bad copy paste. I've fixed it up now and added acceptance tests to verify this is fixed. Can you have another look?

jayco (approved):
Going over this at a high level - LGTM! 🎉

Just a question, not a blocker - I love the tests, are we just testing just the happy path? Do we eventually need to add-in any other paths?

jradtilbrook:
@jayco yeah I think the tests could do with some TLC. We have some work scheduled to go over the testing stuff later. I was just shoehorning stuff in for now and tackling a revamp in a future cycle

PR #205, opened 2022-06-02
GraphQL demo with Khan/genqlient

jradtilbrook:
This is a demo PR to experiment with a new client and showing what it would look like

It'll need some work to move over but I think its must cleaner and much more obvious what is going on in the GraphQL parts


buildkite/graphql/teams.graphql
```diff
@@ -0,0 +1,15 @@
+query getTeam($slug: ID!) {
+  # maybe change the generated type name with @genqlient(typename: "Team")?
+  team(slug: $slug) {
+    defaultMemberRole
+    description
+    id
+    isDefaultTeam
+    membersCanCreatePipelines
+    name
+    # @genqlient(typename: "GenqlientTeamPrivacy")
```
jradtilbrook, line 10: I've done this for now because the generated default clashed with the existing `TeamPrivacy` type already present

genqlient.yaml
```diff
@@ -0,0 +1,9 @@
+schema: schema.graphql
```
jradtilbrook, line 1: Let me know if you think a different directory structure would be better

jeremybumsted (approved):
LGTM!

PR #206, opened 2022-06-09
SUP-191 Generate test coverage

jradtilbrook:
- SUP-191 Add tparse for nicer test output
- SUP-191 Generate coverage and parse with tparse
- SUP-191 Create annotation from test output


jradtilbrook:
Annotation output: 
Screen Shot 2022-06-10 at 10 10 17

PR #208, opened 2022-06-15
Add allow_rebuilds documentation. Grammar.

jmctune:
`allow_rebuilds` was added with [this PR](https://github.com/buildkite/terraform-provider-buildkite/pull/193).

Adds documentation for `allow_rebuilds`.

PR #227, opened 2023-03-14
Update version and referencing

mcncl:
We haven't changed the version for some time, it needs to be so we can add a new tag and include the already written docs for `allow_rebuilds`.

This will update the version to 0.11.1, which is a patch bump. Patch bump because it's an error fix; the errors is inconsistent version updates. This will also allow us to push tag v0.11.1 and provide clear support for allow_rebuilds. Given it has a default of `true` it's not a breaking change and folks will see the diff in their `terraform plan/apply` output if they are using `^` in their versioning.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ terraform {
   required_providers {
     buildkite = {
       source = "buildkite/buildkite"
-      version = "0.5.0"
+      version = "0.11.1"
     }
   }
 }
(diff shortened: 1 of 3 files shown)


mcncl:
This PR added the docs: https://github.com/buildkite/terraform-provider-buildkite/pull/208

We need to release a tag, `v0.11.1` in order to ensure Hashi picks it up.

PR #229, opened 2023-03-15
Update version and tag so that documentation is merged in to provider…

mcncl:
We have had some documentation updates, for example regarding `allow_rebuilds` but these aren't reflected in the provider due to inconsistent `tag` use.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to this project will be documented in this file.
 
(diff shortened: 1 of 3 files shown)


mcncl:
The `tag` should merge to, else I'll have to just push the new tag.

jradtilbrook:
Have you pushed a tag to this commit? I don't think it will move with the merge, it'll probably need another tag afterwards and a whole release process

jradtilbrook (approved):

mcncl:
A tag was pushed yeah, but it doesn't move with branches IIRC

Issue #226, opened 2023-03-06
`buildkite_pipeline` should manage settings on "Builds" page


jfirebaugh:
Settings on the `https://buildkite.com/<org>/<pipeline>/settings/builds` page should be managed by the `buildkite_pipeline` resource:

* [x] Build Skipping
* [x] Rebuilds
* [x] Timeouts
* [ ] Build Retention

jradtilbrook:
Hey @jfirebaugh thanks for raising this. We are looking at implementing this and will update you soon.

jradtilbrook:
Hey @jfirebaugh just giving an update that we have released support for configuring timeouts in [v0.12.2](https://github.com/buildkite/terraform-provider-buildkite/releases/tag/v0.12.2)

Build retention is in the work but a little more complex to implement. We'll let you know when it lands

jradtilbrook:
I'm closing this off now as build retention is not going to be configurable

PR #230, opened 2023-03-22
Organizational settings

mcncl:
## Changes

- `+` Added an Organization data source, this will allow folks to query an org
- `+` Added the Organization resource
- `+` Added in GraphQL queries for the Org additions to use
- `~` Updated the provider to utilise the new resource and data source
- `~` Modified the genqlient file to accept all `.graphql` files
- `~` Adjusted the naming of the `TeamMemberRole` type within the query; looks to have been previously changed manually and caused type errors on re-generation


buildkite/resource_organization_settings.go
```diff
@@ -0,0 +1,101 @@
...
+  d.Set("slug", response.Organization.Slug)
+  d.Set("uuid", response.Organization.Uuid)
+  d.Set("name", response.Organization.Name)
+  d.Set("allowed_api_ip_addresses", apiResponse.OrganizationApiIpAllowlistUpdate.Organization.AllowedApiIpAddresses)
+
+  return diags
+}
+
+// ReadOrganization retrieves a Buildkite organization
+func ReadOrganizationSettings(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
```
jradtilbrook, line 81: We may as well share the read from the datasource and the read from the resource. It should do the same things. And it probably deserves to live in this file and the datasource file can reference this function

buildkite/resource_organization_settings.go
```diff
@@ -0,0 +1,101 @@
...
+  var diags diag.Diagnostics
+  client := m.(*Client)
+
+  response, err := getOrganization(client.genqlient, d.Get("slug").(string))
+
+  assertError(err)
+
+  if response.Organization.Id == "" {
+    return diag.FromErr(fmt.Errorf("organization not found: '%s'", d.Get("slug")))
+  }
```
jradtilbrook, line 65: I think you're doing this because the mutation below requires the org ID and not the slug (which we already have through configuration of the provider). We are already doing this in at least one other place so maybe its time to extract this somewhere else - I'll create another ticket to follow up on this though

tf-proj/main.tf
```diff
@@ -17,6 +17,11 @@ resource "buildkite_pipeline" "test" {
   steps = ""
 }
 
+resource "buildkite_organization_settings" "test_settings" {
+  slug = buildkite_pipeline.test.organization_slug
+  allowed_api_ip_addresses = "0.0.0.0/0"
```
jradtilbrook, line 22: I want this to be a list instead of a space-delimited string so that users of the provider don't have to do the mapping themselves (other providers output CIDRs as a list so if we accept that it improves usability)

jradtilbrook (changes requested):
Theres some errors in the tests and some refactors I think are crucial here

jradtilbrook:
This closes #226

buildkite/resource_organization_settings.go
```diff
@@ -0,0 +1,101 @@
...
+
+    Schema: map[string]*schema.Schema{
+      "id": {
+        Type:     schema.TypeString,
+        Computed: true,
+      },
+      "slug": {
+        Type:     schema.TypeString,
+        Required: true,
+      },
```
jradtilbrook, line 29: We don't actually need this in the schema (and probably shouldn't have it duplicated). Its part of the provider configuration so is accessible in the CRUD functions

buildkite/resource_organization_settings.go
```diff
@@ -38,31 +38,27 @@ func resourceOrganizationSettings() *schema.Resource {
...
 	}
 
 	cidrs := strings.Join(d.Get("allowed_api_ip_addresses").([]string), " ")
-	apiResponse, queryError := setApiIpAddresses(client.genqlient, response.Organization.Id, cidrs)
+	apiResponse, err := setApiIpAddresses(client.genqlient, response.Organization.Id, cidrs)
 
-	assertError(queryError)
+	if err != nil {
+		return diag.FromErr(err)
+	}
```
mcncl, line 57: I don't really agree with the approach to this, given it causes repetition in the code. Consider refactoring if more errors need to be handled in the future.

jradtilbrook: I think the problem with the function you had was that it would still allow continued execution which we don't want. If there's an error, we want to exit early, so you can't really get away without an `if` condition in the main function body

buildkite/resource_organization_settings.go
```diff
@@ -68,6 +68,30 @@ func CreateUpdateDeleteOrganizationSettings(ctx context.Context, d *schema.Resou
...
+	client := m.(*Client)
+
+	response, err := getOrganization(client.genqlient, client.organization)
+
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	if response.Organization.Id == "" {
+		return diag.FromErr(fmt.Errorf("organization not found: '%s'", client.organization))
```
jradtilbrook, line 83: Should we expand on this error message to mention about the IP allow list now?

tf-proj/main.tf
```diff
@@ -2,32 +2,10 @@ terraform {
...
     buildkite = {
       source  = "buildkite/buildkite"
-      version = "0.12.0"
+      version = "0.11.1"
     }
   }
 }
 
 provider "buildkite" {
-}
```
jradtilbrook, line 11: I think leave this stuff in for future testing (we can just play with this file locally and not commit it when testing)

jradtilbrook (approved):

buildkite/resource_organization_settings_test.go
```diff
@@ -0,0 +1,154 @@
...
+					// Confirm that the allowed IP addresses are set correctly in Buildkite's system
+					testAccCheckResourceDisappears(testAccProvider, resourceOrganizationSettings(), "buildkite_organization_settings.let_them_in"),
+				),
+				ExpectNonEmptyPlan: true,
+			},
+		},
+	})
+}
+
+func testAccOrganizationSettingsConfigBasic(ip_addresses []string) string {
```
jradtilbrook, line 106: I think you'll need to iterate over the list to build up a string or maybe json marshal it or something - tests are failing with an incorrect datatype

jradtilbrook (changes requested):

jradtilbrook (approved):

PR #233, opened 2023-03-27
enable dependabot for gomod

yob:
Some of the direct and indirect dependencies in gomod are looking a bit out of date. This configures dependabot to occasionally bump them, but only up to 2 PRs at a time. We want to slowly keep them updated and avoid bitrot, without being overwhelmed by the rate of PRs.


jradtilbrook:
That sounds good to me!

jradtilbrook (approved):

PR #240, opened 2023-04-04
Update Go and the package dependendies

mcncl:


mcncl:
Includes SUP-841 suggestions from old mate Dependabot.

PR #246, opened 2023-04-04
SUP-857 Manually update go getter & deps

mcncl:
Dependabot build is failing, testing manual update instead.


mcncl:
This might require a staggered upgrade path as some differences in the way the decompression works.

jradtilbrook (approved):

Issue #195, opened 2022-04-26
Archive pipeline on destroy


christophermaier:
The [Github Terraform provider](https://registry.terraform.io/providers/integrations/github/latest/docs) has an [archive_on_destroy](https://registry.terraform.io/providers/integrations/github/latest/docs/resources/repository#archive_on_destroy) argument for its [`github_repository` resource](https://registry.terraform.io/providers/integrations/github/latest/docs/resources/repository), which allows you to optionally archive a Github repository when you remove the resource from your Terraform program, rather than completely deleting it.

Since Buildkite pipelines can also be archived, it would be convenient to add similar functionality to the Buildkite Terraform provider. This would be very useful for those that have auditing needs that require pipeline history, artifacts, etc. to be retained.

jradtilbrook:
Ohh this is a really good idea @christophermaier! It probably ties into #177 as well but might need some changes. I'll have a further think/look into this

mcncl:
Closing this issue as completed, support for this is addressed in https://github.com/buildkite/terraform-provider-buildkite/pull/296

PR #248, opened 2023-04-05
SUP-838 Ensure CI passes/fails correctly

jradtilbrook:
I'm not sure if it was originally intentional that it didn't error out early so that we could get the annotation and test analytics stuff still executed. But doing it that way has lost trust in the state of CI so failing fast is preferred

### DIFF
--- a/.buildkite/steps/annotate.sh
+++ b/.buildkite/steps/annotate.sh
@@ -1,4 +1,5 @@
-#!/bin/bash -e
+#!/bin/bash
+set -ueo pipefail
 
 go test -v -cover -json ./... | tee test_output
 


mcncl (commented):
It might be great to see a test of this? Perhaps have something with a failing test and ensure this step doesn't still pass?

jradtilbrook:
Yeah I was trying to find one of the recent ones I've seen that have actually failed but the step is green. But I think with the squash merging I can't access any commits from recent PRs (and obviously, the merge commit passes). 

But actual test failures were properly failing I think. It was more the panics or some other issue I'm not quite sure about that was actually causing the false positive.

Do you have any ideas on how to test this will work? Otherwise, I was just thinking of waiting for another occurrence and then run this branch against that change to confirm it fails 🤷‍♂️

mcncl:
@jradtilbrook my assumption is that this should be fine; it can't be _less_ helpful than it is now. I think we can merge in. I'm working on https://github.com/buildkite/terraform-provider-buildkite/issues/195 at the moment and that will definitely have some tests in, perhaps I can include a purposefully failing one on that branch.

mcncl (approved):

PR #256, opened 2023-04-17
Set User-Agent header in client

danstn:
This sets the user agent for api client.

Using [Provider.UserAgent](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2@v2.26.1/helper/schema#Provider.UserAgent) to set the header.

Also, this can now take a `version` coming from ldflags which is added as a suffix.

Rendered example:
```
Terraform/ (+https://www.terraform.io) Terraform-Plugin-SDK/2.0.3 buildkite/v0.14.0-1-g3539448
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.20@sha256:f7099345b8e4a93c62dc5102e7eb19a9cdbad12e7e322644eeaba355d70e616d
+FROM 445615400570.dkr.ecr.us-east-1.amazonaws.com/ecr-public/docker/library/golang:1.20@sha256:f7099345b8e4a93c62dc5102e7eb19a9cdbad12e7e322644eeaba355d70e616d
 
 RUN apt-get update \
     && apt-get install -y unzip \
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 default: build
 
 build:
-	go build -o terraform-provider-buildkite .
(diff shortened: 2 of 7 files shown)


jradtilbrook (dismissed):
This looks like a good addition so we can track usage of versions. 
Some tests are failing though

yob:
Yesterday we had an acceptance spec failure caused by the test org not being cleaned properly, and I manually deleted a lingering pipeline. It's possible rebasing this on main (to get the merged GET body fix) and triggering a new build might Just Work.

mcncl (approved):

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.20@sha256:f7099345b8e4a93c62dc5102e7eb19a9cdbad12e7e322644eeaba355d70e616d
+FROM 445615400570.dkr.ecr.us-east-1.amazonaws.com/ecr-public/docker/library/golang:1.20
```
yob, line 1: We should keep the digest if we can, to ensure the image doesn't change if upstream mutates the tag.

jradtilbrook (approved):

mcncl (approved):

yob:
I don't think it's a deal breaker in the short term, but looking at the user agents we're logging in production now it seems like maybe this only sets it for REST API calls, and not for graphql ones?

jradtilbrook:
Actually yes @yob that is correct. I'll create another task to do this for graphql as well

PR #259, opened 2023-04-18
Set User-Agent header on both REST and graphql requests

yob:
In #256 we starting setting a custom User-Agent header on REST API calls, to help us understand the ways our API is used. It looks like this:

    Terraform/ (+https://www.terraform.io) Terraform-Plugin-SDK/2.0.3 buildkite/0.15.0

This builds on that work and sets the same User-Agent on graphql API requests as well.

It drops the unnecessary use of golang.org/x/oauth2 to set a static bearer token header, and uses the [tripperware pattern](https://dev.to/stevenacoffman/tripperwares-http-client-middleware-chaining-roundtrippers-3o00) to set both headers we care about:

1. the static bearer token for auth
2. the user agent

The same http.Client is used for both REST and graphql, so we'll get identical behaviour on both APIs.

### DIFF
--- a/buildkite/client.go
+++ b/buildkite/client.go
@@ -2,7 +2,6 @@ package buildkite
 
 import (
 	"bytes"
-	"context"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -11,7 +10,6 @@ import (
 
 	genqlient "github.com/Khan/genqlient/graphql"
 	"github.com/shurcooL/graphql"
-	"golang.org/x/oauth2"
 )
 
 // Client can be used to interact with the Buildkite API
@@ -21,7 +19,6 @@ type Client struct {
 	http         *http.Client
 	organization string
 	restUrl      string
-	userAgent    string
 }
 
 type clientConfig struct {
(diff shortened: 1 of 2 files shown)


yob:
@jradtilbrook @mcncl absolutely no hurry on this one, I just thought I'd poke at it while the changes in #256 were fresh in our brains. If we merge it, it could wait for the next release you do for other reasons.

buildkite/client.go
```diff
@@ -32,21 +29,55 @@ type clientConfig struct {
...
+	if next == nil {
+		next = http.DefaultTransport
+	}
+	return &headerRoundTripper{
+		next:   next,
+		Header: Header,
+	}
+}
+
+func (rt *headerRoundTripper) RoundTrip(req *http.Request) (resp *http.Response, err error) {
```
danstn, line 72: Named returns are often considered harmful, also not used here.

```suggestion
func (rt *headerRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
```

buildkite/client.go
```diff
@@ -32,21 +29,55 @@ type clientConfig struct {
...
 
+func newHeaderRoundTripper(next http.RoundTripper, Header http.Header) *headerRoundTripper {
+	if next == nil {
+		next = http.DefaultTransport
+	}
+	return &headerRoundTripper{
+		next:   next,
+		Header: Header,
+	}
+}
```
danstn, line 70: Linter yelling continues :) Local bindings [should be](https://go-critic.com/overview.html#captlocal) lower-case.

```suggestion
func newHeaderRoundTripper(next http.RoundTripper, header http.Header) *headerRoundTripper {
	if next == nil {
		next = http.DefaultTransport
	}
	return &headerRoundTripper{
		next:   next,
		Header: header,
	}
}
```

buildkite/client.go
```diff
@@ -32,21 +29,55 @@ type clientConfig struct {
...
+
+func (rt *headerRoundTripper) RoundTrip(req *http.Request) (resp *http.Response, err error) {
+	if rt.Header != nil {
+		for k, v := range rt.Header {
+			req.Header[k] = v
+		}
+	}
+	return rt.next.RoundTrip(req)
```
danstn, line 78: `RoundTrip` returns an error, so normally we'd want to wrap it to provide additional context.

danstn (dismissed):
Great idea @yob! 

There are a few minor comments, but I think it's safe to skip these and add https://github.com/golangci/golangci-lint in a separate PR instead.

mcncl:
Happy to merge once @danstn suggestions are committed in. Also happy to just commit them from comment too.

danstn:
> Happy to merge once @danstn suggestions are committed in. Also happy to just commit them from comment too.

Oh these are not blocking, happy to land it as is. @yob up to you

yob:
They're good suggestions, I'll make them in a bit :heart:

mcncl (approved):
LGTM

buildkite/client.go
yob: I've address the other feedback, but left this one. I could wrap the error, but... I really wasn't sure what to add that was meaningful.

yob:
I've confirmed that we see the Terraform useragent in our API access logs for both the REST **and** graphql calls

PR #261, opened 2023-04-20
Update terraform SDKv2 to the latest version

yob:
According to [terraform docs](https://developer.hashicorp.com/terraform/plugin/sdkv2), SDKv2 is deprecated in favour of [framework](https://developer.hashicorp.com/terraform/plugin/framework). Eventually I assume we'll need to update to framework, but for now lets at least get to a recent version of SDKv2.

The [changelog for SDKv2](https://github.com/hashicorp/terraform-plugin-sdk/blob/v2.26.1/CHANGELOG.md) says v2.0.3 was released September 2020, v2.26.1 in March 2023. The list of changes is long, but nothing is called out as breaking.

v2.0.4 does have a change that caused our AgentToken acceptance specs to fail:

> Fix a bug that would pass the post-destroy state to
> helper/resource.TestCase.CheckDestroy instead of the documented
> pre-destroy state (https://github.com/hashicorp/terraform-plugin-sdk/issues/591)

It turns out that our CheckDestroy function was silently failing to confirm our acceptance tests were revoking Agent Registration Tokens when the resource was destroyed. Luckily, the tokens were being revoked correctly so the fix is to confirm that accurately in test code, not to change the code under test.

This is an alternative to #234, because I had to dig into what was causing the acceptance test failures.


jradtilbrook (approved):
I think this makes sense and encompasses #234 anyway. Good to go ahead IMO

mcncl:
Already merged but massively in favour of this and it's a prerequisite to migrate to framework.

PR #262, opened 2023-04-20
Run go mod tidy

yob:
In #261 I noticed that our go.mod had **three** require blocks. Apparently since 1.17 it's normal to have 2 requires - one for direct dependencies and one for indirect. Three isn't a problem, and is probably a result of go being conservative about modifying the file.

Still, no need to keep all the noise around for what is a relatively simple project. I've go mod tidy to merge the two indirect require blocks and remove old entries from go.sum.

Branches off #261, so I'll leave this as a draft until that merges.


yob:
😂

jradtilbrook (approved):

jradtilbrook (approved):

PR #263, opened 2023-04-21
SUP-819 Retrieve org id only once in provider configuration

jradtilbrook:
To improve performance, move the retrieval of organization ID to the provider configuration/startup instead of each time it is required by a resource operation


mcncl (approved):
LGTM

PR #264, opened 2023-04-21
SUP-820 Pull latest 1.4 version terraform from docker image

jradtilbrook:
Update the terraform version used in CI to 1.4

This is implemented through multistage build for caching performance

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,4 +5,4 @@ RUN apt-get update \
     && go install github.com/mfridman/tparse@latest \
     && go install github.com/lox/buildkite-test-analytics-go@latest
 
-COPY --from=hashicorp/terraform:1.4 /bin/terraform /usr/local/bin/terraform
+COPY --from=hashicorp/terraform:1.4@sha256:5ca7188b7566703fd96c6f84c27d8e7aa4fe1c690803157264b8569580a99712 /bin/terraform /usr/local/bin/terraform
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,9 +5,4 @@ RUN apt-get update \
     && go install github.com/mfridman/tparse@latest \
     && go install github.com/lox/buildkite-test-analytics-go@latest
 
-ENV TERRAFORM_VERSION=0.14.11
-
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -5,9 +5,4 @@ RUN apt-get update \
     && go install github.com/mfridman/tparse@latest \
     && go install github.com/lox/buildkite-test-analytics-go@latest
 
-ENV TERRAFORM_VERSION=0.14.11
```
yob, line 8: wowee, ye olde version

jradtilbrook:
@mcncl how do you think we should implement this? In the multistage build I've done now or using the existing method to download binary explicitly?

I think if we do it the multistage way, we could use the pull-through cache again. @yob could you weigh in as well?

Dockerfile
jradtilbrook: Very!!

mcncl:
I use TF 1+ for everything and never use this Dockerfile, so it's reasonably safe to assume we can bump this to the latest stable

jradtilbrook:
I'm very surprised to see this works straight away. Maybe the other recent upgrades we've done has had an impact

mcncl:
@jradtilbrook it's always Just Worked™ with 1+ for me

Dockerfile
```diff
@@ -5,9 +5,4 @@ RUN apt-get update \
...
-
-RUN cd /tmp && \
-    curl -O https://releases.hashicorp.com/terraform/${TERRAFORM_VERSION}/terraform_${TERRAFORM_VERSION}_linux_amd64.zip && \
-    unzip terraform_${TERRAFORM_VERSION}_linux_amd64.zip && \
-    cp terraform /usr/local/bin/terraform
+COPY --from=hashicorp/terraform:1.4 /bin/terraform /usr/local/bin/terraform
```
jradtilbrook, line 8: @mcncl @yob are you okay with this approach?

An added benefit would be that dependabot could help us keep up to date with v1.5+ as well

mcncl (approved):

PR #265, opened 2023-04-27
SUP-707: Pipeline resource's steps made optional, default pipeline upload step  

james2791:
Amended the Pipeline resource's step parameter to be Optional with a default of `buildkite-agent pipeline upload` if not defined in a to-be-made pipeline resource

Utilising a pipeline.tf file like below:

```
resource "buildkite_pipeline" "terraform-bk-example" {
    name = "terraform-bk-example"
    repository = "git@github.com:james2791/terraform-bk-example.git"
    default_branch = "main"
}
```
Pipeline resource creation

image

Pipeline creation with the default step:

image

Acceptance tests created (one with no default steps which gets filled, the other with steps defined) - running against my own org/API key

image

### DIFF
--- a/buildkite/resource_pipeline.go
+++ b/buildkite/resource_pipeline.go
@@ -153,7 +153,7 @@ func resourcePipeline() *schema.Resource {
 			},
 			"steps": {
 				Optional: true,
-				Default:  "buildkite-agent pipeline upload",
+				Default: "    steps:\n    - label: ':pipeline: Pipeline Upload'\n      command: buildkite-agent pipeline upload\n",
 				Type:     schema.TypeString,
 			},
 			"team": {
--- a/buildkite/resource_pipeline.go
+++ b/buildkite/resource_pipeline.go
@@ -10,6 +10,10 @@ import (
 	"github.com/shurcooL/graphql"
 )
 
+const DefaultSteps = `steps:
+- label: ':pipeline: Pipeline Upload'
+  command: buildkite-agent pipeline upload`
+
 // PipelineNode represents a pipeline as returned from the GraphQL API
 type PipelineNode struct {
 	AllowRebuilds                        graphql.Boolean
@@ -153,9 +157,7 @@ func resourcePipeline() *schema.Resource {
 			},
 			"steps": {
 				Optional: true,
-				Default: `steps:
-				- label: ':pipeline: Pipeline Upload
-				  command: buildkite-agent pipeline upload`,
+				Default:  DefaultSteps,
 				Type:     schema.TypeString,
 			},
 			"team": {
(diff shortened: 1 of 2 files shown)


buildkite/resource_pipeline.go
```diff
@@ -152,7 +152,8 @@ func resourcePipeline() *schema.Resource {
 				Type:     schema.TypeString,
 			},
 			"steps": {
-				Required: true,
+				Optional: true,
+				Default: "    steps:\n    - label: ':pipeline: Pipeline Upload'\n      command: buildkite-agent pipeline upload\n",
```
jradtilbrook, line 156: I'd hoped it would be simple 👍
I wonder if using a `const` and/or multiline string would be nicer https://www.calhoun.io/6-tips-for-using-strings-in-go/

jradtilbrook (changes requested):
LGTM, just got a failing `go fmt` in CI. You should be able to run `make fmt` to fix it

buildkite/resource_pipeline.go
james2791: Shall check that option out - TY @jradtilbrook 👍

mcncl (changes requested):
Could we have test(s) written for this?

buildkite/resource_pipeline_test.go
```diff
@@ -520,6 +567,27 @@ func testAccPipelineConfigBasicWithTimeouts(name string) string {
...
+func testAccPipelineConfigBasicWithSteps(name string, steps string) string {
+	config := `
+		resource "buildkite_pipeline" "foobar" {
+		    name = "Test Pipeline %s"
+		    repository = "https://github.com/buildkite/terraform-provider-buildkite.git"
+			steps = %s
+		}
+	`
+	return fmt.Sprintf(config, name)
+}
```
mcncl, line 589: Doesn't look like this is getting used?

james2791: Tweaking - TY @mcncl 👍

buildkite/resource_pipeline.go
```diff
@@ -10,6 +10,10 @@ import (
 	"github.com/shurcooL/graphql"
 )
 
+const DefaultSteps = `steps:
```
jradtilbrook, line 13: I think this const shouldnt be exported from the package. Should we rename it to `defaultSteps`?

jradtilbrook (dismissed):
Looks good. Only one non-blocking comment left

buildkite/resource_pipeline.go
james2791: That is fair - will adjust

mcncl (dismissed):

buildkite/resource_pipeline.go
james2791: Changed to const, file scope

jradtilbrook (approved):

PR #266, opened 2023-04-28
Update the paths where we load environment variables from

yob:
For consistency with other pipelines, we're prefixing all ENV var paths with the org slug in addition to the pipeline slug - note the `/buildkite` inserted into each path here.


mcncl (approved):
Looks like a sensible change.

jradtilbrook (approved):
Looks good 👍

But either the vars don't exist at the new paths yet or the IAM role is not allowed to read from that path (or maybe both).

That shouldn't require a change to this PR though so approving this now assuming it will be fixed and the build re-run to get to green before merging

jradtilbrook:
Oh yes I see the PR to fix IAM now 👍

yob:
Sorry, I could've got the sequencing a little smoother 🙏

PR #271, opened 2023-05-17
DOCS fmt

mattclegg:
Improve readability of docs

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -12,7 +12,7 @@ max_line_length = 120
 [*.go]
 indent_style = tab
 
-[*.{workflow,yaml,yml}]
+[*.{tf,workflow,yaml,yml}]
 indent_size = 2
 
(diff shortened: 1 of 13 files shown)


jradtilbrook:
Thanks for this @mattclegg 👍 
I didn't think there was any kind of standard for terraform but this made me actually go looking. And I found https://developer.hashicorp.com/terraform/language/syntax/style which says 2 spaces.

Can you update https://github.com/buildkite/terraform-provider-buildkite/blob/main/.editorconfig#LL15C19-L15C19 to add in `.tf` files as well and then I think we can merge this

mcncl (approved):

mcncl:
I've approved these changes and the tests have passed, I'll also merge it in.

@mattclegg, given this is an aesthetic/readability change, it'll be included in the next release rather than triggering a new one for it.

PR #272, opened 2023-05-23
Add PR/issue templates

jradtilbrook:


jradtilbrook:
Let me know if I've missed any bits we need to remember to update for PRs/releases

james2791:
💪

`make fmt` in the PR template?

jradtilbrook:
> 💪
> 
> `make fmt` in the new feature template?

@james2791 CI will fail anyway and won't be able to merge it and I think thats a nicer way. If all these steps could be codified into CI that would be even better!

james2791:
> > 💪
> > `make fmt` in the new feature template?
> 
> @james2791 CI will fail anyway and won't be able to merge it and I think thats a nicer way. If all these steps could be codified into CI that would be even better!

Thats fair - LGTM then 👍

james2791 (dismissed):
👍

lizrabuya (dismissed):
Nice! This is very helpful.

.github/PULL_REQUEST_TEMPLATE/new_feature.md
```diff
@@ -0,0 +1,3 @@
+## PR checklist:
+- [ ] `docs/` updated
+- [ ] tests added
```
mcncl, line 3: I think it would be great to include examples, especially where a new feature has been added of the functionality behind it has changed.

Perhaps an updated `main.tf` to show the change quickly for reference but also stop that from getting stale.

mcncl (changes requested):

.github/PULL_REQUEST_TEMPLATE/new_feature.md
jradtilbrook: Yeah I like that. I'll add another item.

I also wonder if we could make better use of the docs that end up on the terraform registry to include examples there. But I'll leave that for another time I think

jradtilbrook:
```suggestion
- [ ] `docs/` updated
- [ ] tests added
- [ ] an example added to `examples/` (useful to demo new field/resource
```

mcncl (approved):

.github/PULL_REQUEST_TEMPLATE/new_feature.md
```diff
@@ -0,0 +1,4 @@
...
+- [ ] an example added to `examples/` (useful to demo new field/resource
```
jradtilbrook, line 4:
```suggestion
- [ ] an example added to `examples/` (useful to demo new field/resource)
```

mcncl (approved):

PR #274, opened 2023-05-25
Use gotestsum for running and reporting tests

jradtilbrook:
I recently came across [gotestsum](https://github.com/gotestyourself/gotestsum) which is much easier to use compared to tparse, etc. So this converts our acceptance tests over to that.

I've also converted to the test collector plugin instead of the other go tool we were using as that is nicer to use as well

While I was here, I couldn't help converting the pipeline datasource to genqlient as well


james2791 (approved):
Looks awesome!

mcncl (approved):

lizrabuya (approved):

PR #276, opened 2023-05-26
SUP-995: Changelog updates, v0.18.0 release prep

james2791:
Bumping and aligning versions to the changelog with past history. Also prepping for the `v0.18.0` release

- [x] CHANGELOG.md updated
- [x] README.md version examples updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 All notable changes to this project will be documented in this file.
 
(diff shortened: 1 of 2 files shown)


mcncl:
@james2791 docs/ updated is checked, is that missing from the commit or was it not required? If the latter then it doesn't need to be checked.

james2791:
> @james2791 docs/ updated is checked, is that missing from the commit or was it not required? If the latter then it doesn't need to be checked.

Copypasta from the to-be-coming PR template - un-needed, check now 👍

jradtilbrook (approved):

Issue #213, opened 2022-08-29
removing cluster_id from a pipeline resource does not remove cluster from pipeline


albertywu:
After making this change to remove the cluster_id from a pipeline resource:

<img width="1210" alt="Screen Shot 2022-08-27 at 9 29 48 AM (1)" src="https://user-images.githubusercontent.com/1920766/187250160-6117d839-c45f-4748-995e-697e2b435086.png">

The cluster was not actually removed from the pipeline after running "terraform apply"

albertywu:
I wonder if this issue is related to this one? https://github.com/buildkite/terraform-provider-buildkite/issues/212

cc: @jradtilbrook

jradtilbrook:
#279 is now merged and we will release a new version next week. I'm going to close this out now ahead of that

PR #279, opened 2023-06-08
SUP-863 Allow pipeline to be removed from a cluster

jradtilbrook:
## PR checklist:
- [x] `docs/` updated
- [x] tests added


jradtilbrook:
Closes #213

james2791 (approved):
Nice ☝️

buildkite/resource_pipeline.go
```diff
@@ -15,46 +15,43 @@ const defaultSteps = `steps:
...
 // PipelineNode represents a pipeline as returned from the GraphQL API
+type Cluster struct {
+	ID graphql.String
+}
+type Repository struct {
+	URL graphql.String
+}
+type Steps struct {
+	YAML graphql.String
+}
```
mcncl, line 26: Is there a reason to _not_ use the `bindings` option in Genqlient for these?

mcncl (dismissed):

buildkite/resource_pipeline.go
jradtilbrook: Oh yeah these are for the shurcool client. So I wasn't too concerned with doing it a nice way since we'll get rid of it shortly. I just couldn't get it working unless I had a dedicated struct to also use elsewhere to map between the shurcool and genqlient types

mcncl: Yeah, OK, my understanding of `bindings` is that it'll create the type for you when you run the generation, but it's a non-blocker.

mcncl (approved):
Huge

PR #280, opened 2023-06-08
SUP-202 Add graphql example queries for finding import IDs

jradtilbrook:
## PR checklist:
- [x] `docs/` updated
- [x] tests added


docs/resources/team.md
```diff
@@ -33,3 +33,28 @@ resource "buildkite_team" "team" {
...
+
+## Import
+
+Teams can be imported using the `GraphQL ID` (not UUID), e.g.
+
+```
+$ terraform import buildkite_team.fleet UGlwZWxpbmUtLS00MzVjYWQ1OC1lODFkLTQ1YWYtODYzNy1iMWNmODA3MDIzOGQ=
+```
+
+To find the ID to use, you can use the GraphQL query below. Alternatively, you could use this [pre-saved query](https://buildkite.com/user/graphql/console/9cfcca03-df1a-4951-9fa9-a7345aa3daaf2), where you will need fo fill in the organization slug and search term (TEAM_SEARCH_TERM) for the team.
```
jradtilbrook, line 45: I think this saved query URL isn't correct. When I navigate, it shows me this query:

```graphql
query pipelineId {
  pipeline(slug: "bk-testing") {
    id
  }
}
```

docs/resources/organization_settings.md
```diff
@@ -25,3 +25,5 @@ Organization settings can be imported by passing the organization slug to the im
 ```
 $ terraform import buildkite_organization_settings.test_settings test_org
 ```
+
+Your organization's slug can be found in your organisation's [settings](https://buildkite.com/organizations/settings) page.
```
jradtilbrook, line 29: I think use this dynamic URL: `https://buildkite.com/organizations/~/settings` which redirects to the active org for the user

jradtilbrook: Actually, we should probably improve the code behind this so that anything can be passed as the slug. That's because, to properly run any terraform command, we need the org slug to be configured for the provider anyway so its redundant. The problem is though that `terraform import` requires an ID to be passed.

🤷‍♂️ I guess it doesn't really matter requiring them to pass the slug, but something that does get at me is that almost everything else is graphql ID and I think that should be consistent. That is definitely out of scope here though

docs/resources/pipeline_schedule.md
```diff
@@ -38,3 +38,43 @@ Pipeline schedules can be imported using a slug (which consists of `$BUILDKITE_O
...
+                uuid
+                cronline
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+}
```
jradtilbrook, line 79: I think all of this should be one query since it only builds on top of the above one, it could be merged together

jradtilbrook (commented):
I'm not sure if this is ready for review as its in draft but here is a review anyway (also, since I created this PR, I can only comment. No approval or change requests are allowed)

docs/resources/organization_settings.md
james2791: That is a fair call - I've put the dynamic URL into the latest commit but worth assessing 👍

mcncl (dismissed):

docs/resources/pipeline_schedule.md
```diff
@@ -38,3 +38,28 @@ Pipeline schedules can be imported using a slug (which consists of `$BUILDKITE_O
 ```
 $ terraform import buildkite_pipeline_schedule.test myorg/test/1be3e7c7-1e03-4011-accf-b2d8eec90222
 ```
+
+Your organization's slug can be found in your organisation's [settings](https://buildkite.com/organizations/~/settingss) page. 
+
+The pipeline slug and its relevant schedule UUID can be found with the GraphQL query below. Alternatively, you could use this [pre-saved query](https://buildkite.com/user/graphql/console/abf9270e-eccf-4c5f-af21-4cd35164ab6c), specifying the organisation slug (when known) and the pipeline search term (PIPELINE_SEARCH_TERM).
+
+query getPipelineScheduleUuid {
```
jradtilbrook, line 46: This needs '```graphql' above it

jradtilbrook: And the indenting of the query look off

jradtilbrook (commented):
A few tweaks and then all good 👍

james2791 (approved):
💚

PR #281, opened 2023-06-09
SUP-1040: Agent Token resource genqlient migration & adjustment

james2791:
## PR checklist:
- [x] `docs/` updated
- [x] tests added


buildkite/graphql/agent_token.graphql
```diff
@@ -0,0 +1,39 @@
+query getAgentToken($slug: ID!) {
+    agentToken(slug: $slug) {
+        id
+        description
+        token
```
jradtilbrook, line 5: May as well remove this field from the query. It will always be empty
```suggestion
```

jradtilbrook (changes requested):

buildkite/resource_agent_token.go
```diff
@@ -48,82 +46,74 @@ func resourceAgentToken() *schema.Resource {
...
+// DeleteToken revokes a Buildkite agent token - they cannot be completely deleted (will have a revoke)
 func DeleteToken(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
+	var diags diag.Diagnostics
 	client := m.(*Client)
-	var mutation struct {
-		AgentTokenRevoke struct {
-			AgentToken AgentTokenNode
-		} `graphql:"agentTokenRevoke(input: {id: $id, reason: $reason})"`
+
+	agentTokenRevCandidate, err := getAgentToken(client.genqlient, fmt.Sprintf("%s/%s", client.organization, d.Get("uuid").(string)))
```
jradtilbrook, line 98: I'm curious why you are getting the token here first. Can you explain the reasoning?

james2791: With the delete fn - its more I think so that agentTokenRevCandidate can be had without the contextual d.Id()

jradtilbrook: Oh right okay. No the `d.Id()` will be in state so that can be used directly and we save on an extra API request. So I think you can remove a few lines here and jump straight to `revokeAgentToken()`

CHANGELOG.md
```diff
@@ -4,8 +4,9 @@ All notable changes to this project will be documented in this file.
 
 ## Unreleased
 
-* Allow pipeline to be removed from a cluster [[PR #279](https://github.com/buildkite/terraform-provider-buildkite/pull/279)]
-* Change default provider settings to match new pipeline [[PR #282](https://github.com/buildkite/terraform-provider-buildkite/pull/282)] 
+* Allow pipeline to be removed from a cluster [[PR #279](https://github.com/buildkite/terraform-provider-buildkite/pull/279)] @jradtilbrook
+* Change default provider settings to match new pipeline [[PR #282](https://github.com/buildkite/terraform-provider-buildkite/pull/282)] @jradtilbrook
+* Agent Token resource genqlient migration & adjustment [[PR #281](https://github.com/buildkite/terraform-provider-buildkite/pull/281)] @james2791
```
jradtilbrook, line 9: Do you want to format all this up in this PR or follow up with a release PR? I think it will be good to have a section explaining further the changes to agent tokens and linking to https://buildkite.com/changelog/207-agent-token-being-deprecated-from-graphql-apis

jradtilbrook (approved):

CHANGELOG.md
james2791: Release PR :)

PR #282, opened 2023-06-09
SUP-1043 Change default provider settings to match new pipeline

jradtilbrook:
When creating a new pipeline in the UI, these settings default to on. We should have the terraform provider do the same thing.

The specific settings are:
Screenshot 2023-06-09 at 15 04 25

And that maps to the `build_pull_requests`, `skip_pull_request_builds_for_existing_commits`, `build_branches`, `publish_commit_status`.

## PR checklist:
- [x] `docs/` updated
- [x] tests added


jradtilbrook:
I don't think we need to call it out in the docs. But mention it in the changelog when this is released at least

mcncl (approved):

PR #283, opened 2023-06-09
Add another checkbox to PR templates

jradtilbrook:
I think it might be a good idea to add entries to the changelog with each feature PR created so that the release process is a little easier and the changelog can just have some minor tweaks added


james2791 (approved):
100%!

PR #284, opened 2023-06-14
Spike converting to framework

jradtilbrook:
This spikes out upgrading the provider to the terraform plugin framework from SDKv2

Docs on this are at: https://developer.hashicorp.com/terraform/plugin/framework/migrating

With just these changes, we are running our old code through the new platform. Unfortunately, until we configure it more and migrate resources/datasources over, we can't update any tests.

But I've run the tf from `tf-proj/main.tf` against the build output of this branch and it works without errors. It's not extensively tested but I think that validates enough that the provider is working through this different means of instantiation and that this is safe to merge.

We can follow up with other migrations


mcncl (approved):
Looks fine to me.

james2791 (approved):
LGTM

PR #286, opened 2023-06-16
SUP-1072 Create new provider using framework plugin

jradtilbrook:
## PR checklist:
- [x] `docs/` updated
- [x] tests added

### DIFF
--- a/buildkite/provider.go
+++ b/buildkite/provider.go
@@ -3,11 +3,13 @@ package buildkite
 import (
 	"context"
 	"fmt"
+	"os"
 
 	"github.com/hashicorp/terraform-plugin-framework/datasource"
 	"github.com/hashicorp/terraform-plugin-framework/provider"
 	framework_schema "github.com/hashicorp/terraform-plugin-framework/provider/schema"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/types"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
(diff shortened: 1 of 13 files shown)


james2791 (dismissed):
LGTM 👍

buildkite/provider.go
```diff
@@ -16,6 +25,108 @@ const (
...
+	return []func() resource.Resource{}
+}
+
+func (*terraformProvider) Schema(ctx context.Context, req provider.SchemaRequest, resp *provider.SchemaResponse) {
+	resp.Schema = framework_schema.Schema{
+		Attributes: map[string]framework_schema.Attribute{
+			SchemaKeyOrganization: framework_schema.StringAttribute{
+				Optional:    true,
+				Description: "The Buildkite organization slug",
+				// MarkdownDescription: "The Buildkite organization slug",
```
mcncl, line 102: Are we keeping these in for any reason? The `Description` maintains the same value and isn't unclear in what it is.

mcncl (approved):

buildkite/provider.go
mcncl: The `MarkdownDescription` would be in the commit history too, so I don't see a reason to just keep old code around.

lizrabuya: I agree with @mcncl on this one.

lizrabuya (approved):
Awesome!

buildkite/provider.go
jradtilbrook: Oh yeah we can add these in later really. The problem is with this muxing between sdkv2 and framework, the provider definition has to match _exactly_!
So I wanted to include `MarkdownDescription` because apparently we can start to automatically generate documentation which I think will be better. But sdkv2 doesn't have this property and it was breaking with it included.
Its the same reason I had to remove `DefaultFunc` from the sdkv2 provider, because that doesn't exist in framework now 🤷‍♂️

Once we have fully migrated to framework we could uncomment these. But I'll just create a task to come back to docs generation after the migration. And then we can remove them now

james2791 (approved):

Issue #287, opened 2023-06-16
Error: input: Clusters has not been activated for your organization when modifying provider settings without clusters enabled


HugeIRL:
**Describe the bug**

Setup:
**Agent:** v3.47.0 (local)
**Provider:** v0.19.0

Using a basic setup of the provider and creating a pipeline resource, once the pipeline is created and you attempt to add the following provider settings:

```
provider_settings {
        publish_commit_status_per_step = true
}
```

You will get the error:

```
│ Error: input: Clusters has not been activated for your organization
│ 
│ 
│   with buildkite_pipeline.test-pipeline,
│   on test-pipeline.tf line 1, in resource "buildkite_pipeline" "test-pipeline":
│    1: resource "buildkite_pipeline" "test-pipeline" {
```

**To Reproduce**
1. Create a basic pipeline resource (mine just has buildkite-agent pipeline upload as the command) without **provider_settings publish_commit_status_per_step** being set.
2. Init, Plan and Apply using Terraform and check that the pipeline created successfully.
3. Go back to the resource and attempt to add:
```
provider_settings {
        publish_commit_status_per_step = true
}
```
4. You will get the error:
```
│ Error: input: Clusters has not been activated for your organization
│ 
│ 
│   with buildkite_pipeline.test-pipeline,
│   on test-pipeline.tf line 1, in resource "buildkite_pipeline" "test-pipeline":
│    1: resource "buildkite_pipeline" "test-pipeline" {
```

**Expected behavior**
No error as you are not passing a clusterID at this point in time. The apply should complete successfully.

**Additional context**
I also tested this by adding cluster_id = "" to see if it was a work around, to no success.

jradtilbrook:
Oh no! Thanks for the bug report.

I think this might actually apply to any time you try update a pipeline, regardless of what you update.
#279 modified the UpdatePipeline function to use a new GraphQL library and it now sets `cluster_id` in the mutation no matter what. It should work because it sets it as null if not used but the API musn't be handling that properly. 

I'll look into it and get back with a path forward. For now, downgrading to <0.19.0 should fix this issue

HugeIRL:
> Oh no! Thanks for the bug report.
> 
> I think this might actually apply to any time you try update a pipeline, regardless of what you update. #279 modified the UpdatePipeline function to use a new GraphQL library and it now sets `cluster_id` in the mutation no matter what. It should work because it sets it as null if not used but the API musn't be handling that properly.
> 
[...]

PR #288, opened 2023-06-19
SUP-1078 Fix input error for org not using Clusters

mcncl:
## Release checklist:
- [x] `CHANGELOG.md` updated
- [x] `README.md` version examples updated
- [x] `docs/` updated

[Issue 287](https://github.com/buildkite/terraform-provider-buildkite/issues/287) highlights an error when modifying a pipeline if an organization doesn't have Clusters enabled. We can patch in an `omitempty` field in to the genqlient attribute in order to not pass the `cluster_id` if one is not set.

Caveat; a test had to be removed due to this change


james2791 (approved):
Awesome 💚

james2791 (approved):
✅

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to this project will be documented in this file.
 
+## [v0.19.1](https://github.com/buildkite/terraform-provider-buildkite/compare/v0.19.0...v0.19.1)
+### Fixes
+* Error: input: Clusters has not been activated for your organization [[Issue #287](https://github.com/buildkite/terraform-provider-buildkite/issues/287)] @mcncl
```
jradtilbrook, line 7: We've been using the PR and title for this instead of the issue link. We can link back from the PR to the issue it closes, but its the PR that introduces the change that is logged here

jradtilbrook:
```suggestion
* SUP-1078 Fix input error for org not using Clusters [[PR #288](https://github.com/buildkite/terraform-provider-buildkite/pull/288)] @mcncl
```

buildkite/resource_pipeline_test.go
```diff
@@ -95,19 +95,19 @@ func TestAccPipeline_add_remove_withcluster(t *testing.T) {
...
+			// 		// Confirm the pipeline exists in the buildkite API
+			// 		testAccCheckPipelineExists("buildkite_pipeline.foobar", &resourcePipeline),
+			// 		// Confirm the pipeline has the correct values in Buildkite's system
+			// 		testAccCheckPipelineRemoteValues(&resourcePipeline, "Test Pipeline foo"),
+			// 		// Confirm the pipeline has the correct values in terraform state
+			// 		resource.TestCheckResourceAttr("buildkite_pipeline.foobar", "name", "Test Pipeline foo"),
+			// 		resource.TestCheckResourceAttr("buildkite_pipeline.foobar", "cluster_id", "Q2x1c3Rlci0tLTRlN2JmM2FjLWUzMjMtNGY1OS05MGY2LTQ5OTljZmI2MGQyYg=="),
+			// 		resource.TestCheckResourceAttr("buildkite_pipeline.foobar", "allow_rebuilds", "true"),
+			// 	),
+			// },
```
jradtilbrook, line 110: Why is this test removed? I don't think we should have to change any acceptance tests for this bug

jradtilbrook (changes requested):

buildkite/resource_pipeline_test.go
jradtilbrook: Oh right yes. Its probably a bad way to test because its not so obvious, but the point of this second section is to modify the pipeline to remove the `cluster_id` definition so that it makes an API request to remove the pipeline from the cluster. Without this test, it doesn't verify that the pipeline is actually removed.

I've just uncommented and run it locally and it passes but it shouldn't. Because the tf config for this step should generate a plan to remove the cluster.

I think this shows the reason why `omitempty` wasn't included here originally (I wrote it but had forgotten). With `omitempty`, the `clusterId` is not passed with the GraphQL mutation when it is empty/null. But that is actually the functionality we need because that is how you remove a pipeline from a cluster

jradtilbrook:
We are going to fix this server-side

PR #296, opened 2023-06-28
Support option to archive on delete.

mcncl:
## What
Supports the option to `archive` a pipeline on `destroy` rather than completely removing it.

## Why
[Issue 195](https://github.com/buildkite/terraform-provider-buildkite/issues/195) requests the option to archive a pipeline on `destroy` rather than actually deleting it. This makes sense as it enables better auditing.

## How
Utilises the GraphQL `mutation` for `pipelineArchive` when the `archive_on_delete` field is set to `true`. When a user runs a `destroy` in the command line it will go through the normal process (output) of destroying the pipeline but will instead archive it. Archived pipelines will be accessible to Organization admins.


james2791 (changes requested):
Looks good. just needs the changelog bump (unreleased) with the callout of this PR 👍

james2791 (approved):
Nice 🙂

PR #297, opened 2023-06-29
SUP-1085: Cluster Queue resource implementation

james2791:
## PR checklist:
- [x] `docs/` updated
- [x] tests added

Adds the support for creating, updating, deleting, reading and importing Cluster Queue resources.


mcncl:
Reviews have been requested but this is in draft, should it be moved to ready for review?

buildkite/graphql/cluster_queue.graphql
```diff
@@ -0,0 +1,78 @@
+fragment ClusterQueueValues on ClusterQueue {
```
jradtilbrook, line 1: Any reason you didn't name this as just `ClusterQueue`?
```suggestion
fragment ClusterQueue on ClusterQueue {
```

buildkite/graphql/cluster_queue.graphql
```diff
@@ -0,0 +1,78 @@
...
+    id
+    uuid
+    key
+    description
```
jradtilbrook, line 5: You probably want to mark this as a pointer as well so we can store it as `null` in state. And check for any others that are defined as nullable types on this ClusterQueue type (I don't think there are, but just in case)

buildkite/resource_cluster_queue.go
```diff
@@ -0,0 +1,210 @@
...
+
+	// Find the Cluster Q from the returned queues to update state
+	for i := range queues.Organization.Cluster.Queues.Edges {
+		if queues.Organization.Cluster.Queues.Edges[i].Node.Id == state.Id.ValueString() {
+			// Update ClusterQueueResourceModel with Node values and append
+			updateClusterQueueResource(queues.Organization.Cluster.Queues.Edges[i].Node, &state)
+			resp.Diagnostics.Append(resp.State.Set(ctx, &state)...)
+			break
+		}
+	}
```
jradtilbrook, line 144: You can make this a bit nicer to read

```suggestion
	for i, edge := range queues.Organization.Cluster.Queues.Edges {
		if edge.Node.Id == state.Id.ValueString() {
			// Update ClusterQueueResourceModel with Node values and append
			updateClusterQueueResource(edge.Node, &state)
			resp.Diagnostics.Append(resp.State.Set(ctx, &state)...)
			break
		}
	}
```

jradtilbrook: I doubt we will hit it but there is the chance that 50 cluster queues is not enough. I don't think we need to code for it right now, but at least logging if we go through all 50 queues and don't find the correct one, error out. And if we get reports of hitting that error, we can fix it then

docs/resources/cluster_queue.md
```diff
@@ -0,0 +1,29 @@
...
+
+* `cluster_id` - (Required) The ID of the cluster that this cluster queue belongs to.
+* `key` - (Required) The key of the cluster queue.
+* `description` - (Required) This is the description of the cluster queue.
+
+## Attribute Reference
+
+* `id` - The Graphql ID of the created cluster queue.
+* `uuid` - The UUID of the created cluster queue.
+* `cluster_uuid` - The UUID of the cluster that this cluster queue belongs to.
```
jradtilbrook, line 29: Docs on importing and an example query would be great to have as well

jradtilbrook (changes requested):

buildkite/graphql/cluster_queue.graphql
mcncl: I'd argue that `Values` makes this more readable; it doesn't fill an entire ClusterQueue.

james2791: Also for readability - i.e `ClusterQueue on ClusterQueue` -`ClusterQueueValues` being the values returned via said fragment in the response(s) of the query/mutations 👍

jradtilbrook: Ahh yep okay fair enough 👍

docs/resources/cluster_queue.md
james2791: Added with latest commit 👍 - needed a fresh `ImportState` function to handle imports (with passed in Cluster UUID, queue ID)

buildkite/graphql/cluster_queue.graphql
james2791: `description` is now a pointer - nullable in state

buildkite/resource_cluster_queue.go
james2791: Added a check to the bottom of `Read` to log that case - more a tradeoff until the fetching of 50 queues in a cluster can be switched to a `node` or a `clusterQueue(id: "x")` query

buildkite/resource_cluster_queue.go
```diff
@@ -0,0 +1,245 @@
...
+
+	// If not returned by this point, the cluster queue could not be found
+	// This is a tradeoff of the current getClusterQueues Genqlient query (searches for 50 queues via the cluster UUID in state)
+	if !queueFound {
+		resp.Diagnostics.AddError(
+			"Unable to find Cluster Queue",
+			fmt.Sprintf("Unable to find Cluster Queue: %s", err.Error()),
+		)
+		return
+	}
```
mcncl, line 160: This doesn't require a conditional for 2 reasons; 1 is that queueFound is never set to `true`, so checking that its value has changed is redundant.

The second is that on a number of occasions (line 127, 138 and 148) the `Read` function is exited, so the program cannot make it to this point and not be in a position to produce an error. Thus, the below should behave the same way.

```go
resp.Diagnostics.AddError(
    "Unable to find Cluster Queue",
	fmt.Sprintf("Unable to find Cluster Queue: %s", err.Error()),
)
return
```

buildkite/resource_cluster_queue_test.go
```diff
@@ -0,0 +1,218 @@
...
+		// If clusterQueueResourceModel isnt set from the queues slice
+		if clusterQueueResourceModel.Id.ValueString() == "" {
+			return fmt.Errorf("No Cluster queue found with graphql id: %s", resourceState.Primary.ID)
+		}
+
+		return nil
+	}
+}
+
+func testAccCheckClusterQueueRemoteValues(cq *ClusterQueueResourceModel, description string, key string) resource.TestCheckFunc {
```
mcncl, line 164: When 2 parameters share a type you can declare the type once, ie:

```go
func testAccCheckClusterQueueRemoteValues(cq *ClusterQueueResourceModel, description, key string) resource.TestCheckFunc {
```

buildkite/resource_cluster_queue.go
mcncl: Example: https://go.dev/play/p/XYAsrs7q8u0

buildkite/resource_cluster_queue_test.go
james2791: 👍

buildkite/resource_cluster_queue.go
james2791: Thinking it would be the case - amended 👍

buildkite/resource_cluster_queue_test.go
```diff
@@ -0,0 +1,218 @@
...
+
+		return nil
+	}
+}
+
+func testAccCheckClusterQueueRemoteValues(cq *ClusterQueueResourceModel, description, key string) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+
+		if cq.Key.ValueString() != key {
+			return fmt.Errorf("Remote Cluster queue key (%s) doesn't match expected value (%s)", cq.Description, description)
```
mcncl, line 168: Copy/paste gotcha?

Should be `cq.Key, key`.

james2791: Yep 🙂 Cheers

examples/main.tf
```diff
@@ -25,10 +25,17 @@ resource "buildkite_agent_token" "fleet" {
   description = "token used by build fleet"
 }
 
+resource "buildkite_cluster_queue" "queue1" {
+  cluster_id = "Q2x1c3Rlci0tLTMzMDc0ZDhiLTM4MjctNDRkNC05YTQ3LTkwN2E2NWZjODViNg=="
```
mcncl, line 29: I was wondering if we want to expose the GraphQL IDs for our own resources, but we have them hard coded in tests anyway so guess it's not a huge concern.

mcncl (approved):

docs/resources/cluster_queue.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Import
+
+Cluster queues can be imported using its `GraphQL ID`, along with its respective cluster `UUID`, separated by a comma. e.g.
+
+```
+$ terraform import buildkite_cluster_queue.test Q2x1c3RlclF1ZXVlLS0tYjJiOGRhNTEtOWY5My00Y2MyLTkyMjktMGRiNzg3ZDQzOTAz,35498aaf-ad05-4fa5-9a07-91bf6cacd2bd
```
jradtilbrook, line 36: The ideal use-case here is only 1 ID. And ideally one of the computed fields on the resource. Thats because the other fields required to read the resource should be defined in terraform before you import it anyway. Eg:

```hcl
resource "buildkite_cluster_queue" "test" {
  cluster_id = "<id>"
  key = "main"
  description = "main pipelines"
}
```

Then you simply run `terraform import buildkite_cluster_queue.test Q2x1c3RlclF1ZXVlLS0tYjJiOGRhNTEtOWY5My00Y2MyLTkyMjktMGRiNzg3ZDQzOTAz` and the `ImportState` function uses the `cluster_id` defined in the `resource.ImportStateRequest`.

buildkite/resource_cluster_queue.go
```diff
@@ -0,0 +1,242 @@
...
+	var state ClusterQueueResourceModel
+
+	resp.Diagnostics.Append(req.State.Get(ctx, &state)...)
+
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	log.Printf("Getting cluster queues for cluster %s ...", state.ClusterUuid.ValueString())
+	queues, err := getClusterQueues(cq.client.genqlient, cq.client.organization, state.ClusterUuid.ValueString())
```
jradtilbrook, line 130: I notice the `cluster` field on `Organization` has an `id` argument, but the docs for that say its a UUID which I imagine is why you are using `state.ClusterUuid` here. Is that correct? If it is, that seems like a bug in the GraphQL because everywhere else `id` is used as an argument, its the Relay Node ID, not the UUID.

It's probably not something we can really fix, but something to be aware of

jradtilbrook (changes requested):
This is looking great. I just have a request about importing 👇

buildkite/resource_cluster_queue.go
james2791: Correct - described [here](https://buildkite.com/user/graphql/documentation/type/Organization) - needing the `UUID` since the query is based from an `Organisation` query 👍

docs/resources/cluster_queue.md
jradtilbrook: Discussed offline: we'll look to improve this when refactoring to the node query when its implemented server side

jradtilbrook (dismissed):

jradtilbrook (approved):

PR #301, opened 2023-07-04
SUP-1084 Add Cluster resource

mcncl:
## What
- Adds support for Cluster resource
    - All CRUD functionas + Import
    - Tested locally
- Tests for use in CI to ensure CRUD + Import work as expected
- Usage examples in the docs

TODO:
- Update CHANGELOG
    - Will import latest when Queues PR merged in
- Fix Import tests
    - Fail because resource values are `null` even though the resources are created fine

## PR checklist:
- [x] `docs/` updated
- [x] tests added


buildkite/graphql/cluster.graphql
```diff
@@ -0,0 +1,70 @@
...
+    organization(slug: $orgSlug) {
+        cluster(id: $id) {
+            ...ClusterFields
+        }
+    }
+}
+
+# @genqlient(for: "ClusterCreateInput.description", pointer: true)
+# @genqlient(for: "ClusterCreateInput.emoji", pointer: true)
+# @genqlient(for: "ClusterCreateInput.color", pointer: true)
```
jradtilbrook, line 20: I thought you only had to do it like this if you are accepting a struct type as input for the mutation? Otherwise you could add these annotations above the individual arguments below on lines 24-26

buildkite/resource_cluster.go
```diff
@@ -0,0 +1,216 @@
...
+
+	if err != nil {
+		resp.Diagnostics.AddError(
+			"Unable to read Cluster",
+			fmt.Sprintf("Unable to read Cluster: %s", err.Error()),
+		)
+		return
+	}
+
+	resp.Diagnostics.Append(resp.State.Set(ctx, &state)...)
```
jradtilbrook, line 131: I think this needs to set the other fields on the cluster

buildkite/resource_cluster.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+
+	id := cluster.Organization.Cluster.Id
+
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("id"), types.StringValue(id))...)
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("uuid"), uuid)...)
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("name"), types.StringValue(cluster.Organization.Cluster.Name))...)
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("description"), types.StringValue(cluster.Organization.Cluster.Description))...)
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("emoji"), types.StringValue(cluster.Organization.Cluster.Emoji))...)
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("color"), types.StringValue(cluster.Organization.Cluster.Color))...)
```
jradtilbrook, line 215: These should be moved to the Read function. That function will be called after ImportState anyway and Read should set these values. ImportState is responsible for setting the minimal fields for Read to work

jradtilbrook (changes requested):

buildkite/resource_cluster.go
james2791: Also got a hunch that this is why the import test is failing - similar behaviour with queues

buildkite/graphql/cluster.graphql
mcncl: You do, I don't personally like the inline style however can change it if there are strong opinions on it. I prefer directives at the top bet not a hill I'll die on.

jradtilbrook: Neither, as long as it works 👍

docs/resources/cluster.md
```diff
@@ -0,0 +1,29 @@
+# Resource: cluster
+
+This resource allows you to create, manage and import Clusters.
+
+Buildkite documentation: https://buildkite.com/docs/clusters/overview
+
+## Example Usage
+
+```hcl
+provider "buildkitE" {}
```
jradtilbrook, line 10:
```suggestion
provider "buildkite" {}
```

docs/resources/cluster.md
```diff
@@ -0,0 +1,29 @@
...
+* `description` - (Optional) This is a description for the cluster, this may describe the usage for it, the region, or something else which would help identify the Cluster's purpose.
+* `emoji` - (Optional) An emoji to use with the Cluster, this can either be set using `:buildkite:` notation, or with the emoji itself, such as 😎.
+* `color` - (Optional) A color to associate with the Cluster. Perhaps a team related color, or one related to an environment. This is set using hex value, such as `#BADA55`.
+
+## Attribute Reference
+
+* `id` - The GraphQL ID that is created with the Cluster.
+* `uuid` - The UUID created with the Cluster.
+
+The `uuid` is used when *getting* a Cluster. The `id` is used when *modifying* a Cluster.
```
jradtilbrook, line 29: Add docs and example query for importing

jradtilbrook (changes requested):

go.mod
```diff
@@ -4,6 +4,9 @@ go 1.20
 
 require (
 	github.com/Khan/genqlient v0.6.0
+	github.com/hashicorp/terraform-plugin-framework v1.3.1
```
jradtilbrook, line 7: I have a feeling the changes to go.mod and go.sum are the cause of the unrelated failing tests. Maybe some other pieces need changes with these changes to dependencies. And that is out of scope for this ticket. I've rebuilt `main` here https://buildkite.com/buildkite/terraform-provider-buildkite-main/builds/574 and it passes just fine. So I'm advocating for reverting these changes to see if it helps

buildkite/resource_cluster.go
```diff
@@ -0,0 +1,200 @@
...
+		resp.Diagnostics.AddError(
+			"Unable to delete Cluster",
+			fmt.Sprintf("Unable to delete Cluster: %s", err.Error()),
+		)
+		return
+	}
+}
+
+func (c *clusterResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
+	req.ID = ""
```
jradtilbrook, line 189: You can't empty out the ID here, its used in the function below to proxy this id through to the UUID of the cluster
```suggestion
```

james2791 (approved):
💚

jradtilbrook (approved):

PR #304, opened 2023-07-05
Migrate organization datasource to plugin framework

jradtilbrook:
## PR checklist:
- [x] `docs/` updated
- [x] tests added

### DIFF
--- a/buildkite/data_source_organization.go
+++ b/buildkite/data_source_organization.go
@@ -1,13 +1,80 @@
 package buildkite
 
 import (
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-framework/datasource"
+	"github.com/hashicorp/terraform-plugin-framework/datasource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/types"
 )
 
-func dataSourceOrganization() *schema.Resource {
-	resource := resourceOrganizationSettings()
-	return &schema.Resource{
-		ReadContext: resource.ReadContext,
-		Schema:      resource.Schema,
+type organizationDatasourceModel struct {
+	AllowedApiIpAddresses types.List   `tfsdk:"allowed_api_ip_addresses"`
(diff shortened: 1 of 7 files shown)


jradtilbrook:
We have a `resource.buildkite_organization_settings` but a `data.buildkite_organization`. They refer to the same thing in the API so I'd like to consolidate the naming. Any opinions on which way to go?

buildkite/data_source_organization.go
```diff
@@ -1,13 +1,85 @@
...
+		Attributes: map[string]schema.Attribute{
+			"id": schema.StringAttribute{
+				Computed: true,
+			},
+			"uuid": schema.StringAttribute{
+				Computed: true,
+			},
+			"allowed_api_ip_addresses": schema.ListAttribute{
+				Computed:    true,
+				ElementType: types.StringType,
```
mcncl, line 81: I don't think this value is a `Computed` one. It's optional in the resource and so should also be in the datasource. If it's unset then it is set in the UI as an empty string.

The response when unset is:

```json
{
  "data": {
    "organization": {
      "allowedApiIpAddresses": ""
    }
  }
}
```

mcncl (changes requested):

buildkite/data_source_organization.go
jradtilbrook: I'm confused though because this is a datasource. It's not about putting any changes up to the API, its about getting data from the API for use somewhere else in terraform.

Technically I think its all computed by default for a datasource right? Except for maybe the ID or some means to actually identify the object in the API. But since this is tied into the providers `slug` configuration, its redundant to specify it here as well.
Perhaps we can just remove `Computed` and not specify any other fields as well?

jradtilbrook: Ahh from the docs on datasource schema: 

```
// Computed indicates whether the provider may return its own value for
// this Attribute or not. Required and Computed cannot both be true. If
// Required and Optional are both false, Computed must be true, and the
// attribute will be considered "read only" for the practitioner, with 
// only the provider able to set its value.                            
```

So its not required or optional. This field should not be set by the user at all. So it has to be marked as computed

mcncl (approved):

Issue #307, opened 2023-07-10
Terraform provider documentation lists rebuilds are true by default, but they are false 


butterkitty:
New pipelines have their rebuild option set to false by default.

Tested via provider version 0.19.2. Customer is using 0.19.0.

The [terraform provider](https://registry.terraform.io/providers/buildkite/buildkite/latest/docs) currently states that rebuilds are set to true by default, but they are currently being set to false by default.

As a workaround, Setting it to true manually does work.

PR #308, opened 2023-07-10
Fixed default for allow_rebuilds on provider to match docs

butterkitty:
Removed computed since this value should be known on creation.

Added default value to true since this is described in the docs to be true by default.

Fixes: https://github.com/buildkite/terraform-provider-buildkite/issues/307


mcncl (approved):
🚀

PR #309, opened 2023-07-12
SUP-1086 Add cluster agent token resource

lizrabuya:
Add support to Cluster Agent token resource
-  Implement CRUD functions
- Add tests
- Updated docs with examples


buildkite/resource_cluster_agent_token.go
```diff
@@ -0,0 +1,193 @@
...
+				Computed: true,
+			},
+			"uuid": resource_schema.StringAttribute{
+				Computed: true,
+			},
+			"description": resource_schema.StringAttribute{
+				Optional:            true,
+				MarkdownDescription: "A description about what this cluster agent token is used for",
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
```
jradtilbrook, line 58: I don't think this requires replacing the cluster since there is the update mutation. The old agent tokens didnt have that so needed to be replaced

docs/resources/cluster_agent_token.md
```diff
@@ -0,0 +1,66 @@
...
+resource "buildkite_cluster_agent_token" "cluster-token-1" {
+  cluster_id = "Q2x1c3Rlci0tLTkyMmVjOTA4LWRmNWItNDhhYS1hMThjLTczMzE0YjQ1ZGYyMA==" 
+  description = "agent token for cluster-1" 
+}
+```
+
+## Argument Reference
+
+* `cluster_id` - (Required) The ID of the cluster that this cluster queue belongs to.
+* `description` -  (Required) A description about what this cluster agent token is used for.
```
jradtilbrook, line 21:
```suggestion
* `description` -  (Optional) A description about what this cluster agent token is used for.
```

docs/resources/cluster_agent_token.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Attribute Reference
+
+* `id` - The Graphql ID of the created cluster queue.
+* `uuid` - The UUID of the created cluster queue.
+* `cluster_uuid` - The UUID of the cluster that this cluster queue belongs to.
+
+
+To find the cluster's `UUID` to utilize, you can use the below GraphQL query below. Alternatively, you can use this [pre-saved query](https://buildkite.com/user/graphql/console/3adf0389-02bd-45ef-adcd-4e8e5ae57f25), where you will need fo fill in the organization slug (ORGANIZATION_SLUG) for obtaining the relevant cluster name/`UUID` that the cluster queue is in.
```
jradtilbrook, line 30: I think we actually need the ID for this resource right? The UUID is a computed property

jradtilbrook (changes requested):
Ohh missed out on a few things

lizrabuya (commented):
👍

buildkite/resource_cluster_agent_token.go
```diff
@@ -0,0 +1,193 @@
...
+func (ct *ClusterAgentToken) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) {
+	var state, plan ClusterAgentTokenResourceModel
+
+	resp.Diagnostics.Append(req.State.Get(ctx, &state)...)
+	resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)
+
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	_, err := updateClusterAgentToken(
```
james2791, line 156: I think adding another `log.Printf` to the Update fn like the other operations might be good for consistency 👍

```suggestion
    log.Printf("Updating cluster token %s", state.Id.ValueString())
	_, err := updateClusterAgentToken(
```

james2791 (dismissed):

docs/resources/cluster_agent_token.md
lizrabuya: Description is a required field for cluster agent token

buildkite/resource_cluster_agent_token.go
```diff
@@ -0,0 +1,193 @@
...
+		MarkdownDescription: "A Cluster Agent Token is a token used to connect an agent to a cluster in Buildkite.",
+		Attributes: map[string]resource_schema.Attribute{
+			"id": resource_schema.StringAttribute{
+				Computed: true,
+			},
+			"uuid": resource_schema.StringAttribute{
+				Computed: true,
+			},
+			"description": resource_schema.StringAttribute{
+				Optional:            true,
```
jradtilbrook, line 55: You'll need to remove this as its technically required by the backend

docs/resources/cluster_agent_token.md
jradtilbrook: This new resource needs the `cluster_id` to link to a cluster, but this example query demonstrates how to get the `cluster_uuid`. So we need to update the wording and query to retrieve `cluster_id` instead since that is what is required

buildkite/resource_cluster_agent_token.go
jradtilbrook: @lizrabuya the cluster agent token has an update mutation which allows updating the description of the token in place - it doesn't create a new one. This plan modifier tells the terraform framework that if the user changes the description attribute that it should automatically delete and recreate, which is not necessary. The unclustered agent tokens don't have an update method so needed to be recreated.

So this can be removed from the schema for this attribute.

jradtilbrook (dismissed):

jradtilbrook (approved):

james2791 (approved):
Looks the goods 👍

buildkite/graphql/cluster_agent_token.graphql
```diff
@@ -0,0 +1,71 @@
+fragment ClusterAgentTokenValues on ClusterToken {
+    cluster {
+        id
+        uuid
+    }
+    description
+    id
+    uuid
+}
```
mcncl, line 9: Love this.

buildkite/provider.go
```diff
@@ -94,8 +94,9 @@ func (tf *terraformProvider) Metadata(ctx context.Context, req provider.Metadata
 func (*terraformProvider) Resources(context.Context) []func() resource.Resource {
 	return []func() resource.Resource{
 		NewAgentTokenResource,
-		newClusterResource,
+		NewClusterAgentTokenResource,
```
mcncl, line 97: Nit: Functions and variables in this don't need to be capitalised, for the most part. They're not exported from the package and so can remain internal.

mcncl (approved):

buildkite/provider.go
jradtilbrook: I would like us to do more idiomatic go like this 👍
There is so much already that we need to clean up as well. Lets just lump that onto future us to tidy up at some point

PR #311, opened 2023-07-13
SUP-1067: Organization resource conversion to Framework

james2791:
## PR checklist:
- [x] `docs/` updated
- [x] tests added


buildkite/provider.go
```diff
@@ -135,11 +136,10 @@ func New(version string) provider.Provider {
 func Provider(version string) *schema.Provider {
 	provider := &schema.Provider{
 		ResourcesMap: map[string]*schema.Resource{
-			"buildkite_pipeline":              resourcePipeline(),
-			"buildkite_pipeline_schedule":     resourcePipelineSchedule(),
-			"buildkite_team":                  resourceTeam(),
-			"buildkite_team_member":           resourceTeamMember(),
-			"buildkite_organization_settings": resourceOrganizationSettings(),
```
jradtilbrook, line 142: This is still draft but one first comment: remember we will need to maintain this name `buildkite_organization_settings` as well until v1 or at least for a little while

buildkite/resource_organization_settings.go
```diff
@@ -3,114 +3,262 @@ package buildkite
...
+type organizationResource struct {
+	client *Client
+}
+
+func newOrganizationResource() resource.Resource {
+	return &organizationResource{}
+}
+
+func (organizationResource) Metadata(ctx context.Context, req resource.MetadataRequest, resp *resource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_organization_settings"
```
mcncl, line 32: Given the `datasource` is just called `organization`, would we not do the same here?

jradtilbrook: Yeah thats right. The new resource built using the framework (this one) is `buildkite_organization` and we will leave the old one around using sdkv2 named `buildkite_organization_settings`. After some deprecation period, we can remove it. Lets say by version v0.21.0 - this will need documenting as well

docs/resources/organization_settings.md
```diff
@@ -9,7 +9,7 @@ Note: The "Allowed API IP Addresses" feature must be enabled on your organizatio
 ## Example Usage
 
 ```hcl
-resource "buildkite_organization_settings" "test_settings" {
+resource "buildkite_organization_settings "test_settings" {
```
jradtilbrook, line 12: Mistakenly removed `"`?

buildkite/resource_organization_settings_test.go
```diff
@@ -86,29 +86,12 @@ func TestAccOrganizationSettings_import(t *testing.T) {
 	})
 }
 
-func TestAccOrganizationSettings_disappears(t *testing.T) {
```
jradtilbrook, line 89: Why is this removed? As with most of these migrations, we shouldn't need to change the acceptance tests

jradtilbrook: We also need a test against the new resource name buildkite_organization

jradtilbrook (changes requested):
Its looking good 👍
Im confused about why the getOrganization function is getting called everwhere. FWIW you can get the org ID from client.organizationID

buildkite/resource_organization_settings_test.go
mcncl: Yeah, this can technically be removed if you’re checking for the removal on the Create test, but given that only tests that the resource is added you’ll need to check it has been removed too.

docs/resources/organization.md
```diff
@@ -0,0 +1,29 @@
+# Resource: organization
```
jradtilbrook, line 1: I think we should call out in the docs for `organization_settings` that it is deprecated as well

jradtilbrook: And you can also add a `Deprecated` field to the schema of the SDKv2 implementation so it will output on the log when running terraform

buildkite/resource_organization.go
```diff
@@ -0,0 +1,267 @@
...
+		return
+	}
+
+	if response.Organization.Id == "" {
+		resp.Diagnostics.AddError(
+			"Organization not found",
+			fmt.Sprintf("Organization not found: %s", err.Error()),
+		)
+		return
+	}
```
jradtilbrook, line 92: These errors will surface from the request on line 98 itself so to save an extra request slowing down the provider we can remove these. There's a few below in the other functions as well

jradtilbrook (changes requested):

buildkite/resource_organization_settings.go
```diff
@@ -9,6 +9,9 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
+var deprecationMessage = `This resource has been deprecated in favour of the newer buildkite_organization resource. 
+Please visit the provider's documentation at https://registry.terraform.io/providers/buildkite/buildkite/latest/docs fo more details.`
```
mcncl, line 13: "fo" 😉

mcncl (approved):

jradtilbrook (approved):

mcncl (approved):
🚀

james2791:
🎆

PR #313, opened 2023-07-18
SUP-1051: Team Member resource Framework/Genqlient conversion

james2791:
## PR checklist:
- [x] `docs/` updated
- [x] tests added

This PR also introduces a `node.graphql` - to its name, allows for the easy obtaining of any type from the GraphQL that implements the `Node` interface. The `getNode` query currently has a Team Member type handling inside (with the corresponding `TeamMemberValues` fragment in `team_member.graphql` (note this was the approach instead of aliasing each `node` subquery which might've re-generated the same code with different function signatures)


buildkite/resource_team_member.go
```diff
@@ -3,174 +3,208 @@ package buildkite
...
+	if req.ProviderData == nil {
+		return
+	}
+
+	tm.client = req.ProviderData.(*Client)
+}
+
+func (teamMemberResource) Schema(ctx context.Context, req resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = resource_schema.Schema{
+		MarkdownDescription: "A team member resourc allows for the management of team team membership for existing organization users.",
```
mcncl, line 48: resource* and double "team"

buildkite/resource_team_member.go
```diff
@@ -3,174 +3,208 @@ package buildkite
...
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
 			},
-			"user_id": &schema.Schema{
-				Required: true,
-				Type:     schema.TypeString,
+			"team_id": resource_schema.StringAttribute{
+				Required:            true,
+				MarkdownDescription: "The GraphQL ID of the team to add/remove the user to/from.",
```
mcncl, line 64: Is there a function to remove the user from a team? Or does simple removing it from the config remove them from the team? Not sure this description is clear.

buildkite/resource_team_member.go
```diff
@@ -3,174 +3,208 @@ package buildkite
...
 			},
-			"uuid": &schema.Schema{
-				Computed: true,
-				Type:     schema.TypeString,
+			"user_id": resource_schema.StringAttribute{
+				Required:            true,
+				MarkdownDescription: "The GraphQL ID of the user to add/remove.",
```
mcncl, line 68: Same. Would simply not providing a `user_id` remove that user from the team? When setting this value, I don't think I set it if I want to `remove`.

buildkite/resource_team_member.go
```diff
@@ -3,174 +3,208 @@ package buildkite
...
-
-	var mutation struct {
-		TeamMemberCreate struct {
-			TeamMemberEdge struct {
-				Node TeamMemberNode
-			}
-		} `graphql:"teamMemberCreate(input: {teamID: $teamId, userID: $userId})"`
-	}
+func (tm *teamMemberResource) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
+	var plan, state teamMemberResourceModel
```
mcncl, line 82: If it's just a Create, you don't need 2 iterations of the Model. Maybe just use `plan` for clarity as on Create there is no `state`.

buildkite/resource_team_member.go
```diff
@@ -3,174 +3,208 @@ package buildkite
...
-	return nil
+	// Update state with values from API response/plan
+	state.Id = types.StringValue(apiResponse.TeamMemberCreate.TeamMemberEdge.Node.Id)
+	state.Uuid = types.StringValue(apiResponse.TeamMemberCreate.TeamMemberEdge.Node.Uuid)
+	state.TeamId = plan.TeamId
+	state.UserId = plan.UserId
+	// The role of the user will by default be "MEMBER" if none was entered in the plan
+	state.Role = types.StringValue(string(apiResponse.TeamMemberCreate.TeamMemberEdge.Node.Role))
+
+	resp.Diagnostics.Append(resp.State.Set(ctx, &state)...)
```
mcncl, line 114: `&plan` would be clearer here; you're setting state to the values of `&THING`.

buildkite/graphql/node.graphql
```diff
@@ -0,0 +1,9 @@
+query get(
```
mcncl, line 1: Just `get` as a query name doesn't provide clarity on what it's getting.

buildkite/resource_team_member.go
```diff
@@ -3,174 +3,208 @@ package buildkite
...
 
-	vars := map[string]interface{}{
-		"id": d.Id(),
+	if resp.Diagnostics.HasError() {
+		return
 	}
 
-	err := client.graphql.Query(context.Background(), &query, vars)
+	log.Printf("Reading team member %s ...", state.Id.ValueString())
+	apiResponse, err := get(tm.client.genqlient, state.Id.ValueString())
```
mcncl, line 127: See comment in `graphql` file; consider a clearer function name.

buildkite/resource_team_member.go
```diff
@@ -3,174 +3,208 @@ package buildkite
...
-	vars := map[string]interface{}{
-		"id": graphql.ID(d.Id()),
+	if resp.Diagnostics.HasError() {
+		return
 	}
 
-	err := client.graphql.Mutate(context.Background(), &mutation, vars)
+	log.Printf("Deleting team member with ID %s ...", id)
+	_, err := deleteTeamMember(tm.client.genqlient, id)
+
```
mcncl, line 194: Why not just `_, err := deleteTeamMember(tm.client.genqlient, state.ID.ValueString())`?

buildkite/resource_team_member_test.go
```diff
@@ -22,9 +21,9 @@ func TestAccTeamMember_add_remove(t *testing.T) {
 				Config: testAccTeamMemberConfigBasic("MEMBER"),
 				Check: resource.ComposeAggregateTestCheckFunc(
 					// Confirm the team member exists in the buildkite API
-					testAccCheckTeamMemberExists("buildkite_team_member.test", &resourceTeamMember),
+					testAccCheckTeamMemberExists("buildkite_team_member.test", &tm),
 					// Confirm the team member has the correct values in Buildkite's system
-					testAccCheckTeamMemberRemoteValues(&resourceTeamMember, "MEMBER"),
+					testAccCheckTeamMemberRemoteValues(&tm, "MEMBER"),
```
mcncl, line 26: Knit, but passing in `string` then `&reference` in the first function, but then `&reference` then `string` in the second is a little jarring.

examples/main.tf
```diff
@@ -8,6 +8,12 @@ resource "buildkite_team" "test" {
   default_member_role = "MEMBER"
 }
 
+resource "buildkite_team_member" "member1" {
+  role = "MEMBER"
+	team_id = "VXfhnVUS78HavgtP55WhWGzT401guK38Vm9LMMeCgQD124m8xaKBRq0Fth=="
+	user_id = "VXNbwSA9hwVPpMgUXu1dWIDWf45ZwU6J7deETygiLUrKBg2TZBxuDr6aKj=="
+}
+
```
mcncl, line 16: So, this is fine, but I think going forward we could create both a team and a user and then dynamically add in the values rather than hardcode.

mcncl (changes requested):
Some that could be changed, some that are future things.

buildkite/graphql/node.graphql
jradtilbrook: Yeah I like getNode

buildkite/graphql/team_member.graphql
```diff
@@ -0,0 +1,60 @@
...
+    user {
+        id
+    }
+    role 
+}
+
+mutation createTeamMember(
+    $teamID: ID!, 
+    $userID: ID!,
+    # @genqlient(omitempty: true)
```
jradtilbrook, line 16: Whats the reasoning behind omitempty instead of pointer: true?

buildkite/resource_team_member.go
jradtilbrook: Yeah removing this resource removes them from the team; it removes the link between user and team. So the descriptions here are simply to "add"

buildkite/resource_team_member.go
```diff
@@ -3,174 +3,208 @@ package buildkite
...
 	}
 
-	return nil
+	// Update state with values from API response/plan
+	state.Id = types.StringValue(apiResponse.TeamMemberCreate.TeamMemberEdge.Node.Id)
+	state.Uuid = types.StringValue(apiResponse.TeamMemberCreate.TeamMemberEdge.Node.Uuid)
+	state.TeamId = plan.TeamId
+	state.UserId = plan.UserId
+	// The role of the user will by default be "MEMBER" if none was entered in the plan
+	state.Role = types.StringValue(string(apiResponse.TeamMemberCreate.TeamMemberEdge.Node.Role))
```
jradtilbrook, line 112: Is this still true? There used to be a bug in the API but I think its been resolved. If not specified, the default will be set to the default configured on the team. So this could possibly modify state and cause a subsequent apply to generate a diff which is something we don't want. We should add a test for that edge case. Happy to pair on that with you

jradtilbrook (changes requested):

buildkite/graphql/node.graphql
james2791: Me thee actually, thanks for the callout - more to get around the generated Node type naming - but just a few characters longer anyway

buildkite/resource_team_member.go
james2791: Adjusting these - thanks both

buildkite/resource_team_member.go
james2791: Adjusting these - thanks both

buildkite/resource_team_member.go
james2791: Strikes again - thanks: adjusting in to-come-commit

buildkite/resource_team_member.go
james2791: Dropping to one model in the latest commit

buildkite/resource_team_member.go
james2791: Dropping to one model in the latest commit

buildkite/resource_team_member.go
james2791: Adjusting in latest commit

buildkite/resource_team_member.go
james2791: Adjusting in latest commit

buildkite/resource_team_member_test.go
james2791: Solid - flipping the `testAccCheckTeamMemberRemoteValues` params in the adjustment commit

examples/main.tf
james2791: Fair - honestly these values are length 58 sequences for examples only in truth 👍

mcncl (dismissed):

buildkite/resource_team_member.go
james2791: Per the chat - I think leaving the resource as is from the older SDKv2 impl (with role always required) would keep this for a future improvement/update 
Happy for other thoughts though 👍

buildkite/graphql/team_member.graphql
james2791: More to get around the eventual case that we can omit it entirely (and then the API should create said TM with the default role for the team) - more below (and also an improvement case since the older resource had the role as required (not optional) 👍

jradtilbrook (approved):

james2791:
@mcncl one more before I wrap this in 👍

mcncl (approved):

james2791:
🚀

PR #315, opened 2023-07-20
SUP-1065: Pipeline Datasource conversion to Framework

james2791:
## PR checklist:
- [x] `docs/` updated
- [x] tests added


buildkite/data_source_pipeline.go
```diff
@@ -2,70 +2,118 @@ package buildkite
...
+		resp.Diagnostics.AddError(
+			"Unable to read pipeline",
+			fmt.Sprintf("Unable to read pipeline: %s", err.Error()),
+		)
+		return
 	}
 
 	if pipeline.Pipeline.Id == "" {
-		return diag.FromErr(errors.New("Pipeline not found"))
+		if err != nil {
```
jradtilbrook, line 100: This condition is covered above. I think it should be removed here

jradtilbrook (changes requested):
Looks good. Just one thing

buildkite/data_source_pipeline.go
james2791: Multi err - in latest 👍

jradtilbrook (approved):

PR #318, opened 2023-07-20
SUP 1049/migrate team

mcncl:
- Migrate all team to Framework & Genqlient
- Team datasource
- members_can_create_pipelines needs to be set as Member sets as pointer
- Update CHANGELOG and docs/

## PR checklist:
- [x] `docs/` updated
- [x] tests added


buildkite/provider.go
```diff
@@ -140,12 +142,10 @@ func Provider(version string) *schema.Provider {
 		ResourcesMap: map[string]*schema.Resource{
 			"buildkite_pipeline":              resourcePipeline(),
 			"buildkite_pipeline_schedule":     resourcePipelineSchedule(),
-			"buildkite_team":                  resourceTeam(),
```
james2791, line 143: Latest merge with the conversion of the Pipelines DS brought conflicts here

buildkite/graphql/team.graphql
```diff
@@ -0,0 +1,84 @@
...
+	# @genclient(pointer: true)
+	description
+	slug
+	privacy
+	isDefaultTeam
+	defaultMemberRole
+	membersCanCreatePipelines
+}
+
+query getTeam($id:ID!) {
```
james2791, line 14: Could this  query be moved into the `getNode` query in `node.graphql`?

james2791 (changes requested):

buildkite/graphql/team.graphql
mcncl: That feels a bit refactor-y? It would require a change to most of the source files in this.

james2791: Might be a little bit - I _think_ the only difference would be the type naming of the generated code, more keeping things in one place thing

mcncl: @james2791 if it's an agreed upon _thing_ then I can change it; it felt experimental at the time. Feels like maybe something else we migrate to but I'm not sure if it's a requirement?

@jradtilbrook can decide.

buildkite/data_source_team.go
```diff
@@ -2,84 +2,118 @@ package buildkite
...
+		resp.Diagnostics.AddError(
+			"Unable to get team",
+			fmt.Sprintf("Error getting team: %s", err.Error()),
+		)
+		return
 	}
 
-	if response.Team.Id == "" {
-		return diag.FromErr(errors.New(fmt.Sprintf("Team not found: '%s'", d.Get("slug"))))
+	if converted, ok := res.GetNode().(*getTeamNodeTeam); ok {
```
jradtilbrook, line 106: This shouldn't happen anyway, but if its not `ok`, we should error out and not set the state. I think that's the safest approach

buildkite/graphql/team.graphql
jradtilbrook: Theres a few things on the refactor list already, lets lump it on there. I _think_ the single `getNode` is the way to go

CHANGELOG.md
```diff
@@ -3,16 +3,21 @@
...
 * Convert DeletePipeline to genqlient [[PR #317](https://github.com/buildkite/terraform-provider-buildkite/pull/317)] @jradtilbrook
+* SUP-1049 Migrate Team to framework & Genqlient [[PR #318](https://github.com/buildkite/terraform-provider-buildkite/pull/317)] @mcncl
+
+### Forthcoming changes
+This release deprecates the `buildkite_organization_settings` [resource](./docs/resources/organization_settings.md). In a future minor release, we will remove this resource in favour of the newer `buildkite_organization` [resource](./docs/resources/organization.md) that aligns with the [datasource](./docs/data-sources/organization.md) of the same name. 
+[[PR #318](https://github.com/buildkite/terraform-provider-buildkite/pull/317)] introduces a change in how we handle
+
+Team data-sources importing Teams. Instead of `slug` the data-source will now require ID as the input. This is to avoid
+breaking changes which *may* be caused with the rename of a Team and the resulting change of `slug` value. GraphQL IDs
+of resources to not change, regardless of values within the resource being changed.
```
jradtilbrook, line 20: Ohh I misunderstood our chat. This isn't related to a `terraform import`, its related to how to identify the team for loading through a datasource.

In that case, I think it would be better not to use the word "importing" in this description because of the possibility of confusion there.  
I also think this will be a breaking change in the next release for anyone already using the datasource through slug -- and in that case, I'd like to support a deprecation/migration period and then switch over.

I actually think some other providers (eg github, aws) allow their datasources to use more than one identifier to load the resource and implement the searching functionality themselves. That way, users can use a human readable thing like slug to load it and then reference the ID of the datasource making a nicer UX. So we could perhaps support both and document the caveat when using slug?

jradtilbrook (changes requested):

buildkite/graphql/team.graphql
james2791: As long as it works in the end 👍

buildkite/data_source_team.go
```diff
@@ -2,84 +2,153 @@ package buildkite
...
-				Required: true,
-				Type:     schema.TypeString,
+			"uuid": schema.StringAttribute{
+				MarkdownDescription: "The UUID of the team.",
+				Computed:            true,
 			},
-			"uuid": {
-				Computed: true,
-				Type:     schema.TypeString,
+			"slug": schema.StringAttribute{
```
jradtilbrook, line 57: I reckon for all these slug, id, uuid attributes, to allow us to use any of them to find the team we should add validation that only one is supplied to use as well: https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework-validators@v0.10.0/stringvalidator#ExactlyOneOf

mcncl: Ha, knew there was a thing but tbh this has been such a pain to implement due to the Node thing too I forgot.

jradtilbrook: Yeah I bet its pretty crappy to implement. But I think its useful for users to use this way at least 🙏

james2791 (dismissed):
LGTM 🚀

docs/data-sources/team.md
```diff
@@ -8,14 +8,21 @@ Buildkite documentation: https://buildkite.com/docs/pipelines/permissions
 ## Example Usage
 
 ```hcl
-data "buildkite_team" "my_team" {
-  slug = "my_team"
+data "buildkite_team" "my_team_data" {
+    depends_on = [buildkite_team.my_team]
```
jradtilbrook, line 12: I dont think we should put this in the example. A datasource would be used when the resource itself is not created through terraform
```suggestion
```

docs/data-sources/team.md
```diff
@@ -8,14 +8,21 @@ Buildkite documentation: https://buildkite.com/docs/pipelines/permissions
...
+    id = buildkite_team.my_team.id
```
jradtilbrook, line 13: This can just be any string or a placeholder for this example. Something like: 
```suggestion
    id = "<team id>"
```

docs/data-sources/team.md
```diff
@@ -8,14 +8,21 @@ Buildkite documentation: https://buildkite.com/docs/pipelines/permissions
...
+The above will automatically fill the ID attribute with that of the team created by the resource, then fill the other
+attributes of that team.
+
+The datasource can also reference an imported Team. For more on using Import check
+[here](https://github.com/buildkite/terraform-provider-buildkite/blob/main/docs/resources/team.md#import).
+
 ## Argument Reference
 
-* `slug` - (Required) The slug of the team, available in the URL of the team on buildkite.com
+* `id` - (Required) The GraphQL ID of the team, available in the Settings page for the team.
```
jradtilbrook, line 25: Need to update the rest of this docs page after the changes to the code above. This should mention how either id or slug is required, but not both

jradtilbrook (changes requested):
Looking good, just a few comments

docs/data-sources/team.md
```diff
@@ -8,14 +8,25 @@ Buildkite documentation: https://buildkite.com/docs/pipelines/permissions
...
+    id = "<team id>"
 }
 ```
 
+The above will automatically fill the ID attribute with that of the team created by the resource, then fill the other
+attributes of that team.
+
+The datasource can also reference an imported Team. For more on using Import check
+[here](https://github.com/buildkite/terraform-provider-buildkite/blob/main/docs/resources/team.md#import).
+
```
jradtilbrook, line 21: Last thing: I don't think this section fits either. Importing and datasource are basically 2 sides of the same coin. You wouldn't import a resource and then use a datasource for the same resource. I think this entire section could probably be removed

jradtilbrook (changes requested):

jradtilbrook (approved):

james2791 (approved):

PR #319, opened 2023-07-21
SUP-1281: Pipeline resource ReadPipeline conversion to Genqlient

james2791:
## PR checklist:
- [x] `docs/` updated
- [x] tests added


buildkite/resource_pipeline.go
```diff
@@ -406,21 +406,27 @@ func CreatePipeline(ctx context.Context, d *schema.ResourceData, m interface{})
...
+	if pipelineNode, ok := apiResponse.GetNode().(*getNodeNodePipeline); ok {
+		if pipelineNode == nil {
+			return diag.FromErr(
+				fmt.Errorf("Unable to get pipeline with ID %s (%v)", d.Id(), err),
+			)
+		}
+		updatePipelineResourceNode(d, pipelineNode)
+	} else {
+		//clear state
+		d.SetId("")
```
jradtilbrook, line 427: I think this actually is the way to signal to terraform that the resource is gone. If you check out the docs for that function:

> // SetId sets the ID of the resource. If the value is blank, then the
> // resource is destroyed.

But I think you should write an extra test (or just change the disappears test) to verify it more correctly than just expecting a non-empty plan. Because that only really asserts that something in the resource changed whereas we want to be very sure that the plan is to recreate the pipeline

james2791: Seems like this case has been brought up [before](https://github.com/hashicorp/terraform-plugin-testing/issues/73) with Hashi. Same case with [framework](https://github.com/hashicorp/terraform-plugin-framework/issues/617) too

jradtilbrook: Actually, I think this should set it empty in the if block on line 403 as well right? Ie if the query succeeds but the node is null then it didnt find the pipeline so we should remove it from state?

jradtilbrook: i think from some minimal testing, in the case a node query returns no data, the json looks like `node:{}` which means it wont marshal into the pipeline node type and hit the else condition youve already got here. So I think we can call this done

jradtilbrook (approved):

PR #325, opened 2023-07-24
Add Terraform manifest.json file to root

mcncl:
- Add a manifest file for easier compatibility pickup
- Update CHANGELOG for manifest

## PR checklist:
- [-] `docs/` updated
- [-] tests added
- [-] an example added to `examples/` (useful to demo new field/resource)

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -34,6 +34,9 @@ archives:
 checksum:
   name_template: '{{ .ProjectName }}_{{ .Version }}_SHA256SUMS'
   algorithm: sha256
+  extra_files:
+    - glob: 'terraform-registry-manifest.json'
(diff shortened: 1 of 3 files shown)


jradtilbrook (changes requested):
The docs seem to say that the manifest file should be in the release assets so I think we need to update goreleaser as well to pull in that file. I'm not sure how to configure that though.

https://developer.hashicorp.com/terraform/registry/providers/publishing#terraform-registry-manifest-file

mcncl:
@jradtilbrook this seems to be how AWS do it: https://github.com/hashicorp/terraform-provider-aws/blob/be9d5ac322736b8c599459a755cb3d63be051e6c/.goreleaser.yml#L45

Implemented that change.

jradtilbrook (approved):

james2791 (approved):

PR #327, opened 2023-07-24
Add buildkite_test_suite resource

### LINKED ISSUE #299: Support for test analytics?
**Describe the bug**
Terraform provider doesn't support test analytics. 

**Expected behavior**

**Additional context**
We use the terraform provider for all our buildkite needs, and we've recently started looking into test analytics, but had to configure it through the UI online. We'd like to add management of our test analytics test suites to terraform.

jradtilbrook:
## PR checklist:
- [x] `docs/` updated
- [x] tests added

This resolves #299


buildkite/graphql/test_suite.graphql
```diff
@@ -0,0 +1,94 @@
+query getTestSuite($id: ID!, $teamCount: Int) {
```
jradtilbrook, line 1: This isn't using the `getNode` query because it needs the teams linked to the suite with a dynamic number of tests returned

jradtilbrook:
The proposed way this is currently implemented would like like this:

```hcl
resource "buildkite_test_suite" "rspec" {
  name = "rspec"
  default_branch = "main"
  team_owner_id = "<graphql id>"
}
```

So it has a single team associated with it. This is required by the API. Technically, the API allows adding more than one team to a suite upon creation, but all teams are given MANAGE_AND_READ access. 

We can use GraphQL to add other teams with either permission set after its created. So I'm proposing that for these other teams, we use a separate resource (to be implemented separately). Eg

```hcl
resource "buildkite_test_suite_team" "read_only" {
  team_id = "<graphql id>"
  suite_id = buildkite_test_suite.rspec.id
  access_level = "READ_ONLY"
}
```

I've also hardcoded resolving the graphql ID of the suite but I'm chatting with TA to see if they can add that to the REST response to make it more reliable

buildkite/resource_test_suite.go
```diff
@@ -0,0 +1,252 @@
...
+
+	state.ApiToken = types.StringValue(response.ApiToken)
+	state.DefaultBranch = types.StringValue(response.DefaultBranch)
+	state.ID = types.StringValue(response.GraphqlID)
+	state.UUID = types.StringValue(response.UUID)
+	state.Name = types.StringValue(response.Name)
+	state.Slug = types.StringValue(response.Slug)
+	state.TeamOwnerId = plan.TeamOwnerId
+
+	resp.Diagnostics.Append(resp.State.Set(ctx, state)...)
```
mcncl, line 99: Any reason this isn't `&state`, which seems to be the standard way we do this? This applies to all CRUD methods as far as I can tell.

buildkite/resource_test_suite.go
```diff
@@ -0,0 +1,252 @@
...
+		resp.Diagnostics.AddError("Failed to delete test suite", err.Error())
+		return
+	}
+}
+
+func (*testSuiteResource) Metadata(ctx context.Context, req resource.MetadataRequest, resp *resource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_test_suite"
+}
+
+func findTeamSuite(uuid string, suites []listTeamSuitesOrganizationSuitesSuiteConnectionEdgesSuiteEdge) *listTeamSuitesOrganizationSuitesSuiteConnectionEdgesSuiteEdgeNodeSuite {
```
mcncl, line 122: That's some name 😂

buildkite/provider.go
```diff
@@ -95,16 +95,17 @@ func (tf *terraformProvider) Metadata(ctx context.Context, req provider.Metadata
 	resp.Version = tf.version
 }
 
-func (*terraformProvider) Resources(context.Context) []func() resource.Resource {
+func (p *terraformProvider) Resources(context.Context) []func() resource.Resource {
```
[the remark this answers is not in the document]
jradtilbrook: I was mucking around with trying to configure support for proto v5/6 dynamically but don't need that now. I'll remove this

buildkite/resource_test_suite.go
```diff
@@ -0,0 +1,252 @@
...
+}
+
+type testSuiteResponse struct {
+	ApiToken      string `json:"api_token"`
+	DefaultBranch string `json:"default_branch"`
+	UUID          string `json:"id"`
+	GraphqlID     string `json:"graphql_id"`
+	Name          string `json:"name"`
+	Slug          string `json:"slug"`
+	TeamUuids     string `json:"team_ids"`
```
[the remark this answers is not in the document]
jradtilbrook: Everything in the rest api uses an `ID` name but its actually a UUID. However, this isn't actually used so I'll just remove it

buildkite/resource_test_suite.go
jradtilbrook: Hmm that is probably a copy-paste from the `State.Get()` but not having to copy a struct around is good so I'll change it

buildkite/resource_test_suite.go
jradtilbrook: No longer used actually so I'll remove it

mcncl (dismissed):
Looks good to me, especially given the current frailties of the TA API.

buildkite/resource_test_suite_test.go
```diff
@@ -0,0 +1,175 @@
...
+				`,
+				Check: resource.ComposeAggregateTestCheckFunc(
+					resource.TestCheckResourceAttrSet("buildkite_test_suite.suite", "id"),
+					resource.TestCheckResourceAttrSet("buildkite_test_suite.suite", "api_token"),
+					resource.TestCheckResourceAttr("buildkite_test_suite.suite", "default_branch", "main"),
+					resource.TestCheckResourceAttr("buildkite_test_suite.suite", "name", "test suite"),
+					resource.TestCheckResourceAttrSet("buildkite_test_suite.suite", "team_owner_id"),
+					checkTestSuiteExists("buildkite_test_suite.suite"),
+					checkTestSuiteRemoteValue("buildkite_test_suite.suite", "Name", "test suite"),
+					checkTestSuiteRemoteValue("buildkite_test_suite.suite", "DefaultBranch", "main"),
```
jradtilbrook, line 43: I've implemented remote testing in this way to see how it feels. Instead of the first function here setting a pointer variable and remaining functions using that pointer to check values, it retrieves the suite each time. It's quite slow and I'm not sure I like it much. Let me know what you think

james2791 (dismissed):
Given current API implementations, this looks great

buildkite/resource_test_suite_test.go
mcncl: I don't mind it, but if one of our pieces of upcoming work is speeding up the tests and properly implementing `Parallel()` then opting for a testing approach which is slower might go against this.

I'm indifferent to whether it stays in or changes to a pointer approach, but it feels like that's what's going to happen anyway and it might be good to have that implemented so that core tests don't need to be changed, merely out use of `Parallel`.

jradtilbrook: Yeah I don't like it very much. I think there is some improvement we could do to the testing structure like this but not quite this. I think it would still involve pointers.
I'll convert this back to the other approach

james2791 (approved):
LGTM

PR #339, opened 2023-07-31
SUP-196: Pipeline Schedule GraphQL transition

### LINKED ISSUE #143: Standardise IDs used for importing resources
In #140 @thomasvdpol observed that we're inconsistent with the type of ID we use for importing resources. Some use the graphql ID, some use the UUID. It'd be nice to standardise on a single type of ID (or consistently support both I guess).

If we pick one, we should gracefully transition and support both kinds of ID on the resources that need to change for a while.

Here's the ID types we currently require:

* agent_token: graphql ID
* pipeline: graphql ID
* pipeline_schedule: `<org slug>/<pipeline slug>/<schedule uuid>`
* team: graphql ID

james2791:
## PR checklist:
- [x] `docs/` updated
- [x] tests added

Added deprecation warnings to necessary examples/docs as well as changelog

Resolves #143


buildkite/resource_pipeline_schedule.go
```diff
@@ -249,36 +249,41 @@ func (ps *pipelineSchedule) Delete(ctx context.Context, req resource.DeleteReque
...
-			fmt.Sprintf("Unable to read Pipeline schedule: %s", err.Error()),
+			"Missing Import Identifier",
+			fmt.Sprintf("Import identifier is expected with format: $BUILDKITE_ORGANIZATION_SLUG/$BUILDKITE_PIPELINE_SLUG/$PIPELINE_SCHEDULE_UUID or PIPELINE_SCHEDULE_ID. Got: %q", req.ID),
 		)
 		return
 	}
-
-	var newReq resource.ImportStateRequest
-	newReq.ID = apiResponse.PipelineSchedule.Id
-	resource.ImportStatePassthroughID(ctx, path.Root("id"), newReq, resp)
```
jradtilbrook, line 278: FWIW importing a resource is a manual process that I think if we just change straight to graphql ID it would be fine. We don't need to support the slug method as well. For people adopting terraform and importing resources, they are highly likely to check out our docs every time anyway. So I think this could technically be a one-line function of just this line

james2791: Nice - I guess supporting the older way was more "nice to have" but can whisk this down to just the passthrough 👍

mcncl: Something like the below will make this backwards compatible:

```go
func (ps *pipelineSchedule) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
	if strings.HasPrefix(req.ID, ps.client.organization) {
		importOrg := strings.Split(req.ID, "/")[0]
		if importOrg != ps.client.organization {
			resp.Diagnostics.AddError(
				"Import Organization does not match",
				fmt.Sprintf("Mismatch in import organization. Expected %s, got %s", ps.client.organization, importOrg),
			)
			return
		}
		apiResponse, err := getPipelineScheduleBySlug(
			ps.client.genqlient,
			req.ID,
		)

		if err != nil {
			resp.Diagnostics.AddError(
				"Unable to read Pipeline schedule",
				fmt.Sprintf("Unable to read Pipeline schedule: %s", err.Error()),
			)
			return
		}

		req.ID = apiResponse.PipelineSchedule.Id
		resource.ImportStatePassthroughID(ctx, path.Root("id"), req, resp)
		return
	}
	resource.ImportStatePassthroughID(ctx, path.Root("id"), req, resp)
}
```

mcncl:
Be awesome to see the `RandomString` stuff to be implemented too to save name clashes.

jradtilbrook:
> Be awesome to see the `RandomString` stuff to be implemented too to save name clashes.

I agree, since the changes here are actually quite small (I didn't realise). You could also look at parallelising the tests using the random string @james2791

buildkite/resource_pipeline_schedule.go
james2791: Going for the later - excluding the backwards-compatible implementation (I did it with an import components check but one-liner with the GQL ID is much more simple 👍)

jradtilbrook (approved):

mcncl (approved):

PR #341, opened 2023-08-01
Add deprecation warning. Update CHANGELOG.

mcncl:
## PR checklist:
- [x] `docs/` updated
- [x] tests added

### Proposed Changes
Mark `deletion_protection` as `Deprecated` in favour of moving towards the use of [lifecycles](https://developer.hashicorp.com/terraform/language/meta-arguments/lifecycle).

This enables us to continue to support the attribute for now, but gives warning to users that it's been deprecated. The deprecation `message` links to lifecycles.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 
 All notable changes to this project will be documented in this file.
+### Forthcoming Changes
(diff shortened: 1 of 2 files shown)


james2791 (approved):

jradtilbrook (changes requested):
Just need to update it in the docs as well. Mention its deprecated and give an example with lifecycle

PR #344, opened 2023-08-01
Added note on write_suites API access token scope to README

james2791:
With the addition of the Test Suite resource in [v0.22.0](https://github.com/buildkite/terraform-provider-buildkite/compare/v0.21.2...v0.22.0), one's API access token that is used to interact with the provider needs to additionally have the `write_suites` scope for said API token to create/update and delete suites (interaction with the REST API)

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,9 +9,9 @@ The provider allows you to manage resources in your Buildkite organization.
 Two configuration values are required:
 
 -   An API token, generated at https://buildkite.com/user/api-access-tokens. The
-    token must have the `write_pipelines, read_pipelines` REST API scopes and be enabled for GraphQL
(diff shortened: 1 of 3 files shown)


jradtilbrook:
@james2791 I think this is documented in a few places that will need updating as well. And to confirm, it doesn't need `read_suites` because it uses GraphQL for that side of things right?

james2791:
> @james2791 I think this is documented in a few places that will need updating as well. And to confirm, it doesn't need `read_suites` because it uses GraphQL for that side of things right?

Just added - correct too: the `Read` of the Test suite resource being node based,. Its more for the C-U-D ops otherwise that interacts with REST

jradtilbrook (approved):

Issue #194, opened 2022-04-21
Invalid states


keith:
I noticed when managing pipelines via terraform you can get yourself into invalid states that buildkite doesn't actually support. I think it would be ideal if it would error in this case instead. For example:

```
  provider_settings {
    trigger_mode = "none"
    build_branches = false
    build_pull_requests = false
    publish_commit_status = true
    cancel_deleted_branch_builds = true
  }
```

Using `publish_commit_status` and `cancel_deleted_branch_builds` here seems to be invalid when `trigger_mode = "none"`. You can pretty much see this in the UI because those checkboxes aren't exposed in this case:

<img width="786" alt="image" src="https://user-images.githubusercontent.com/283886/164513064-95e1e7f9-3210-4939-a920-cd72b05038bf.png">

And when you change the trigger mode in the UI you can see the checkboxes are set correctly, just likely ignored:

<img width="455" alt="image" src="https://user-images.githubusercontent.com/283886/164513148-d7817121-006d-4e50-be5b-a467b95256da.png">

In this case I think it's safe to change `trigger_mode = "code"` because we actually still don't trigger in any case because of the other settings, but I think it would be better UX if buildkite errored in this case instead so you couldn't accidentally end up here.

jradtilbrook:
Thanks for reporting this 👍 

We are approaching a v1.0 release and have done major refactors to the provider_settings part of the provider. I don't believe this is a relevant bug any more. If in fact this comes up again after the v1.0 release, please reopen it

PR #345, opened 2023-08-02
SUP-1068 Migrate pipeline resource to framework

jradtilbrook:
## PR checklist:
- [x] `docs/` updated
- [x] tests added

Theres a few things I want to test manually
- [x] changing cluster from empty string to null
- [x] provider_settings is technically a list block but we shouldn't allow that. what happens if its just changed under the hood
- [x] ... TBD


buildkite/resource_organization_settings_test.go
```diff
@@ -86,24 +86,6 @@ func TestAccOrganizationSettings_import(t *testing.T) {
...
-	resource.Test(t, resource.TestCase{
-		PreCheck:                 func() { testAccPreCheck(t) },
-		ProtoV6ProviderFactories: protoV6ProviderFactories(),
-		CheckDestroy:             testCheckOrganizationSettingsResourceRemoved,
-		Steps: []resource.TestStep{
-			{
-				Config: testAccOrganizationSettingsConfigBasic([]string{"0.0.0.0/0", "1.1.1.1/32", "1.0.0.1/32"}),
-				Check: resource.ComposeAggregateTestCheckFunc(
-					// Confirm that the allowed IP addresses are set correctly in Buildkite's system
-					testAccCheckResourceDisappears(Provider("testing"), resourceOrganizationSettings(), "buildkite_organization_settings.let_them_in"),
```
jradtilbrook, line 99: I've removed this whole test because I've refactored this function. This resource is going away soon anyway so we'll be deleting everything. I figure that removing this a little early is probably fine as well

buildkite/resource_pipeline_test.go
```diff
@@ -105,7 +105,7 @@ func TestAccPipeline_add_remove_withcluster(t *testing.T) {
 					testAccCheckPipelineRemoteValues(&resourcePipeline, "Test Pipeline foo"),
 					// Confirm the pipeline has the correct values in terraform state
 					resource.TestCheckResourceAttr("buildkite_pipeline.foobar", "name", "Test Pipeline foo"),
-					resource.TestCheckResourceAttr("buildkite_pipeline.foobar", "cluster_id", ""),
+					resource.TestCheckNoResourceAttr("buildkite_pipeline.foobar", "cluster_id"),
```
jradtilbrook, line 108: The resource now sets this as null instead of an empty string

buildkite/resource_pipeline.go
```diff
@@ -61,479 +71,570 @@ type TeamPipelineNode struct {
...
-			"deletion_protection": {
-				Optional: true,
-				Default:  false,
-				Deprecated: "Deletion protection will be removed in a future release. A similar solution already " +
-					"exists and is supported by Terraform. See https://developer.hashicorp.com/terraform/language/meta-arguments/lifecycle.",
-				Type:        schema.TypeBool,
+			"deletion_protection": schema.BoolAttribute{
+				Optional:    true,
+				Computed:    true,
+				Default:     booldefault.StaticBool(false),
```
mcncl, line 357: This should be marked as `Deprecated` by setting `Deprecated: <reason>`

buildkite/resource_pipeline.go
```diff
@@ -61,479 +71,570 @@ type TeamPipelineNode struct {
...
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
+		},
+		Blocks: map[string]schema.Block{
+			"team": schema.SetNestedBlock{
+				NestedObject: schema.NestedBlockObject{
+					Attributes: map[string]schema.Attribute{
+						"slug": schema.StringAttribute{
```
mcncl, line 424: `team` object uses `id` and `access_level`, why not accept that? Makes it easier to have the inheritance in the config anyway.

buildkite/resource_pipeline.go
```diff
@@ -61,479 +71,570 @@ type TeamPipelineNode struct {
...
-								case "fork":
-								case "none":
-									return
-								default:
-									errs = append(errs, fmt.Errorf("%q must be one of code, deployment, fork or none, got: %s", key, v))
-									return
-								}
-								return
+							Validators: []validator.String{
+								stringvalidator.OneOf("code", "deployment", "fork", "none"),
```
mcncl, line 449: There **must** be a way to stop people filling the settings if not `code`

buildkite/resource_pipeline.go
jradtilbrook: The slug and access_level matches the SDKv2 version of this resource. I've only added team_id and pipeline_team_id as computed to make the code simpler. Not sure why I didn't originally do this. It'll be transparent to the user. It'll automatically update those fields in state when they run a terraform refresh but it makes the code much easier to work with.

And this will be going away soon as well anyway

buildkite/resource_pipeline.go
jradtilbrook: Yeah maybe if we write our own validator. But with this PR I want to mimic as close as possible (exactly) the SDKv2 behaviour. We can revisit improving this block when resolving #194

buildkite/resource_pipeline.go
mcncl: Yeah, sweet. I think we're moving away from anything _but_ ID for, though we may support additional stuff for data source.

mcncl (dismissed):

buildkite/resource_organization_settings_test.go
james2791: I think it'll be a similar story for all the SDKv2 implemented resources - but thats fair

james2791 (dismissed):
💪🚀

buildkite/resource_pipeline_test.go
```diff
@@ -105,7 +105,54 @@ func TestAccPipeline_add_remove_withcluster(t *testing.T) {
...
+				),
+			},
+		},
+	})
+}
+
+// TestAccPipeline_add_remove_withcluster_old_version tests that the provider doesn't get stuffed up when coming from an
+// older version to this new version. the first step runs against the published v0.21.1 version and the second step runs
+// against the current provider. the configuration is the same between each one so there should be no state/plan change
+func TestAccPipeline_add_remove_withcluster_old_version(t *testing.T) {
```
jradtilbrook, line 119: I'd like specific and thorough review of these additional tests. Their aim is ensure that the state remains the same during the upgrade path of users. The first step runs against the published version of the provider. The second step uses the same config but runs against the framework version of the resource

buildkite/resource_pipeline_test.go
```diff
@@ -109,6 +189,20 @@ func TestAccPipeline_add_remove_withcluster(t *testing.T) {
...
+				Config:                   testAccPipelineConfigBasic("foo"),
+				ProtoV6ProviderFactories: protoV6ProviderFactories(),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					// Confirm the pipeline exists in the buildkite API
+					testAccCheckPipelineExists("buildkite_pipeline.foobar", &resourcePipeline),
+					// Confirm the pipeline has the correct values in Buildkite's system
+					testAccCheckPipelineRemoteValues(&resourcePipeline, "Test Pipeline foo"),
+					// Confirm the pipeline has the correct values in terraform state
+					resource.TestCheckResourceAttr("buildkite_pipeline.foobar", "name", "Test Pipeline foo"),
+					resource.TestCheckNoResourceAttr("buildkite_pipeline.foobar", "cluster_id"),
```
jradtilbrook, line 202: You can see the state change here. In the previous versions of the provider, if no cluster is attached to the pipeline, the `cluster_id=""` in state. But in this version of the provider it is set to null. Its effectively the same and upon a `terraform refresh` the state will be updated under-the-hood for users

buildkite/resource_pipeline_test.go
james2791: Sound - given `0.21.1` containing SDKv2 implementation - ensuring the later versions as the provider stays consistent and retains the functionality upon what we expect to happen between old/new

buildkite/resource_pipeline_test.go
james2791: Noting ` testAccPipelineConfigBasicWithTeam("foo")` clears the cluster ID off the pipeline - and new impl sets it as null - this LG

buildkite/resource_pipeline_test.go
james2791: s above with my commentary 👍

buildkite/resource_pipeline_test.go
```diff
@@ -98,6 +98,127 @@ func TestAccPipeline_add_remove_withcluster(t *testing.T) {
...
+					// Confirm the pipeline exists in the buildkite API
+					testAccCheckPipelineExists("buildkite_pipeline.foobar", &resourcePipeline),
+					// Confirm the pipeline has the correct values in Buildkite's system
+					testAccCheckPipelineRemoteValues(&resourcePipeline, "Test Pipeline foo"),
+					// Confirm the pipeline has the correct values in terraform state
+					resource.TestCheckResourceAttr("buildkite_pipeline.foobar", "name", "Test Pipeline foo"),
+					resource.TestCheckResourceAttr("buildkite_pipeline.foobar", "cluster_id", "Q2x1c3Rlci0tLTFkNmIxOTg5LTJmYjctNDRlMC04MWYyLTAxYjIxNzQ4MTVkMg=="),
+					resource.TestCheckResourceAttr("buildkite_pipeline.foobar", "allow_rebuilds", "true"),
+					// check provider settings are present
+					resource.TestCheckResourceAttr("buildkite_pipeline.foobar", "provider_settings.0.trigger_mode", "code"),
```
jradtilbrook, line 160: This asserts that provider settings are present in state. There is a separate test that ensures its present in state when its present in the resource attributes. But this test doesn't specify the values so it doesn't put it into state. This test is currently failing because of that.

Previous versions of the provider put this into state regardless of whether the user provided a `provider_settings` block. This is somewhat an artifact/bug of the old SDKv2 that this was supposed to be required and makes sense for it to be required I think. The API will set default values on the pipeline settings when it creates a pipeline, but those aren't properly reflected in state and is the source of a few open issues we have currently.

I think long term, provider settings should be made required and not a list but a single block attribute (or nested attribute in protocol v6). I'm unsure whether to do that as part of this or at all. Or whether to achieve parity with the SDK version of this resource.

jradtilbrook: I think the way to approach this is to put as much effort into parity here with just the migration from SDKv2 to framework. But I think its worth looking at the state schema afterwards to clean it up, prior to a v1 release of the provider. Hopefully the new state migration functionality in framework will make that easier.

mcncl: > I think long term, provider settings should be made required

I don't agree with this, or, I think it's maybe the easiest approach but not the most user focused; making them required is fine in a sense that we're then able to be hands off, in a way, but folks who won't be using `code`, or source control for the trigger will find this convoluted. I think maybe we can find an alternative solution, though that may require additional engineering in the GraphQL API

buildkite/resource_pipeline_test.go
```diff
@@ -157,7 +157,7 @@ func TestAccPipeline_add_remove_withcluster_old_version(t *testing.T) {
 					resource.TestCheckResourceAttr("buildkite_pipeline.foobar", "cluster_id", "Q2x1c3Rlci0tLTFkNmIxOTg5LTJmYjctNDRlMC04MWYyLTAxYjIxNzQ4MTVkMg=="),
 					resource.TestCheckResourceAttr("buildkite_pipeline.foobar", "allow_rebuilds", "true"),
 					// check provider settings are not present
-					resource.TestCheckNoResourceAttr("buildkite_pipeline.foobar", "provider_settings"),
+					resource.TestCheckNoResourceAttr("buildkite_pipeline.foobar", "provider_settings.0"),
```
mcncl, line 160: Would this be better off checking the length of the list?

```go
TestCheckResourceAttr("buildkite_pipeline.foobar", "provider_settings.#", "0")
```

The above would check that the length is actually `0` and remove the risk that _something_ is set at `1`.

Ref: https://developer.hashicorp.com/terraform/plugin/sdkv2/testing/acceptance-tests/teststep#builtin-check-functions

buildkite/resource_pipeline_test.go
```diff
@@ -208,7 +208,7 @@ func TestAccPipeline_add_remove_withoutcluster_old_version(t *testing.T) {
 					resource.TestCheckNoResourceAttr("buildkite_pipeline.foobar", "cluster_id"),
 					resource.TestCheckResourceAttr("buildkite_pipeline.foobar", "allow_rebuilds", "true"),
 					// check provider settings are not present
-					resource.TestCheckNoResourceAttr("buildkite_pipeline.foobar", "provider_settings"),
+					resource.TestCheckNoResourceAttr("buildkite_pipeline.foobar", "provider_settings.0"),
```
mcncl, line 211: See above.

mcncl (changes requested):

buildkite/graphql/pipeline.graphql
```diff
@@ -52,60 +54,17 @@ query getPipeline($slug: ID!) {
     }
 }
 
+# @genqlient(for: "PipelineCreateInput.branchConfiguration", pointer: true)
 # @genqlient(for: "PipelineCreateInput.clusterId", pointer: true)
-# @genqlient(for: "PipelineCreateInput.visibility", omitempty: true)
-# @genqlient(for: "PipelineCreateInput.pipelineTemplateId", omitempty: true)
 # @genqlient(for: "PipelineCreateInput.nextBuildNumber", omitempty: true)
+# @genqlient(for: "PipelineCreateInput.pipelineTemplateId", omitempty: true)
+# @genqlient(for: "PipelineCreateInput.visibility", omitempty: true)
```
mcncl, line 61: ++ `# @genqlient(for: "PipelineCreateInput.defaultTimeoutInMInutes", pointer: true)`

mcncl: Same on Update

buildkite/resource_pipeline.go
```diff
@@ -61,479 +71,578 @@ type TeamPipelineNode struct {
...
+	}
+
+	input := PipelineCreateInput{
+		AllowRebuilds:                        plan.AllowRebuilds.ValueBool(),
+		BranchConfiguration:                  plan.BranchConfiguration.ValueStringPointer(),
+		CancelIntermediateBuilds:             plan.CancelIntermediateBuilds.ValueBool(),
+		CancelIntermediateBuildsBranchFilter: plan.CancelIntermediateBuildsBranchFilter.ValueString(),
+		ClusterId:                            plan.ClusterId.ValueStringPointer(),
+		DefaultBranch:                        plan.DefaultBranch.ValueString(),
+		DefaultTimeoutInMinutes:              int(plan.DefaultTimeoutInMinutes.ValueInt64()),
```
mcncl, line 191: This will be a value pointer I think (with changes made in GQL)

buildkite/resource_pipeline.go
```diff
@@ -61,479 +71,578 @@ type TeamPipelineNode struct {
...
-		log.Printf("Unable to create pipeline %s", d.Get("name"))
-		return diag.FromErr(err)
+	input := PipelineUpdateInput{
+		AllowRebuilds:                        plan.AllowRebuilds.ValueBool(),
+		BranchConfiguration:                  plan.BranchConfiguration.ValueStringPointer(),
+		CancelIntermediateBuilds:             plan.CancelIntermediateBuilds.ValueBool(),
+		CancelIntermediateBuildsBranchFilter: plan.CancelIntermediateBuildsBranchFilter.ValueString(),
+		ClusterId:                            plan.ClusterId.ValueStringPointer(),
+		DefaultBranch:                        plan.DefaultBranch.ValueString(),
+		DefaultTimeoutInMinutes:              int(plan.DefaultTimeoutInMinutes.ValueInt64()),
```
mcncl, line 558: Probably a pointer.

buildkite/resource_pipeline.go
```diff
@@ -61,479 +71,578 @@ type TeamPipelineNode struct {
...
-		if err != nil {
-			return diag.FromErr(err)
+func setPipelineModel(model *pipelineResourceModel, data pipelineResponse) {
+	model.AllowRebuilds = types.BoolValue(data.GetAllowRebuilds())
+	model.BranchConfiguration = types.StringPointerValue(data.GetBranchConfiguration())
+	model.CancelIntermediateBuilds = types.BoolValue(data.GetCancelIntermediateBuilds())
+	model.CancelIntermediateBuildsBranchFilter = types.StringValue(data.GetCancelIntermediateBuildsBranchFilter())
+	model.ClusterId = types.StringPointerValue(data.GetCluster().Id)
+	model.DefaultBranch = types.StringValue(data.GetDefaultBranch())
+	model.DefaultTimeoutInMinutes = types.Int64Value(int64(data.GetDefaultTimeoutInMinutes()))
```
mcncl, line 614: Pointer here too

mcncl (commented):
Given https://3.basecamp.com/3453178/buckets/23112918/card_tables/cards/6193536361#__subscriptions, we can now make the default timeout a pointer as it can return `null` in the API

jradtilbrook:
Alright I think this is ready for review again. With framework, when defining a block attribute you cannot set it as required or a default value. And the framework also performs validation on the state. This causes trouble for the `provider_settings` block. It is optional which means that the provider cannot set any values into state if the user has not defined it. The SDKv2 version of the provider would still perform some actions using default values which is not possible now. And I cannot find a way around it. 

I think long-term, moving to protocol v6 and converting it to a nested attribute is the way to go

mcncl:
@jradtilbrook I think maybe we let Pagerduty know they need to look at updating sooner rather than later?

jradtilbrook:
> @jradtilbrook I think maybe we let Pagerduty know they need to look at updating sooner rather than later?

@mcncl yeah I think so. The good news is that the latest release has basically all the new features we were aiming for for v1. Its just some bug fixes from here. So if they can pin their version now, they won't really miss out on anything (especially if they haven't hit the cases for the bugs)

jradtilbrook:
Actually, lets land #351 first because that will be useful to start migrating. And then we can finish this PR off

james2791 (approved):
I think this is at a point of rolling-in, LGTM 👍

buildkite/graphql/pipeline.graphql
```diff
@@ -1,13 +1,16 @@
...
     cancelIntermediateBuilds
     cancelIntermediateBuildsBranchFilter
     cluster {
+        # @genqlient(pointer: true)
         id
     }
     defaultBranch
+    # @genqlient(pointer: true)
     defaultTimeoutInMinutes
     maximumTimeoutInMinutes
```
mcncl, line 15: This can also be a pointer, I think? `null` is a value option on the `query`.

mcncl: NB: This comment includes changes to other calls in the query and changing to `ValuePointer` as a result in the code.

buildkite/resource_pipeline.go
```diff
@@ -61,479 +73,599 @@ type TeamPipelineNode struct {
...
+type pipelineResponse interface {
+	GetId() string
+	GetAllowRebuilds() bool
+	GetBranchConfiguration() *string
+	GetCancelIntermediateBuilds() bool
+	GetCancelIntermediateBuildsBranchFilter() string
+	GetCluster() PipelineValuesCluster
+	GetDefaultBranch() string
+	GetDefaultTimeoutInMinutes() *int
+	GetMaximumTimeoutInMinutes() int
```
mcncl, line 144: Adding a note here for start of changes to this being a pointer.

mcncl (changes requested):

mcncl (approved):

james2791 (approved):

PR #346, opened 2023-08-02
SUP-1305: Test Suite Team Resource addition

james2791:
## PR checklist:
- [x] `docs/` updated
- [x] tests added


buildkite/resource_test_suite_team.go
```diff
@@ -0,0 +1,203 @@
...
+func (tst *testSuiteTeamResource) Read(ctx context.Context, req resource.ReadRequest, resp *resource.ReadResponse) {
+	var state testSuiteTeamModel
+
+	resp.Diagnostics.Append(req.State.Get(ctx, &state)...)
+
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	log.Printf("Reading test suite team with ID %s ...", state.ID.ValueString())
```
mcncl, line 123: When does this ever get output?

buildkite/resource_test_suite_team.go
```diff
@@ -0,0 +1,203 @@
...
+}
+
+func (tst *testSuiteTeamResource) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) {
+	var state testSuiteTeamModel
+	var testSuiteTeamAccessLevel string
+
+	resp.Diagnostics.Append(req.Plan.GetAttribute(ctx, path.Root("access_level"), &testSuiteTeamAccessLevel)...)
+	resp.Diagnostics.Append(req.State.Get(ctx, &state)...)
+
+	log.Printf("Updating team %s in test suite %s to %s ...", state.TeamID.ValueString(), state.TestSuiteId.ValueString(), testSuiteTeamAccessLevel)
```
mcncl, line 159: Same, when will this get output? It doesn't during an `apply` as it's not a Diagnostic.

buildkite/resource_test_suite_team.go
```diff
@@ -0,0 +1,203 @@
...
+func (tst *testSuiteTeamResource) Delete(ctx context.Context, req resource.DeleteRequest, resp *resource.DeleteResponse) {
+	var state testSuiteTeamModel
+
+	resp.Diagnostics.Append(req.State.Get(ctx, &state)...)
+
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	log.Printf("Deleting team %s's access to test suite %s ...", state.TeamID.ValueString(), state.TestSuiteId.ValueString())
```
mcncl, line 189: As above

examples/main.tf
```diff
@@ -8,6 +8,14 @@ resource "buildkite_team" "test" {
   default_member_role = "MEMBER"
 }
 
+resource "buildkite_team" "testtwo" {
```
mcncl, line 11: IS there any reason for this example resource instead of just using `test` for the `suite` too?

james2791: More to simulate that a team not owning the suite (`team_owner_id` is added as either a RO/manage&read permissions) - we'd get the "already added to the suite" message if the owning team is attempted to be assigned to their own suite 👍

buildkite/resource_test_suite_team.go
james2791: TF_LOG at INFO and higher if its set. More to catch cases if thats set 👍

buildkite/resource_test_suite_team.go
james2791: Above ⬆️ - more a TF_LOG thing

buildkite/resource_test_suite_team.go
james2791: Above ⬆️ - more a TF_LOG thing

buildkite/resource_test_suite_team.go
```diff
@@ -0,0 +1,207 @@
...
+		if teamSuiteNode == nil {
+			resp.Diagnostics.AddError(
+				"Unable to get test suite team",
+				"Error getting test suite team: nil response",
+			)
+			return
+		}
+		updateTeamSuiteTeamResource(&state, *teamSuiteNode)
+		resp.Diagnostics.Append(resp.State.Set(ctx, &state)...)
+	}
```
jradtilbrook, line 145: I think this needs to handle the case where the team is removed outside of terraform right? Its usually covered by the tests `TestAcc.*disappears()`

jradtilbrook (changes requested):

buildkite/resource_test_suite_team.go
james2791: How could I forget the SetId! Adding...

jradtilbrook (approved):

jradtilbrook (approved):

PR #348, opened 2023-08-04
refactor: use Conventional Commits for PR template

mcncl:
refactor: use Conventional Commits for PR template

[Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) strives to make commits more human readable and contain relevant information within. Marking breaking changes and better grouping of the contents


mcncl:
This is breaking in CI because of current broken code in CI. Will merge in `main`once that's fixed.

james2791 (approved):

PR #351, opened 2023-08-08
SUP-1307 Implement Pipeline Team Resource

lizrabuya:
## PR checklist:
- [ ] `docs/` updated
- [x] tests added

Tests to fix (currently skipped)
- [x] TestAccPipelineTeam_update


buildkite/resource_pipeline_team.go
```diff
@@ -0,0 +1,250 @@
...
+		)
+		return
+	}
+
+	// Update state with values from API response/plan
+	state.Id = types.StringValue(apiResponse.TeamPipelineCreate.TeamPipelineEdge.Node.Id)
+	state.Uuid = types.StringValue(apiResponse.TeamPipelineCreate.TeamPipelineEdge.Node.Uuid)
+	state.PipelineId = types.StringValue(apiResponse.TeamPipelineCreate.TeamPipelineEdge.Node.Pipeline.Id)
+	state.TeamId = types.StringValue(apiResponse.TeamPipelineCreate.TeamPipelineEdge.Node.Team.Id)
+	state.AccessLevel = types.StringValue(string(apiResponse.TeamPipelineCreate.TeamPipelineEdge.Node.PipelineAccessLevel))
```
mcncl, line 110: I don't think these need to be in here; they're known at run time and so can just be a part of the `resp.State.Set`.

mcncl (changes requested):
Marking as changes requested so it doesn't get merged with the test being skipped 🙂

Looks great though!

buildkite/resource_pipeline_team.go
```diff
@@ -0,0 +1,250 @@
...
+			"team_id": resource_schema.StringAttribute{
+				Required:            true,
+				MarkdownDescription: "The GraphQL ID of the team.",
+			},
+			"pipeline_id": resource_schema.StringAttribute{
+				Required:            true,
+				MarkdownDescription: "The GraphQL ID of the pipeline.",
+			},
+			"access_level": resource_schema.StringAttribute{
+				Required:            true,
```
jradtilbrook, line 71: I notice the graphql mutation to create a team pipeline has `omitempty: true` for the access level. I assume thats because a team has a default member access level so its not necessary. But this marks it as required. So I think either the omitempty is wrong or this should be optional?

buildkite/resource_pipeline_team.go
```diff
@@ -0,0 +1,250 @@
...
+				MarkdownDescription: "The access level for the team. Either READ_ONLY, BUILD_AND_READ or MANAGE_BUILD_AND_READ.",
+				Validators: []validator.String{
+					stringvalidator.OneOf("READ_ONLY", "BUILD_AND_READ", "MANAGE_BUILD_AND_READ"),
```
jradtilbrook, line 74: These values come from the graphql generated.go file as constants so we should use those instead

buildkite/resource_pipeline_team.go
jradtilbrook: Yeah I think you only need to set id and uuid here right? Maybe access level depending on whether its required or not (as above)

buildkite/resource_pipeline_team.go
```diff
@@ -0,0 +1,250 @@
...
+			PipelineAccessLevels(accessLevel),
+		)
+
+		if err != nil {
+			resp.Diagnostics.AddError(
+				"Unable to create team pipeline",
+				fmt.Sprintf("Unable to create team pipeline: %s", err.Error()),
+			)
+			return
+		}
```
jradtilbrook, line 185: This is an interesting case. Whats the reason for finding it first? If it doesn't exist it should be the job of the update to create it, that will be handled by terraform itself which will instead call create

jradtilbrook (changes requested):

buildkite/resource_pipeline_team.go
lizrabuya: I have removed the `omitempty: true` in the graphql

jradtilbrook:
I think what we need to apply to this PR now is:

- update the `team` block on pipeline resource to ignore unknown teams returned from the API. This will allow people to make use of `pipeline_team` without it interfering with `pipeline.team`
- if any existing teams in the `pipeline.team` block have changed or been removed, update terraform state to match
- deprecate the `team` block on pipeline, referring to the `pipeline_team` resource instead

This means that terraform will no longer do anything about new teams it finds but that is a defacto standard for other providers and how this will work in the future once the `team` block is removed in favour of the separate resource

CHANGELOG.md
```diff
@@ -13,6 +15,12 @@ All notable changes to this project will be documented in this file.
 - Bump github.com/hashicorp/terraform-plugin-framework-validators [[PR #350](https://github.com/buildkite/terraform-provider-buildkite/pull/350)]
 - refactor 🧹: Refactor templates to use Conventional Commits[[PR #348](https://github.com/buildkite/terraform-provider-buildkite/pull/348)] @mcncl
 
+### Forthcoming Changes
+`deletion_protection` is being deprecated and will be removed in a future release (`v1`). This feature offers similar
+functionality to [lifecycles](https://developer.hashicorp.com/terraform/language/meta-arguments/lifecycle) which are supported by Terraform.
+
+=======
+
```
jradtilbrook, line 23: This whole section has somehow made its way here but its already present on lines 39-41 below
```suggestion
```

buildkite/resource_pipeline_team.go
jradtilbrook: eg `PipelineAccessLevelsManageBuildAndRead`

buildkite/resource_pipeline_team.go
```diff
@@ -0,0 +1,232 @@
...
+
+func updateTeamPipelineResourceState(tpState *pipelineTeamResourceModel, tpNode getNodeNodeTeamPipeline) {
+	tpState.Id = types.StringValue(tpNode.Id)
+	tpState.Uuid = types.StringValue(tpNode.Uuid)
+	tpState.TeamId = types.StringValue(tpNode.Team.Id)
+	tpState.PipelineId = types.StringValue(tpNode.Pipeline.Id)
+	tpState.AccessLevel = types.StringValue(string(tpNode.PipelineAccessLevel))
+}
+
+func findTeamInPipelineNode(client *graphql.Client, tpState *pipelineTeamResourceModel) bool {
```
jradtilbrook, line 216: This looks unused now. Can it be removed?

buildkite/resource_pipeline_team_test.go
```diff
@@ -0,0 +1,226 @@
...
+	t.Parallel()
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:                 func() { testAccPreCheck(t) },
+		ProtoV6ProviderFactories: protoV6ProviderFactories(),
+		CheckDestroy:             testCheckPipelineTeamResourceRemoved,
+		Steps: []resource.TestStep{
+			{
+				Config:             testAccPipelineTeamConfigBasic(teamName, "READ_ONLY"),
+				ExpectNonEmptyPlan: true,
```
jradtilbrook, line 25: Why does this test expect a non-empty plan? I would expect this test case to create the 3 resources and then not have any other changes to make after that

buildkite/resource_pipeline_team_test.go
```diff
@@ -0,0 +1,226 @@
...
+	t.Parallel()
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:                 func() { testAccPreCheck(t) },
+		ProtoV6ProviderFactories: protoV6ProviderFactories(),
+		CheckDestroy:             testCheckPipelineTeamResourceRemoved,
+		Steps: []resource.TestStep{
+			{
+				Config:             testAccPipelineTeamConfigBasic(teamName, "BUILD_AND_READ"),
+				ExpectNonEmptyPlan: true,
```
jradtilbrook, line 53: Same with this. I'd expect an empty plan here

buildkite/resource_pipeline_team_test.go
```diff
@@ -0,0 +1,226 @@
...
+			{
+				// re-import the resource and confirm they match
+				ResourceName:      "buildkite_pipeline_team.pipelineteam",
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+		},
+	})
+}
+
```
jradtilbrook, line 140: I think it would be good to have a test where terraform creates the pipeline_team but then its removed on the API side. To ensure that it removes it from state and plans to recreate it

buildkite/resource_pipeline_test.go
```diff
@@ -581,9 +581,11 @@ func TestAccPipeline_remove_team(t *testing.T) {
 					// Quick check to confirm the local state is correct before we update it
 					resource.TestCheckResourceAttr("buildkite_pipeline.foobar", "name", "Test Pipeline foo"),
 				),
+				ImportStateVerifyIgnore: []string{"slug"},
```
jradtilbrook, line 584: Is this what you were talking about at the standup?

What was the error showing up? I'm wondering if there is something else going wrong here and adding this is masking it

jradtilbrook (changes requested):
I think most tests have `ExpectNonEmptyPlan: true` but shouldn't have that. Terraform should create the resources and then have no plan afterwards

buildkite/resource_pipeline_team.go
mcncl: Consider using a `const` as Jarryd has mentioned above. If the "value" of the string ever changes, it won't cause an error.;

```go
stringvalidator.OneOf(
	string(PipelineAccessLevelsReadOnly),
	string(PipelineAccessLevelsBuildAndRead),
	string(PipelineAccessLevelsManageBuildAndRead),
),
```

lizrabuya: I was confused for a sec as I have already done these changes. It got replaced in this commit [Fix: Missing } on setPipelineModel and removed unused function blocks](https://github.com/buildkite/terraform-provider-buildkite/pull/351/commits/4d9cd24cff75d2785f6796d07ce666600f2fc52e)

buildkite/resource_pipeline_team.go
lizrabuya: Same as the issue in line 74 comment, the unused function got reverted in this commit [Fix: Missing } on setPipelineModel and removed unused function blocks](https://github.com/buildkite/terraform-provider-buildkite/pull/351/commits/4d9cd24cff75d2785f6796d07ce666600f2fc52e). I will make sure it got reverted to that.

buildkite/resource_pipeline_team_test.go
lizrabuya: After running the tests, a terraform refresh is run, and this generates a non empty plan because the pipeline resource gets updated with the team block

buildkite/resource_pipeline_team_test.go
lizrabuya: same as above reason for adding the ExpectNonEmptyPlan

buildkite/resource_pipeline_test.go
lizrabuya: Yes, this is what we discussed during standup yesterday. Interestingly, the test no longer fails locally if I remove `ImportStateVerifyIgnore`

buildkite/resource_pipeline_team_test.go
jradtilbrook: Yeah thats right, but thats not what we want. We want the pipeline.team block to ignore the appearance of new teams

buildkite/resource_pipeline.go
```diff
@@ -183,6 +183,7 @@ func (p *pipelineResource) Create(ctx context.Context, req resource.CreateReques
 	if len(plan.Teams) > 0 {
```
jradtilbrook, line 183: Don't forget as part of this PR to add the deprecation message to the teams block on pipeline

jradtilbrook (approved):

Issue #214, opened 2022-08-30
Add retries to mitigate frequent "504 Gateway Timeout" errors


albertywu:
We frequently get 504 Gateway Timeout errors when using this Terraform provider.

Example: https://buildkite.com/uber/terraform/builds/3472#0182ed92-3c4f-4cc2-b069-8248c94d9adb

```
╷
--
  | │ Error: non-200 OK status code: 504 Gateway Timeout body: "<html>\r\n<head><title>504 Gateway Time-out</title></head>\r\n<body>\r\n<center><h1>504 Gateway Time-out</h1></center>\r\n</body>\r\n</html>\r\n"
  | │
  | │   with buildkite_pipeline.uber-keyring-debmirror-deb-production,
  | │   on debian-builds.tf line 8472, in resource "buildkite_pipeline" "uber-keyring-debmirror-deb-production":
  | │ 8472: resource "buildkite_pipeline" "uber-keyring-debmirror-deb-production" {
```

Add retries with exponential backoff such that this issue is mitigated

jradtilbrook:
Thanks @albertywu. It's hard to tell from that error which request is specifically timing out (because it involves graphql and rest requests) but we are aware of general timeouts in the provider and have created issue #201 to add in retries.

We don't have a timeline just yet though but are aiming for usability fixes like this to land this year ahead of releasing a stable v1

Issue #201, opened 2022-05-24
Introduce API request retries


jradtilbrook:
Sometimes the API errors out and causes a terraform plan or apply to fail. This is sometimes fixable by retrying so we should introduce such functionality to make the use of the provider better.

PR #356, opened 2023-08-17
Add retries to cluster resource create

mcncl:
## PR checklist:
- [x] `docs/` updated
- [x] tests added

feat: Add timeouts and retries to the Cluster resource.


mcncl:
Step 1 of addressing/closing https://github.com/buildkite/terraform-provider-buildkite/issues/201

mcncl:
The furstrating thing is that it needs to use the `Block`s, but this could be adjusted before merge if we were planning a `v1.0.0-beta` release.

buildkite/resource_cluster.go
```diff
@@ -75,25 +78,50 @@ func (c *clusterResource) Schema(ctx context.Context, req resource.SchemaRequest
...
+			state.Description.ValueStringPointer(),
+			state.Emoji.ValueStringPointer(),
+			state.Color.ValueStringPointer(),
+		)
+		if err != nil {
+			return retry.NonRetryableError(err)
+		}
+
+		return nil
+	})
```
jradtilbrook, line 124: **Summary of my code diving:**  
I think this all makes sense. `createTimeout` is the maximum time the client will retry, which includes N retries within that time until it can get a value. The SDK has baked in to wait 500ms between tries up to 10s max. So it'll run through a few retries by the time it hits `DefaultTimeout`.

This should capture non-2xx response codes from the API. I'm thinking 401/3 which we don't want to retry, but we do want to allow retrying on 429 and possibly some/most 5xx codes. This is not captured currently so won't really resolve #214.

Graphql is also pretty tricky in that it returns 200 status codes but an `error` field in the response JSON. This is not captured by this either.

**Some things to figure out:**  
- Do we want to have a maximum timeout users can't go above? I think maybe not, at least at first. The timeout block is the absolute timeout which includes retrying so its not the timeout of the HTTP requests themselves
- Per above, we could extend this to cancel the HTTP requests themselves by making use of https://github.com/Khan/genqlient/blob/main/docs/genqlient.yaml#L55-L61 which would be separate from this timeout. I think the `timeout` block per resource makes sense to be a maximum timeout for that resource to be operated on by terraform. A HTTP level timeout would be less than that
- How do we figure out the status code or graphql error that occurred? So that we can make use of the `retry.RetryableError`

buildkite/resource_cluster.go
```diff
@@ -75,25 +78,50 @@ func (c *clusterResource) Schema(ctx context.Context, req resource.SchemaRequest
 				MarkdownDescription: "A color representation of the Cluster. Accepts hex codes, eg #BADA55.",
 			},
 		},
+		Blocks: map[string]resource_schema.Block{
+			"timeouts": timeouts.Block(ctx, timeouts.Opts{
```
jradtilbrook, line 82: I think we probably want timeouts for all operations right? So using `timeouts.BlockAll()`

buildkite/data_source_cluster.go
```diff
@@ -106,5 +107,10 @@ func (*clusterDatasource) Schema(ctx context.Context, req datasource.SchemaReque
 				MarkdownDescription: "The color of the cluster.",
 			},
 		},
+		Blocks: map[string]schema.Block{
+			"timeouts": timeouts.Block(ctx, timeouts.Opts{
+				Create: true,
```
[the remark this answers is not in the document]
jradtilbrook: And make use of the value in the Read function

jradtilbrook:
We could even add a timeouts thing to the provider configuration to act as defaults for all resources 👌

buildkite/resource_cluster.go
jradtilbrook: Turns out genqlient does it for us already: https://github.com/Khan/genqlient/blob/main/graphql/client.go#L161-L177

So if there is a HTTP level error, we get that returned. 
If there is a json decode error (ie wrong object) we get that returned.
If there are errors in the json response, we get that returned.

The only thing that is missing is access to response headers for things like rate-limiting. But through the `retry.Retry` we have no place to change the wait between retries anyway to make use of the specified time

jradtilbrook (changes requested):
Looks good. Just have to remove the extra definition of timeout blocks on the resources themselves since they are on the provider now

mcncl:
Good call!

buildkite/data_source_cluster.go
```diff
@@ -124,5 +125,10 @@ func (*clusterDatasource) Schema(ctx context.Context, req datasource.SchemaReque
 				MarkdownDescription: "The color of the cluster.",
 			},
 		},
+		Blocks: map[string]schema.Block{
```
jradtilbrook, line 128: @mcncl and this one

buildkite/resource_cluster.go
```diff
@@ -4,25 +4,28 @@ import (
...
-	Emoji       types.String `tfsdk:"emoji"`
-	Color       types.String `tfsdk:"color"`
-	UUID        types.String `tfsdk:"uuid"`
+	ID          types.String   `tfsdk:"id"`
+	Name        types.String   `tfsdk:"name"`
+	Description types.String   `tfsdk:"description"`
+	Emoji       types.String   `tfsdk:"emoji"`
+	Color       types.String   `tfsdk:"color"`
+	UUID        types.String   `tfsdk:"uuid"`
+	Timeouts    timeouts.Value `tfsdk:"timeouts"`
```
james2791: Here too 👍 

```suggestion

```

docs/resources/cluster.md
```diff
@@ -20,14 +23,15 @@ resource "buildkite_cluster" "linux" {
 * `description` - (Optional) This is a description for the cluster, this may describe the usage for it, the region, or something else which would help identify the Cluster's purpose.
 * `emoji` - (Optional) An emoji to use with the Cluster, this can either be set using `:buildkite:` notation, or with the emoji itself, such as 😎.
 * `color` - (Optional) A color to associate with the Cluster. Perhaps a team related color, or one related to an environment. This is set using hex value, such as `#BADA55`.
+* `timeouts` - (Optional) A `block` (see above) for timeout values. The default is 60s (1m).
```
jradtilbrook, line 26: Damn and the docs here

jradtilbrook (changes requested):

james2791 (approved):
💪

jradtilbrook (approved):
Haha I love the commit fury at the end here.
I think thats everything

PR #358, opened 2023-08-17
Migrate to Protocol v6 - ONLY MERGE WHEN v1 READY

mcncl:
## PR checklist:
- [ ] `docs/` updated
- [ ] tests added

feat: Migrate to protocol v6
BREAKING CHANGE: This will mean that versions of Terraform < 0.15.4 are no longer supported


jradtilbrook (commented):
I think by the time we are ready for this, we wont need to mux in the old provider. It only has `buildkite_organization_settings` which is scheduled for removal so we'll just remove the entire mux provider

jradtilbrook:
Superseded by #400

PR #365, opened 2023-08-21
Move Cluster tests to t.Run

mcncl:
refactor(cluster resource tests): Migrate the Cluster resource tests to use the t.run approach

## PR checklist:
- [ ] `docs/` updated
- [x] tests added


buildkite/resource_cluster_test.go
```diff
@@ -9,105 +9,130 @@ import (
...
+							resource.TestCheckResourceAttr("buildkite_cluster.foo", "name", fmt.Sprintf("%s_test_cluster", randNameUpdated)),
+							resource.TestCheckResourceAttrSet("buildkite_cluster.foo", "id"),
+							resource.TestCheckResourceAttrSet("buildkite_cluster.foo", "uuid"),
+						),
+					},
+				},
+			})
+		}
+		t.Run("with basic settings", func(t *testing.T) {
+			testCase(t, basic(randName), basic(randNameUpdated))
```
jradtilbrook, line 98: There isn't much point nesting this at this level with the `testCase` function. I think it should just merged above and run under the "updates a cluster" run

buildkite/resource_cluster_test.go
```diff
@@ -9,105 +9,130 @@ import (
...
-					return c.UUID.ValueString(), nil
+	t.Run("Imports a Cluster", func(t *testing.T) {
+		randName := acctest.RandString(5)
+		check := resource.ComposeAggregateTestCheckFunc(
+			testAccCheckClusterExists("buildkite_cluster.foo", &c),
+			resource.TestCheckResourceAttr("buildkite_cluster.foo", "name", fmt.Sprintf("%s_test_cluster", randName)),
+		)
+
+		testCase := func(t *testing.T, config string) {
+			resource.Test(t, resource.TestCase{
```
jradtilbrook, line 110:
```suggestion
			resource.ParallelTest(t, resource.TestCase{
```

buildkite/resource_cluster_test.go
```diff
@@ -9,105 +9,130 @@ import (
...
+				CheckDestroy:             testAccCheckClusterDestroy,
+				Steps: []resource.TestStep{
+					{
+						Config: config,
+						Check:  check,
+					},
+					{
+						ResourceName: "buildkite_cluster.foo",
+						ImportStateIdFunc: func(s *terraform.State) (string, error) {
+							return c.UUID.ValueString(), nil
```
jradtilbrook, line 122: Oh I didn't realise cluster is using UUID for importing. We'll need to change this for v1 to use ID like everything else

buildkite/resource_cluster_test.go
```diff
@@ -9,105 +9,130 @@ import (
...
+							resource.TestCheckResourceAttr("buildkite_cluster.foo", "name", fmt.Sprintf("%s_test_cluster", randName)),
+							resource.TestCheckResourceAttrSet("buildkite_cluster.foo", "id"),
+							resource.TestCheckResourceAttrSet("buildkite_cluster.foo", "uuid"),
+						),
+					},
+					{
+						Config: newConfig,
+						Check: resource.ComposeAggregateTestCheckFunc(
+							testAccCheckClusterExists("buildkite_cluster.foo", &c),
+							testAccCheckClusterRemoteValues(&c, fmt.Sprintf("%s_test_cluster", randNameUpdated)),
```
jradtilbrook, line 88: This only checks the name but I think it should also check the other properties to make sure its all working correctly

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file.
 
 - Move archive pipeline config to provider [[PR #354](https://github.com/buildkite/terraform-provider-buildkite/pull/354)] @jradtilbrook
 - SUP-1076 Convert testing to framework [[PR #361](https://github.com/buildkite/terraform-provider-buildkite/pull/361)] @jradtilbrook
+    - SUP-1076 Move Cluster tests to use t.Run [[PR #365](https://github.com/buildkite/terraform-provider-buildkite/pull/365) @mcncl
```
jradtilbrook, line 9:
```suggestion
- SUP-1076 Move Cluster tests to use t.Run [[PR #365](https://github.com/buildkite/terraform-provider-buildkite/pull/365)] @mcncl
```

jradtilbrook (changes requested):

buildkite/resource_cluster_test.go
mcncl: @jradtilbrook there may be a time in which we want to add additional runs to this, such as using the `complex` config.

buildkite/resource_cluster_test.go
mcncl: Agreed, but that shouldn't be done in this PR.

CHANGELOG.md
mcncl: The above makes it a nested list, showing that it's a part of the same Linear work.

buildkite/resource_cluster_test.go
mcncl: This errors if you use `Parallel` due to `foo` being used by both runs.

buildkite/resource_cluster_test.go
```diff
@@ -9,105 +9,130 @@ import (
...
+		check := resource.ComposeAggregateTestCheckFunc(
+			testAccCheckClusterExists("buildkite_cluster.foo", &c),
+			testAccCheckClusterRemoteValues(&c, fmt.Sprintf("%s_test_cluster", randName)),
+			resource.TestCheckResourceAttr("buildkite_cluster.foo", "name", fmt.Sprintf("%s_test_cluster", randName)),
+			resource.TestCheckResourceAttrSet("buildkite_cluster.foo", "id"),
+			resource.TestCheckResourceAttrSet("buildkite_cluster.foo", "uuid"),
+		)
+
+		testCase := func(t *testing.T, config string) {
+			resource.Test(t, resource.TestCase{
```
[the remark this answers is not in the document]
mcncl: This errors if you use `Parallel` due to `foo` being used by both runs.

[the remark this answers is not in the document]
mcncl: This is why the tests now fail on `Import`.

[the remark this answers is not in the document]
mcncl: Latest run when only using one `t.Run` passes. Using `Parallel` with multiple `t.Run` which share the resource model will fail.

buildkite/resource_cluster_test.go
mcncl: With the `basic` config, there is only the `name` set. It also checks that the ID and UUID are set.

mcncl: Added additional to ensure unchanged remain `null`

buildkite/resource_cluster_test.go
jradtilbrook: No thats right. I added SUP-1380

CHANGELOG.md
jradtilbrook: Oh I see 👍 just a missing bracket then

jradtilbrook (changes requested):
I think the test structure needs to change so that we can run these tests in parallel. I'd say the double nesting of t.Run isn't helping in that case and a single level with a bit of duplication is an easier approach

mcncl:
@jradtilbrook I'll change it. I do feel as though that's a preference thing, but if it's going to block the PR then it's not a hill I'm going to die on.

jradtilbrook:
> @jradtilbrook I'll change it. I do feel as though that's a preference thing, but if it's going to block the PR then it's not a hill I'm going to die on.

Yeah because the tests are already slow, we should try to speed them up and parallelise them if we can

jradtilbrook (approved):

PR #368, opened 2023-08-22
SUP-1337: Remove org settings resource & remove mux provider

lizrabuya:
## PR checklist:
- [x] `docs/` updated
- [x] tests added


jradtilbrook (changes requested):

buildkite/provider.go
```diff
@@ -65,13 +64,12 @@ func (tf *terraformProvider) Configure(ctx context.Context, req provider.Configu
...
 	}
 
-	legacyProvider := schema.Provider{}
 	config := clientConfig{
 		apiToken:   apiToken,
 		graphqlURL: graphqlUrl,
 		org:        organization,
 		restURL:    restUrl,
-		userAgent:  legacyProvider.UserAgent("buildkite", tf.version),
+		userAgent:  fmt.Sprintf("buildkite/%s", tf.version),
```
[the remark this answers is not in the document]
lizrabuya: The reason I had to set this was due to the panic error I encountered similar to this [issue](https://github.com/hashicorp/terraform-plugin-testing/issues/93 ) . I still have to find an alternative way to keep the functionality the same with sdkv2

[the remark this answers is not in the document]
jradtilbrook: Really? But how is that getting triggered here now? I would have expected it to cause a panic in an earlier PR when we switched to framework testing package.

We can just copy out that function and put it in here if you want then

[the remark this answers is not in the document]
jradtilbrook: Could add this patch in https://gist.github.com/jradtilbrook/234bc41f3501ad496695099aab80c483

buildkite/provider.go
```diff
@@ -83,6 +83,26 @@ func (tf *terraformProvider) Configure(ctx context.Context, req provider.Configu
...
+func userAgent(name, providerVersion, tfVersion string) string {
+	ua := fmt.Sprintf("Terraform/%s (+https://www.terraform.io)", tfVersion)
+	if name != "" {
+		ua += " " + name
+		if providerVersion != "" {
+			ua += "/" + providerVersion
+		}
+	}
+
+	if add := os.Getenv("TF_APPEND_USER_AGENT"); add != "" {
```
mcncl, line 95: I'm not sure it's clear what `add` is here? Could it be given a clearer name? I'm thinking if someone else were to pick this up, could they immediately know what would happen if they changed/removed `add`?

mcncl (commented):
This seems to do significantly more than just remove the `org settings` resource; it seems to also remove the config for the legacy provider. Is that an intentional inclusion in this PR?

buildkite/provider.go
```diff
@@ -83,6 +83,26 @@ func (tf *terraformProvider) Configure(ctx context.Context, req provider.Configu
 	resp.DataSourceData = client
 }
 
+func userAgent(name, providerVersion, tfVersion string) string {
+	ua := fmt.Sprintf("Terraform/%s (+https://www.terraform.io)", tfVersion)
```
jradtilbrook, line 87: @yob will this cause issues for you monitoring if we change the format of the user agent? With this change it will look like: `Terraform/%s (+https://www.terraform.io) buildkite/v0.25.1` where I think it used to look like `Terraform/%s (+https://www.terraform.io) Terraform-Plugin-SDK/v2.10.0 buildkite/v0.25.1`

jradtilbrook (approved):
Just got a conflict on changelog

buildkite/provider.go
jradtilbrook: Good point this is copied from the SDKv2 UserAgent function verbatim so that we can remove the sdkv2 dependency 🙌

buildkite/provider.go
yob: totally fine, we don't have any processing of the user agent that depends on the precise format

james2791 (approved):
👍 - Just changelog conflict as above

jradtilbrook (dismissed):

buildkite/provider.go
lizrabuya: @mcncl , I've modified the function with more descriptive var names

mcncl (approved):

lizrabuya:
> This seems to do significantly more than just remove the `org settings` resource; it seems to also remove the config for the legacy provider. Is that an intentional inclusion in this PR?

Yes, that was part in the linear ticket [SUP-1337](https://linear.app/buildkite/issue/SUP-1337/remove-buildkite-organization-settings-resource)

PR #369, opened 2023-08-22
SUP-1319: Removal of archive_on_delete from pipeline resource

james2791:
refactor(pipeline resource/pipeline resource tests): Removal of archive_on_delete from pipeline schema/resource after moved to provider settings (`archive_pipeline_on_delete`)

## PR checklist:
- [x] `docs/` updated
- [x] tests added


buildkite/resource_pipeline.go
```diff
@@ -591,11 +615,297 @@ func (*pipelineResource) Schema(ctx context.Context, req resource.SchemaRequest,
 				},
 			},
 		},
+		Version: 1,
+	}
+}
+
+func (p *pipelineResource) UpgradeState(ctx context.Context) map[int64]resource.StateUpgrader {
```
jradtilbrook, line 622: @james2791 heads up: lets not bother with this approach for removing `archive_on_delete`. Its been deprecated and moved to provider config and we are in v0.X territory.

So its fine for this to just go away completely and not need mapping over. Should simplify the logic here significantly

james2791: Fair - it does significantly 😉

jradtilbrook (changes requested):
Looks good. Just needs the docs updated

jradtilbrook (approved):

PR #372, opened 2023-08-23
Use ID for Cluster read and import

mcncl:
refactor(cluster resource)[!]: [Unify import method to use ID]

## PR checklist:
- [x] `docs/` updated
- [x] tests added


buildkite/resource_cluster.go
```diff
@@ -128,8 +128,18 @@ func (c *clusterResource) Read(ctx context.Context, req resource.ReadRequest, re
...
+		if clusterNode == nil {
+			resp.Diagnostics.AddError(
+				"Unable to get Cluster",
+				"Error getting Cluster: nil response",
+			)
+			return
+		}
+		updateClusterResourceState(&state, *clusterNode)
+		resp.Diagnostics.Append(resp.State.Set(ctx, &state)...)
+	}
```
jradtilbrook, line 141: We need an else here for if the cluster is not found. It needs to be removed from state. You can see similar in https://github.com/buildkite/terraform-provider-buildkite/blob/main/buildkite/resource_pipeline.go#L327-L331

buildkite/resource_cluster_test.go
```diff
@@ -103,7 +103,7 @@ func TestAccCluster_Import(t *testing.T) {
 			{
 				ResourceName: "buildkite_cluster.foo",
 				ImportStateIdFunc: func(s *terraform.State) (string, error) {
-					return c.UUID.ValueString(), nil
+					return c.ID.ValueString(), nil
 				},
```
jradtilbrook, line 107: Since its using the actual ID you can completely leave this out

jradtilbrook (changes requested):

buildkite/resource_cluster_test.go
mcncl: Weird. This is removed for me locally. Will make sure it's pushed.

jradtilbrook (approved):

jradtilbrook (approved):

PR #380, opened 2023-08-29
Add retries to pipeline-schedule

lizrabuya:
## PR checklist:
- [x] `docs/` updated
- [x] tests added


buildkite/resource_pipeline_schedule.go
```diff
@@ -38,7 +39,7 @@ func NewPipelineScheduleResource() resource.Resource {
 	return &pipelineSchedule{}
 }
 
-func (ps *pipelineSchedule) Metadata(_ context.Context, req resource.MetadataRequest, resp *resource.MetadataResponse) {
```
james2791, line 41: Nice pickup :)

buildkite/resource_pipeline_schedule.go
```diff
@@ -117,17 +118,37 @@ func (ps *pipelineSchedule) Create(ctx context.Context, req resource.CreateReque
...
+			plan.Commit.ValueStringPointer(),
+			plan.Branch.ValueStringPointer(),
+			&envVars,
+			plan.Enabled.ValueBool())
+
+		if err != nil {
+			if isRetryableError(err) {
+				return retry.RetryableError(err)
+			}
+			return retry.NonRetryableError(err)
```
james2791, line 148: I think this could be changed - and you'd not need the `err` on the `RetryContext`

```suggestion
			resp.Diagnostics.AddError(
			    "Unable to create Pipeline schedule",
			    fmt.Sprintf("Unable to create Pipeline schedule: %s", err.Error()),
		    )
		    return retry.NonRetryableError(err)
```

buildkite/resource_pipeline_schedule.go
```diff
@@ -117,17 +118,37 @@ func (ps *pipelineSchedule) Create(ctx context.Context, req resource.CreateReque
...
+		}
+		return nil
+	})
 
 	if err != nil {
```
james2791, line 153: As above - this could be looped into the retry 👍

buildkite/resource_pipeline_schedule.go
mcncl: This can be an unnamed; it isn't being referenced in the code so `_` is an acceptable way to use it, it's arguably the correct way to use it, too.

buildkite/resource_pipeline_schedule.go
```diff
@@ -155,17 +176,37 @@ func (ps *pipelineSchedule) Create(ctx context.Context, req resource.CreateReque
...
```
james2791, line 211: As above with the Create - this err check could be looped into here 👍

buildkite/resource_pipeline_schedule.go
```diff
@@ -191,37 +232,57 @@ func (ps *pipelineSchedule) Read(ctx context.Context, req resource.ReadRequest,
...
+			if isRetryableError(err) {
+				return retry.RetryableError(err)
+			}
+			return retry.NonRetryableError(err)
+		}
+
+		return nil
+	})
 
 	if err != nil {
```
james2791, line 281: As above with the Create - this err check could be looped into here 👍

buildkite/resource_pipeline_schedule.go
```diff
@@ -230,14 +291,33 @@ func (ps *pipelineSchedule) Update(ctx context.Context, req resource.UpdateReque
...
+		if err != nil {
+			if isRetryableError(err) {
+				return retry.RetryableError(err)
+			}
+			return retry.NonRetryableError(err)
+		}
+		return nil
+	})
+
 	if err != nil {
```
james2791, line 321: As above with the Create - this err check could be looped into here 👍

docs/resources/pipeline_schedule.md
```diff
@@ -11,8 +11,7 @@ resource "buildkite_pipeline_schedule" "repo2_nightly" {
   pipeline_id = buildkite_pipeline.repo2.id
   label       = "Nightly build"
   cronline    = "@midnight"
-  branch      = buildkite_pipeline.repo2.default_branch
-}
```
james2791, line 15: Accidental deletion of the closing `}`?

james2791 (changes requested):

buildkite/resource_pipeline_schedule.go
lizrabuya: @mcncl , @james2791 , it can be unnamed, but was looking at more of a "readability" point of view in the code..

buildkite/resource_pipeline_schedule.go
lizrabuya: I am using `err` outside the code block of RetryContext.  The Create operation exits when an error is encountered on retry with this code block:

```	
       if err != nil {
		resp.Diagnostics.AddError(
			"Unable to create Pipeline schedule",
			fmt.Sprintf("Unable to create Pipeline schedule: %s", err.Error()),
		)
		return
	}
```

buildkite/resource_pipeline_schedule.go
```diff
@@ -155,17 +176,37 @@ func (ps *pipelineSchedule) Create(ctx context.Context, req resource.CreateReque
...
+		)
+
+		if err != nil {
+			if isRetryableError(err) {
+				return retry.RetryableError(err)
+			}
+			return retry.NonRetryableError(err)
+		}
+		return nil
+	})
```
jradtilbrook, line 209: Need to handle the case where there is no schedule found anymore and remove it from state. not sure how that looks like in this retry way of doing things

buildkite/resource_pipeline_schedule.go
```diff
@@ -191,37 +232,57 @@ func (ps *pipelineSchedule) Read(ctx context.Context, req resource.ReadRequest,
...
+		input := PipelineScheduleUpdateInput{
+			Id:       state.Id.ValueString(),
+			Label:    plan.Label.ValueStringPointer(),
+			Cronline: plan.Cronline.ValueStringPointer(),
+			Branch:   plan.Branch.ValueStringPointer(),
+			Commit:   plan.Commit.ValueStringPointer(),
+			Message:  plan.Message.ValueStringPointer(),
+			Env:      &envVars,
+			Enabled:  plan.Enabled.ValueBool(),
+		}
```
jradtilbrook, line 264: Not a big deal, but this would be called over and over again and create a new struct each time it retries. Leaving it outside this function would be more performant. But we are talking nano seconds really so not worth worrying about - just letting you know

junit-.xml
```diff
@@ -0,0 +1,88 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jradtilbrook, line 1: Maybe add this file to .gitignore

docs/resources/pipeline_schedule.md
```diff
@@ -24,7 +24,7 @@ resource "buildkite_pipeline_schedule" "repo2_nightly" {
 * `commit` - (Optional, Default: `HEAD`) The commit ref to use for the build.
 * `message` - (Optional, Default: `Scheduled build`) The message to use for the build.
 * `env` - (Optional) A map of environment variables to use for the build.
-* `enabled` - (Optional, Default: `true`) Whether the schedule should run.
+* `enabled` - (Optional, Default: `true`) Whether the schedule should run.
```
jradtilbrook, line 27: We have a .editorconfig file in this repo that should stop these empty spaces at the end of lines. You should install the plugin for your IDE so we don't get these diffs

jradtilbrook (changes requested):

buildkite/resource_pipeline_schedule.go
lizrabuya: I have added the logic to remove from state if schedule is not found. I think there  is no need to manage a  retry for that use-case.

buildkite/resource_pipeline_schedule.go
lizrabuya: You're right. I have moved this out of the retry loop

jradtilbrook (dismissed):

james2791 (approved):
LGTM - just changelog conflicts now!

james2791 (approved):

PR #383, opened 2023-08-29
SUP-1400: Test suite and test suite team (team suite) resource retries 

james2791:
feat (Timeouts/Retries): Add timeouts (provider config set) and CRUD retries to the Test suite and test suite team resources.

## PR checklist:
- [x] `docs/` updated (provider docs updated previously)
- [x] tests added (TST tests migrated to t.Run())
- [x] an example added to `examples/` (provider example updated with `timeouts` block)

Also converted the Test suite team tests with `t.Run` to enable parallelism / set timeouts in test configuration. Both TS and TST have randomness enabled previously.

This PR also adjusts the Client's `makeRequest` function to returns errors with a `returned error` format line so that REST API requests can benefit from the rate limit/server error retry methods implemented.

### DIFF
--- a/buildkite/resource_test_suite.go
+++ b/buildkite/resource_test_suite.go
@@ -62,6 +62,7 @@ func (ts *testSuiteResource) Create(ctx context.Context, req resource.CreateRequ
 		return
 	}
 
+	// Use the Read timeout for obtaining a Test suite's UUID
 	timeout, diags := ts.client.timeouts.Read(ctx, DefaultTimeout)
 	resp.Diagnostics.Append(diags...)
 
@@ -107,6 +108,11 @@ func (ts *testSuiteResource) Create(ctx context.Context, req resource.CreateRequ
 
 	// Construct URL to call to the REST API
 	url := fmt.Sprintf("/v2/analytics/organizations/%s/suites", ts.client.organization)
+
+	// Use the Create timeout for test suite creation
+	timeout, diags = ts.client.timeouts.Create(ctx, DefaultTimeout)
+	resp.Diagnostics.Append(diags...)
+
 	createErr := retry.RetryContext(ctx, timeout, func() *retry.RetryError {
 		err = ts.client.makeRequest(ctx, "POST", url, payload, &response)
 
--- a/buildkite/resource_test_suite.go
+++ b/buildkite/resource_test_suite.go
@@ -242,6 +242,11 @@ func (ts *testSuiteResource) Read(ctx context.Context, req resource.ReadRequest,
 			// and we didnt find another team with MANAGE_AND_READ
 			state.TeamOwnerId = types.StringUnknown()
 		}
+	} else {
+		// Test suite was removed - remove from state
+		resp.Diagnostics.AddWarning("Test suite not found", "Removing test suite from state")
+		resp.State.RemoveResource(ctx)
+		return
 	}
 
 	if state.TeamOwnerId.IsUnknown() {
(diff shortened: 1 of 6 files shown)


buildkite/resource_test_suite.go
```diff
@@ -54,19 +55,45 @@ func (ts *testSuiteResource) Create(ctx context.Context, req resource.CreateRequ
...
+
 	if resp.Diagnostics.HasError() {
 		return
 	}
 
+	timeout, diags := ts.client.timeouts.Create(ctx, DefaultTimeout)
+	resp.Diagnostics.Append(diags...)
+
 	// The REST API requires team UUIDs but everything else in the provider uses GraphQL IDs. So we map from UUID to ID
 	// here
```
mcncl, line 69: Perhaps this should be moved to where it is relevant?

buildkite/resource_test_suite.go
```diff
@@ -78,10 +105,26 @@ func (ts *testSuiteResource) Create(ctx context.Context, req resource.CreateRequ
...
+
+		if err != nil {
+			if isRetryableError(err) {
+				return retry.RetryableError(err)
+			}
+			return retry.NonRetryableError(err)
+		}
+
+		return nil
+	})
```
jradtilbrook, line 121: I wonder if the whole thing should be a single retry? Maybe not, I don't really know but calling it out if anyone has an argument either way

buildkite/resource_test_suite.go
```diff
@@ -119,20 +183,48 @@ func (*testSuiteResource) Metadata(ctx context.Context, req resource.MetadataReq
...
+
 	if err != nil {
-		resp.Diagnostics.AddError("Failed to load test suite from GraphQL", err.Error())
+		resp.Diagnostics.AddError(
+			"Failed to load test suite from GraphQL",
+			err.Error(),
+		)
 		return
```
jradtilbrook, line 222: I don't think this has the remove from state logic like we've been adding everywhere. Can you add it?

jradtilbrook (changes requested):
Looks good. Just some minor things

buildkite/resource_test_suite.go
james2791: Fair point - I _think_ it makes sense leaving them seperate, as the first technically is a read, the second is the actual creation. Latest commit has the first using the `Read` timeout, and the second (REST POST) using the `Create` timeout via the provider

james2791: Fair point - I think it makes sense leaving them seperate, as the first technically is a read, the second is the actual creation. Latest commit has the first using the Read timeout, and the second (REST POST) using the Create timeout via the provider

buildkite/resource_test_suite.go
```diff
@@ -54,19 +55,45 @@ func (ts *testSuiteResource) Create(ctx context.Context, req resource.CreateRequ
...
 
-	resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)
+	diags := req.Plan.Get(ctx, &plan)
+	resp.Diagnostics.Append(diags...)
+
 	if resp.Diagnostics.HasError() {
 		return
 	}
 
+	timeout, diags := ts.client.timeouts.Read(ctx, DefaultTimeout)
```
jradtilbrook, line 65: Hmm I think this should still use the create timeout value. Or at least the read part of this create function use the read timeout and the create part of it use the create part

jradtilbrook (changes requested):

buildkite/resource_test_suite.go
james2791: Just whipped that in wit the latest commit - `POST` to test suite creation endpoint using the `Create` timeout - above with Team owner UUID -> ID using the `Read` T/O

jradtilbrook (approved):

jradtilbrook (dismissed):

jradtilbrook (approved):

PR #386, opened 2023-09-01
SUP-1393 Detect repository provider type

jradtilbrook:
## PR checklist:
- [x] `docs/` updated
- [x] tests added


james2791 (dismissed):

buildkite/resource_pipeline.go
```diff
@@ -456,6 +472,9 @@ func (*pipelineResource) Schema(ctx context.Context, req resource.SchemaRequest,
 			"repository": schema.StringAttribute{
 				Required: true,
 			},
+			"repository_provider": schema.StringAttribute{
+				Computed: true,
+			},
```
jradtilbrook, line 477: The main reason for putting this in the schema is to allow us to run validation against it. However, because its computed, its actually not known at validation time so isn't useful. We will have to compute it within the validation functions

james2791 (approved):

PR #390, opened 2023-09-04
SUP-1396 Add default queue flag to cluster

### LINKED ISSUE #377: Add support for setting cluster default queue
I created a cluster and queue using Terraform, but found it did not support wait or block steps. It turns out no default queue is set when creating clusters and queues via Terraform. See https://forum.buildkite.community/t/buildkite-clusters-and-wait-block-step-unsupported/3210 for more info.

We need a way in Terraform to be able to set the default queue of a cluster. Ideally the buildkite_cluster resource would have a way to assign a default queue (e.g. set the default_queue_id field in the REST API).

I'd write a PR myself, but this is made tricky since a buildkite_cluster_queue needs a reference to the buildkite_cluster.id field, while a default queue field would need the buildkite_cluster_queue.id field, making for a circular dependency. You might be able to get away with a data resource, but no buildkite_cluster_queue resource exists currently.

Alternatively, a buildkite_cluster_default_queue resource could be added just for this.

mcncl:
feat(cluster default queues): Add support for defining a default Cluster queue in Update

## PR checklist:
- [x] `docs/` updated
- [x] tests added

This is designed in a way that when API support is added in there will be no breaking changes in the provider; the value is set as a pointer, so should the user opt out of defining this in an update it will remain as `null`, however it also means it won't overwrite the value once it is available in the API as default action.

This includes a custom check to ensure the value of `default_queue_id` is actually `null` on create, checking that it isn't set is not sufficient.

Resolves #377


jradtilbrook (changes requested):
This looks good but it leaves it open for a chicken-egg problem since you need to create a cluster to create a queue. I think we should add 2 attributes: `add_default_queue` and `default_queue_id` so we can allow creating a cluster and then immediately creating a queue and adding it as default? Do you think that will work okay or introduce other issues?

mcncl:
@jradtilbrook this PR represents the current functionality of GraphQL. We maybe could go down the boolean route, but it might be worth looking in to that if the roadmap for implementing that behaviour in the API isn't within a desirable timeframe? We'd have to later remove the `add...` attribute if the functionality within the API was improved as it would be redundant.

There are considerable moving parts if we opt for the latter; create the queue, then ensure it is imported in to state correctly. We'd also have to factor in if the user has `add...` set to `true` but then defines a `default` queue...

mcncl:
Perhaps we update the docs around the `queue` to make that requirement clearer. 🤷‍♂️

mbarrien:
How about my original suggestion of a full resource for `buildkite_cluster_default_queue`? As long as other resources aren't touching the graphql defaultQueue field of a cluster (i.e. not included in any mutations for buildkite_cluster) this would work. This would make the Terraform model depart a little bit from the way the backend is modeled, but would solve the circular dependency.

There's the problem of having multiple instances of `buildkite_cluster_default_queue` and Terraform non-deterministically picking one or the other or flapping between which one is default. In such cases, leave it up to the user to ensure there is only a single instance of the `buildkite_cluster_default_queue` resource in their infrastructure.

```hcl
resource "buildkite_cluster" "cluster" {
  name = "cluster"
}

resource "buildkite_cluster_queue" "default" {
  cluster_id  = buildkite_cluster.cicd.id
  key        = "default"
}

resource "buildkite_cluster_default_queue" "default" {
  cluster_id = buildkite_cluster.cluster.id
  queue_id = buildkite_cluster_queue.default.id
}

# Returns a queue_id field. Don't even expose the defaultQueue field on a buildkite_cluster_queue Terraform resource/data source, force it to be queried through the data field.
data "buildkite_cluster_default_queue" "default" {
  cluster_id = buildkite_cluster.cluster.id
}
```

mcncl:
@mbarrien appreciate the additional input!

I think the underlying issue here, however, is that the expected behaviour in the API (not just TF, but the REST and GQL APIs) doesn't align with the behaviour in the UI.

There could be an argument for either, but the way the UI works at the moment arguably makes the most sense; when a user creates a cluster a default queue is created and thus the user is enabled in actually getting going with clusters immediately. I think it makes sense to replicate _that_, and thus allowing folks to use clusters as soon as one is created in the provider. I don't think there should be additional hoops for folks to jump through in order to use clusters.

jradtilbrook:
@mbarrien I like that idea, we are going to explore that as an option but also see about having something like `create_default_queue` on the cluster resource as well

jradtilbrook:
@mbarrien I really appreciate the thought you are putting into this! It really helps us to land the best approach 👌 

Current thoughts are (still experimenting with implementation though):
- this pr will add a `create_default_queue` boolean. if true, will create a queue and assign it as default to the cluster. from then on it will manage that queue. its not configurable so the key and description is fixed. if the attribute is set to false, the queue is removed. so the queue is a managed queue in this fashion its just that the attributes are not modifable
- add another resource just for `buildkite_cluster_default_queue` that takes the queue id and cluster id and sets it as default

These separate resources are mutually exclusive and one will error if using the other. But I think that gives the least amount of foot-guns and user experience.  
So if a user wants to set a default queue and customise the key and description, they can create the queue separately and then use the separate resource to assign it as default

> What's the behavior if the user manually removes the default queue? How do they get it back? Tell them to do manual steps again, or tell them to remove the default_cluster_queue line+apply+read+apply again? Tell them to now create the buildkite_cluster_queue resource, then set default_queue_id on the buildkite_cluster (but now subject them to the circular dependency issue if they want to reuse that module when doing it from scratch?)

We can handle this with the proposed design because terraform performs a read/refresh prior to an apply. Terraform will know from state that the cluster previously had a default queue, but when reading from the API will see it has been removed or changed. In this circumstance it will generate a diff to set the default back to what it was if it still exists, or create a new one if it doesn't and set that as the default. If that is not what the user wants, the boolean will need to be removed/set to false to disable managing the default queue

Thats all to say how it will work, this PR needs some changes to make that happen

jradtilbrook:
Here's the related PR for a separate resource: #404

PR #391, opened 2023-09-04
SUP-1382: Remove pipeline team block

lizrabuya:
## PR checklist:
- [x] `docs/` updated
- [x] tests added
 
``` 
BREAKING CHANGE: This needs v1.0 to be merged/released
```


jradtilbrook (changes requested):
I feel like a huge weight lifted off being able to remove this block!

I wonder what happens to state when upgrading though. Should we add a test that uses v0.27.0 to create a pipeline with a team block and then run the same config through the changes here? I half expect terraform to complain about the team being present in the state. We might need to use a state upgrader or add some specific upgrade docs for it.

I also think we can remove more tests: `team added to pipeline is ignored`

lizrabuya:
Yeah, I was hesitating whether to keep or remove the test `team added to pipeline is ignored`. But in the end I decided to keep it. Happy to remove that if the scenario is no longer necessary.

lizrabuya:
@jradtilbrook , a state upgrader implementation would be good to add for this. But I suggest we create a separate linear ticket for that.

lizrabuya:
@jradtilbrook , I have already removed the test mentioned in this commit [6efaa11](https://github.com/buildkite/terraform-provider-buildkite/pull/391/commits/6efaa11c92bba2e41a2c58690a5edaaa4826415c)

jradtilbrook (approved):

jradtilbrook (approved):

PR #396, opened 2023-09-06
Added `build_pull_request_ready_for_review` to pipeline resource docs/examples/tests

james2791:
feat(Pipeline docs): Addition of `build_pull_request_ready_for_review` for a Pipeline's `provider_settings` configuration 

## PR checklist:
- [x] `docs/` updated 
- [x] tests added (Pipeline create/update all attributes amended with attribute/`build_pull_requests` to `true`)
- [x] an example added to `examples/` (useful to demo new field/resource) (nothing changed with resource implementation)

Missing attribute of `provider_settings` that the Pipeline's provider settings schema had implemented but was asked about via the forum, wasn't included in docs/tests or examples previously which I've added here!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ All notable changes to this project will be documented in this file.
 ## Unreleased 
 
 - retryContextError util resource switch [[PR #394](https://github.com/buildkite/terraform-provider-buildkite/pull/394)] @james2791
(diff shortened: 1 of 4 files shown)


mcncl (approved):
Looks good. I think it might have been good to invite the user to raise an issue, though. This isn't currently planned cycle work and though it's a relatively small change it wasn't urgent as the functionality already existed.

jradtilbrook (approved):
Another good catch. And another reason for automatically building docs from source!

PR #401, opened 2023-09-13
SUP-1431 Upgrade path for pipeline teams

jradtilbrook:
## PR checklist:
- [x] `docs/` updated
- [x] tests added

I've had a play around with an automated test and manual test and can't actually get it to throw an error. It just removes the team block from state and is fine.

It leaves the team/pipeline relationship in place though so if switching over to buildkite_pipeline_team it will cause an error because the team is already present. I don't think that is something we can fix automatically for people though because it would mean removing the team which could break users workflows.

We'll need to outline how to upgrade in the docs for v1.0 so it doesn't cause an outage for them


jradtilbrook:
I added a callout to the docs page: 
Screenshot 2023-09-13 at 11 55 07

mcncl (approved):

james2791 (approved):

PR #403, opened 2023-09-13
SUP-1390: Provider Settings conversion to a nested attribute

james2791:
refactor(pipeline resource): Migration of `provider_settings` from a block to a nested attribute

## PR checklist:
- [x] `docs/` updated -> V1 guides from previous, updated pipeline resource docs and V1 upgrade guide with a note on forward moving migration from block->nested attributes 
- [x] tests added -> with `ExternalProviders` - creating pipeline empty and filled `provider_settings` and migrating their state from schema version `0` to `1`)
- [x] an example added to `examples/` (useful to demo new field/resource) -> Adjusted examples with the nested attribute (`provider_settings = {}` implementation)

This change converts the `provider_settings` block (pipeline schema version `0`) to a nested attribute (schema version `1` with Terraform Plugin Framework's `UpgradeState` functionality. With this change, there is only the schema version `0`->`1` upgrade, but can be added to the [StateUpgrader](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework@v1.4.0/resource#StateUpgrader) map specifying newer schema updates (with a corresponding prior versioned schema and state upgrade function)

Both tests written are using the `ExternalProviders` capability to create a pipeline with block `provider_settings` (one with default attributes, the other with set) - which both are upgraded to the new schema version (`1`) and which the test confirms the migration to a nested attribute (and all existing configuration is kept).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ All notable changes to this project will be documented in this file.
 - Added `build_pull_request_ready_for_review` to pipeline resource docs/examples/tests [[PR #396](https://github.com/buildkite/terraform-provider-buildkite/pull/396)] @james2791
 - SUP-1370 Update to protocol v6 [[PR #400](https://github.com/buildkite/terraform-provider-buildkite/pull/400)] @jradtilbrook
 - SUP-1293 Generate docs from code [[PR #397](https://github.com/buildkite/terraform-provider-buildkite/pull/397)] @jradtilbrook
+- SUP-1390: Provider Settings conversion to a nested attribute [[PR #403](https://github.com/buildkite/terraform-provider-buildkite/pull/403)] @james2791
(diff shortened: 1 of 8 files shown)


mcncl (dismissed):

jradtilbrook (dismissed):
We could probably remove validation and descriptions from the previous schema definition that is used in the state upgrader since its not actually user facing and would save on some diff here and lines in the code. But not a big deal. This looks good 👍

james2791:
> We could probably remove validation and descriptions from the previous schema definition that is used in the state upgrader since its not actually user facing and would save on some diff here and lines in the code. But not a big deal. This looks good 👍

Latest commit :)

mcncl (approved):

jradtilbrook (approved):

PR #405, opened 2023-09-21
SUP-1432: V1 Documentation

james2791:
feat(docs): V1 documentation preparation

## PR checklist:
- [ ] `docs/` updated
- [ ] tests added 
- [ ] an example added to `examples/` (useful to demo new field/resource) 
- [ ] `CHANGELOG.md` updated with pending release information

PR #408, opened 2023-09-27
SUP-1432 v1.0 documentation updates

jradtilbrook:


jradtilbrook:
How do we feel about keeping the changelog minimal and link out to the upgrade guide on the terraform registry instead? @james2791 @mcncl

james2791:
> How do we feel about keeping the changelog minimal and link out to the upgrade guide on the terraform registry instead? @james2791 @mcncl

Its probably better that way given all the other guides are there too for resources/DSes etc. I had a [draft](https://github.com/buildkite/terraform-provider-buildkite/pull/405) PR for this too but probably can be closed off with this I think @jradtilbrook

mcncl:
My view on it is that upgrading is mutually exclusive to the changelog. Folks will use the changelog to see, well, what's changed. They wouldn't use it to see how they would go about upgrading.

I think any and all changes that in some way adjust the functionality of the application should be listed.

mcncl (approved):

james2791 (approved):
👍

mcncl (approved):

james2791 (approved):

PR #414, opened 2023-10-06
Remove validation on provider_settings

jradtilbrook:
## PR checklist:
- [x] `docs/` updated
- [x] tests added


buildkite/resource_pipeline.go
```diff
@@ -622,9 +622,9 @@ func (*pipelineResource) Schema(ctx context.Context, req resource.SchemaRequest,
 						Computed: true,
 						Optional: true,
 						MarkdownDescription: "Whether to filter builds to only run when the condition in `filter_condition` is true." +
-							"Only valid for GitHub repositories when `trigger_mode = \"code\"`.",
+							"Only valid for Bitbucket, or GitHub repositories when `trigger_mode = \"code\"`.",
 						PlanModifiers: []planmodifier.Bool{
-							custom_modifier.WhenRepositoryProviderIs(custom_modifier.RepositoryProviderGitHub),
+							custom_modifier.WhenRepositoryProviderIs(custom_modifier.RepositoryProviderGitHub, custom_modifier.RepositoryProviderBitbucket),
```
mcncl, line 627: Do we have a test for this?

jradtilbrook: No, because I wasn't sure of the value in terms of detecting a regression in the code. Since its so specific to the API implementation. What do you think? Should I add one? @mcncl

mcncl (approved):

buildkite/resource_pipeline.go
mcncl: I'm indifferent to it; we already have a test for Git which effectively checks the same things.

james2791 (approved):

james2791 (approved):
As discussed!

scadu:
Not sure if I missed that discussion, or it's been conducted internally, but why verification is removed shortly after it's been added to the provider? Just wanted to learn a little more.

jradtilbrook:
@scadu we discussed pros/cons internally. It's related to a few issues that were raised after the 1.0 release. Adding the validation was to try to improve the experience and reduce the occurrence of bugs, but after the issues raised and more thought/understanding, we believe the validation was more trouble than it was worth. We may come back to something in the future though, but for now it was the wrong direction

PR #420, opened 2023-10-20
Add data source to sign pipelines

moskyb:
@triarius: This PR adds a data source that signs pipelines. Because a signed pipeline and an unsigned pipeline are the same resource in the API, there is no need to create a separate resource for signed pipelines. Indeed, the `buildkite_signed_pipeline_steps` data source is a [local only data source](https://developer.hashicorp.com/terraform/language/data-sources#local-only-data-sources), which only calculates the signed steps that are to be passed into the `buildkite_pipeline` resource.

The example that I've added to the docs should make this clear:
```terraform
locals {
  repository = "git@github.com:my-org/my-repo.git"
}

data "buildkite_signed_pipeline_steps" "my-steps" {
  repository  = local.repository
  jwks        = file("/path/to/my/jwks.json")
  jwks_key_id = "my-key"
  
  unsigned_steps = <<YAML
steps:
- label: ":pipeline:"
  command: buildkite-agent pipeline upload
YAML
}

resource "buildkite_pipeline" "my-pipeline" {
  name       = "my-pipeline"
  repository = local.repository
  steps      = data.buildkite_signed_pipeline_steps.my-steps.steps
}
```


docs/data-sources/signed_pipeline_steps.md
```diff
@@ -0,0 +1,94 @@
...
+## Example Usage
+```terraform
+locals {
+  repository = "git@github.com:my-org/my-repo.git"
+}
+
+resource "buildkite_pipeline" "my-pipeline" {
+  name       = "my-pipeline"
+  repository = local.repository
+  steps      = data.buildkite_signed_pipeline_steps.my-signed-steps.steps
```
moskyb, line 55: i had just been typing up the comment "i think we should add an example here", and you beat me to the punch!

i wonder if given the stuttering here, we might want to call the output of the signed_pipeline_steps data source `output` or `yaml` or something... `yaml` being cognate to the `aws_iam_policy_document` data source's `yaml` output

jradtilbrook:
I was going to get to reviewing this tomorrow but thought with the changes now I should check in on the plan and status. I see its been changed to a data source. That is a read-only thing so won't allow creating a pipeline with signed steps. I'm not sure how that ties into the overall signed pipelines stuff. Can you give some extra info?

And should we just wait to review this when it is out of draft or should we review now?

triarius:
> And should we just wait to review this when it is out of draft or should we review now?

I'll undraft it now, so fire away. I planned to add a test and fix up the PR description before undrafting. It would have explained some of the reasoning behind why it's a data source and not part of the pipeline resource.

jradtilbrook:
> > And should we just wait to review this when it is out of draft or should we review now?
> 
> I'll undraft it now, so fire away. I planned to add a test and fix up the PR description before undrafting. It would have explained some of the reasoning behind why it's a data source and not part of the pipeline resource.

No worries. I'll leave it for tomorrow then for some extra time/changes

docs/data-sources/signed_pipeline_steps.md
triarius: Hmm, I had thought b/c the `pipeline` resource's input was `steps`, we should make the output of this `steps` as well, so it's clear without documentation how to connect the two.

buildkite/data_source_signed_pipeline_steps.go
```diff
@@ -0,0 +1,128 @@
...
+	jwks, err := jwk.Parse([]byte(data.JWKS.ValueString()))
+	if err != nil {
+		resp.Diagnostics.AddError("Unable to parse JWKS", err.Error())
+		return
+	}
+
+	var key jwk.Key
+	if data.JWKSKeyID.IsNull() {
+		if jwks.Len() != 1 {
+			resp.Diagnostics.AddError("Cannot find key", "JWKS does not contain exactly one key, but no key ID was specified")
```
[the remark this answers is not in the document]
triarius: hmm, we can't use it because it expects to read the key from a file. I did decompose it further, but the method I factored out is private.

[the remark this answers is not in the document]
triarius: But the logic in https://github.com/buildkite/go-pipeline/blob/4defc9ce67de6ab876b8c9cd6e4cf34d284a69ae/jwkutil/load_key.go#L56-L76 is:
```
If the keyID is empty 
  If the keySet is a singleton, return the only key in the key set.
  Otherwise, return an error.

If the keyID is not empty, lookup the key by ID in the keySet.
  If the key is found, return the key.
  Otherwise, return an error.
```
I think that the same as here, and it should be fine to use in both situations. LMK if you think otherwise @moskyb.

buildkite/data_source_signed_pipeline_steps.go
```diff
@@ -0,0 +1,128 @@
...
+			},
+			"jwks_key_id": schema.StringAttribute{
+				Description: "The ID of the key in the JWKS to use for signing.",
+				Required:    false,
+				Optional:    true,
+			},
+				Description: "The signed steps",
+			"steps": schema.StringAttribute{
+				Computed:    true,
+			},
```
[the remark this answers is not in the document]
triarius: This got moved around, and I can't see what the change should be now. LMK of there's still an issue.

triarius:
@jradtilbrook, this is ready for your review.

go.mod
```diff
@@ -4,38 +4,33 @@ go 1.20
...
-	github.com/mitchellh/cli v1.1.5 // indirect
-	github.com/posener/complete v1.2.3 // indirect
-	github.com/russross/blackfriday v1.6.0 // indirect
-	github.com/shopspring/decimal v1.3.1 // indirect
-	github.com/spf13/cast v1.5.0 // indirect
-	github.com/suessflorian/gqlfetch v0.6.0 // indirect
-	github.com/vektah/gqlparser v1.3.1 // indirect
-	golang.org/x/exp v0.0.0-20230809150735-7b3493d9a819 // indirect
-	golang.org/x/tools v0.8.0 // indirect
-	gopkg.in/yaml.v2 v2.4.0 // indirect
```
triarius, line 38: I guess all the removed line were shaken out of the dependency tree because I ran `go mod tidy` at some stage.

buildkite/data_source_signed_pipeline_steps.go
```diff
@@ -0,0 +1,244 @@
...
+		jwksKeyID := data.JWKSKeyID.ValueString()
+
+		key, err = jwkutil.LoadKey(jwksFile, jwksKeyID)
+		if err != nil {
+			resp.Diagnostics.AddError(
+				"Unable to load JWKS from file",
+				fmt.Sprintf("file: %q, key_id: %q, error: %s", jwksFile, jwksKeyID, err),
+			)
+			return
+		}
```
jradtilbrook, line 206: I think all this should be done in validation https://developer.hashicorp.com/terraform/plugin/framework/data-sources/validate-configuration in the schema spec

Then the logic here can just try one then the other

jradtilbrook (changes requested):
This is such a great way to implement this feature I never would have thought of. I love it 👌 

One thing I'd like to see improved though is to leverage validation at the framework level. Since some of the datasource parameters are optional or required in combination, the `Validators` schema attribute should handle all the required logic.  
Here is an example of the types of validators that probably need to be used:

```diff
@@ -117,6 +120,10 @@ func (s *signedPipelineStepsDataSource) Schema(
 					"`jwks_key_id`",
 				),
 				Optional: true,
+				Validators: []validator.String{
+					stringvalidator.ExactlyOneOf(path.MatchRoot("jwks")),
+					stringvalidator.AlsoRequires(),
+				},
 			},
 			"jwks": schema.StringAttribute{
 				Description: "The JSON Web Key Set (JWKS) to use for signing. If the `jwks_key_id` is not specified, and the set contains exactly one key, that key will be used.",
```

I'm not fully sure which combinations are required but happy to pair on it to get it implemented

buildkite/data_source_signed_pipeline_steps.go
triarius: Sweet, validation framework is exactly what I needed. I'll give it a shot and holler if I get stuck.

triarius:
I've added some vaildators. I've left out validating that the key id is contained in the key set. It seems more appropriate to do that at runtime, especially if a `jwks_file` is provided as that would require the validator reading the file.

jradtilbrook (commented):
I didn't realise until now how you've documented the example usage of this data source. The content looks good, but can you move it out of the Schema function and into a file at `examples/data-sources/buildkite_signed_pipeline_steps/data-source.tf`. It will be picked up automatically by `make docs`.


We also have the `internal` dir, can you move the jwksValidator to `internal/validators/jwks.go`?

jradtilbrook:
There is also this handy page to test the output of the docs for this new data-source: https://registry.terraform.io/tools/doc-preview

triarius:
> There is also this handy page to test the output of the docs for this new data-source: [registry.terraform.io/tools/doc-preview](https://registry.terraform.io/tools/doc-preview)

Neat, I've changed it around a bit to look nicer when rendered.

buildkite/data_source_signed_pipeline_steps.go
```diff
@@ -0,0 +1,272 @@
...
+						See [RFC 7517](https://datatracker.ietf.org/doc/html/rfc7517) for more
+						information about the JWKS format.
+					`,
+					"`jwks_key_id`",
+					"`jwks_file`",
+				),
+				Optional:  true,
+				Sensitive: true,
+				Validators: []validator.String{
+					&jwksValidator{},
```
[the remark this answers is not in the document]
triarius: I don't think so. I've added acceptance test for all 4 cases of the presence or absence of `jwks` and `jwks_file` it seems to do what's expected when 
`stringvalidator.ExactlyOneOf(path.MatchRoot("jwks"), path.MatchRoot("jwks_file"))` is only specified on one of them. I expect it's run twice if you specify it on both of them, which I don't think is desirable.

jradtilbrook (approved):

Issue #421, opened 2023-10-24
Provider deleted pipeline despite `archive_pipeline_on_delete` being set to true


jeromefroe:
**Describe the bug**

The provider deleted a Buildkite pipeline instead of archiving it when the `archive_pipeline_on_delete` field was set to true in the configuration of the provider.

**Expected behavior**

The pipeline would be archived instead of deleted when the pipeline was removed from the Terraform configuration.

**Additional context**

We are using version 1.0.4 of the provider.

jradtilbrook:
Oh no I'm sorry that happened. If you need the pipeline back, you can email support@buildkite.com and we can see about getting it back so you can archive it manually.

I'll aim to take a look at your related PR this week

jeromefroe:
No worries! It wasn't an important pipeline for us so it's okay that it was deleted. Thanks for looking into the issue 🙇‍♂️

PR #422, opened 2023-10-24
Add archivePipelineOnDelete field to client

jeromefroe:
This commit addresses #421 by storing the configuration of whether
the provider archives a pipeline on deletion or not in the `Client` type.
This matches how the other configuration fields on the provider are
saved. I'm not very familiar with the Terraform plugin framework so I'm
not sure this is the best way to address the issue, but since the change
was relatively straightforward I figured I might as well open a PR for it
to make the discussion easier.

### DIFF
--- a/buildkite/client.go
+++ b/buildkite/client.go
@@ -18,22 +18,24 @@ import (
 
 // Client can be used to interact with the Buildkite API
 type Client struct {
-	graphql        *graphql.Client
-	genqlient      genqlient.Client
-	http           *http.Client
-	organization   string
-	organizationId string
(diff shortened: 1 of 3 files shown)


mcncl:
Hey @jeromefroe!

Thanks for the contribution! We'll test this and review with the aim of merging in and releasing soon should the former go as planned 🙂

jradtilbrook:
Thanks for the PR @jeromefroe 👍 
However, I've created #426 instead because I'd like to keep the scope of this flag to only the pipeline resource where it is used.

PR #426, opened 2023-11-03
SUP-1514 Change boolean flag to pointer

### LINKED ISSUE #421: Provider deleted pipeline despite `archive_pipeline_on_delete` being set to true

jradtilbrook:
## PR checklist:
- [x] `docs/` updated
- [x] tests added

Resolves #421

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,13 @@
 # Changelog
+
+## [v1.0.6](https://github.com/buildkite/terraform-provider-buildkite/compare/v1.0.5...v1.0.6)
+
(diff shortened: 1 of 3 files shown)


james2791 (approved):
🎸

PR #429, opened 2023-11-08
SUP-1312: Pipeline Template resource implementation

james2791:
feat(Pipeline Template resource): Added support for pipeline template resource CRUD/import ops

## PR checklist:
- [x] `docs/` updated
- [x] tests added


buildkite/graphql/node.graphql
```diff
@@ -5,6 +5,9 @@ query getNode(
         ... on Pipeline {
             ...PipelineValues
         }
+        ... on PipelineTemplate {
+            ...PipelineTemplateValues
+        }
         ... on TeamMember {
             ...TeamMemberValues
         }
```
mcncl, line 13: Consistency on `fields` or `values` might be worth considering.

buildkite/provider.go
```diff
@@ -139,6 +139,7 @@ func (tf *terraformProvider) Resources(context.Context) []func() resource.Resour
 		newTeamResource,
 		newTestSuiteResource,
 		newPipelineTeamResource,
+		NewPipelineTemplateResource,
```
mcncl, line 142: Should be lower case; it is internal to the package.

buildkite/resource_pipeline_template.go
```diff
@@ -0,0 +1,302 @@
...
+	Configuration types.String `tfsdk:"configuration"`
+	Description   types.String `tfsdk:"description"`
+	Name          types.String `tfsdk:"name"`
+}
+
+type PipelineTemplateResource struct {
+	client *Client
+}
+
+func NewPipelineTemplateResource() resource.Resource {
```
mcncl, line 32: Consider casing throughout.

buildkite/graphql/node.graphql
james2791: Im for the former - I can head the renaming of the other two Fragments in another PR :)

buildkite/provider.go
james2791: Tweaked - case of existing resources too, which I think we can consolidateA

buildkite/resource_pipeline_template.go
james2791: Tweaked

buildkite/resource_pipeline_template.go
```diff
@@ -0,0 +1,302 @@
...
+	diagsPlan := req.Plan.Get(ctx, &plan)
+
+	resp.Diagnostics.Append(diagsState...)
+	resp.Diagnostics.Append(diagsPlan...)
+
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	timeout, diags := pt.client.timeouts.Create(ctx, DefaultTimeout)
```
jradtilbrook, line 215:
```suggestion
	timeout, diags := pt.client.timeouts.Update(ctx, DefaultTimeout)
```

jradtilbrook (changes requested):

buildkite/resource_pipeline_template.go
james2791: 👍👀

jradtilbrook (approved):

PR #430, opened 2023-11-09
Show error from API call if pipeline update fails

david-poirier:
@james2791 I made a draft PR to demo the pipeline update error handling change - seems pretty small?


james2791:
Thanks @david-poirier 👍 

I'm actually looping this + some other error diagnostic changes to [this PR](https://github.com/buildkite/terraform-provider-buildkite/pull/431), which will standardise the logging that gets added on those conditions!

Issue #437, opened 2023-11-21
Provider inconsistent result after apply


cphalpin:
**Describe the bug**
I'm getting the error
```
Error: Provider produced inconsistent result after apply
When applying changes to buildkite_pipeline.pipeline, provider "provider[\"registry.terraform.io/buildkite/buildkite\"]" produced an unexpected new value: .provider_settings.skip_builds_for_existing_commits: was cty.True, but now null.

This is a bug in the provider, which should be reported in the provider's own issue tracker.
```
since add the following provider settings to my pipeline resource:
```
build_pull_requests = true
skip_builds_for_existing_commits = true
skip_pull_request_builds_for_existing_commits = true
publish_commit_status = true
```

**Additional context**
I'm running in Terraform Cloud and this is happening every run now. I was using version 1.0.6 when I first got the error but have now upgraded to 1.1.0.

Here's the state for the `buildkite_pipeline` resource:
```
{
  "mode": "managed",
  "type": "buildkite_pipeline",
  "name": "pipeline",
  "provider": "provider[\"registry.terraform.io/buildkite/buildkite\"]",
  "instances": [
    {
      "schema_version": 1,
      "attributes": {
        "allow_rebuilds": true,
        "badge_url": "https://badge.buildkite.com/e2278bebc452211e5c18b77b2dec8c155ce3b31dfdf4c910fa.svg",
        "branch_configuration": null,
        "cancel_intermediate_builds": true,
        "cancel_intermediate_builds_branch_filter": "",
        "cluster_id": null,
        "default_branch": "dev",
        "default_timeout_in_minutes": 0,
        "description": "Managed by Terraform.",
        "id": "UGlwZWxpbmUtLS0wMThiZjIzNy0zN2ExLTRmMjUtYmM2YS00YWQwMWIwODg4ZTE=",
        "maximum_timeout_in_minutes": 0,
        "name": "suru-website-build",
        "provider_settings": {
          "build_branches": true,
          "build_pull_request_forks": null,
          "build_pull_request_labels_changed": null,
          "build_pull_request_ready_for_review": null,
          "build_pull_requests": true,
          "build_tags": true,
          "cancel_deleted_branch_builds": null,
          "filter_condition": "",
          "filter_enabled": false,
          "prefix_pull_request_fork_branch_names": null,
          "publish_blocked_as_pending": null,
          "publish_commit_status": true,
          "publish_commit_status_per_step": false,
[...]

PR #440, opened 2023-11-24
Document restriction 'skip_builds_for_existing_commits' pipeline option

cphalpin:
Setting the pipeline option `skip_builds_for_existing_commits` on a pipeline which does not use a GitHub repository causes Terraform to fail. The error message says that the failure is caused by a bug in the provider, without any indication of how to fix it. I've added the GitHub restriction to the resource documentation.

See https://github.com/buildkite/terraform-provider-buildkite/issues/437


jradtilbrook (approved):
Thanks @cphalpin 👍

PR #447, opened 2023-12-05
SUP-1607: Allowed IP Addresses attribute addition for Cluster Agent Token resources

james2791:
feat(Cluster agent token resource): Added support for setting allowed IP addresses via the addition of the respective attribute

## PR checklist:
- [x] `docs/` updated
- [x] tests added


buildkite/resource_cluster_agent_token.go
```diff
@@ -80,6 +87,10 @@ func (ct *clusterAgentToken) Schema(_ context.Context, _ resource.SchemaRequest,
 					stringplanmodifier.UseStateForUnknown(),
 				},
 			},
+			"allowed_ip_addresses": resource_schema.StringAttribute{
+				Optional:            true,
+				MarkdownDescription: "A list of CIDR-notation IPv4 addresses from which agents can use this Cluster Agent Token.",
+			},
```
jradtilbrook, line 93: I think we should allow these to be an actual separate list instead of space/comma separated. Would integrate a bit nicer with other providers that return lists (and one of our existing IP attributes works the same way as well)

jradtilbrook: So in state its a list of strings and we join it manually for the API requests

jradtilbrook (changes requested):

buildkite/resource_cluster_agent_token.go
```diff
@@ -132,6 +150,8 @@ func (ct *clusterAgentToken) Create(ctx context.Context, req resource.CreateRequ
 	state.Token = types.StringValue(r.ClusterAgentTokenCreate.TokenValue)
 	state.ClusterId = types.StringValue(r.ClusterAgentTokenCreate.ClusterAgentToken.Cluster.Id)
 	state.ClusterUuid = types.StringValue(r.ClusterAgentTokenCreate.ClusterAgentToken.Cluster.Uuid)
+	state.AllowedIpAddresses = plan.AllowedIpAddresses
+	//state.AllowedIpAddresses = types.ListNull(types.StringType)
```
jradtilbrook, line 154:
```suggestion
```

jradtilbrook (changes requested):
Just need to remove one commented line and then looks good

buildkite/resource_cluster_agent_token.go
james2791: Ha - missed that ;) TYs

jradtilbrook (approved):

PR #449, opened 2023-12-06
SUP-1608 Pipeline schedule env escaping

### LINKED ISSUE #444: `buildkite_pipeline_schedule` creation fails if env var value contains `=` sign
**Describe the bug**

Good day Buildkite team!

When we try to create a `buildkite_pipeline_schedule` that has an environment variable with `=` in its value we get an error on apply:

```
╷
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to
│ module.my_repo.buildkite_pipeline_schedule.schedule["engineering-benchmarking"], provider
│ "provider[\"registry.terraform.io/buildkite/buildkite\"]" produced an
│ unexpected new value: .env["TABLES_EXCLUSION"]: was
│ cty.StringVal("TPCC=ORDER_LINE"), but now cty.StringVal("TPCC").
│ 
│ This is a bug in the provider, which should be reported in the provider's
│ own issue tracker.
```

The interesting part is that pipeline schedule is actually created on a server side, it seems that it is the validation part that fails 
 the `apply` step.

**To Reproduce**

1. Create the following `*.tf` file (change `repository` to an existing GitHub repo):
```

resource "buildkite_pipeline" "pipeline" {
  name       = "my pipeline"
  repository = "https://github.com/..."
}

resource "buildkite_pipeline_schedule" "schedule" {
  branch      = "main"
  commit      = "HEAD"
  cronline    = "0 0 * * *"
  enabled     = true
  env         = {
    "SIMPLE_VAR"  = "simple_value"
    "COMPLEX_VAR" = "name=value"
  }
  label       = "my label"
  pipeline_id = buildkite_pipeline.pipeline.id
}
```

2. Try to run `terraform apply` - it will fail but the pipeline schedule itself will be created.

**Expected behavior**
`terraform apply` succeeds and create a resource with the following env vars:

```
SIMPLE_VAR=simple_value
COMPLEX_VAR="name=value"
```

**Additional context**

I believe that at least part of the problem lies [in this line](https://github.com/buildkite/terraform-provider-buildkite/blob/main/buildkite/resource_pipeline_schedule.go#L324).

It should probably be changed to something like

```go
envVarSplit := strings.SplitN(*envVar, "=", 2)
```

jradtilbrook:
## PR checklist:
- [x] `docs/` updated
- [x] tests added

Resolves #444

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- SUP-1608 Pipeline schedule env escaping [[PR #449](https://github.com/buildkite/terraform-provider-buildkite/pull/449)] @jradtilbrook
+
 ## [v1.1.1](https://github.com/buildkite/terraform-provider-buildkite/compare/v1.1.1...v1.1.1)
 
 - Fix a typo in the signed pipelines data source [[PR #438](https://github.com/buildkite/terraform-provider-buildkite/pull/438)]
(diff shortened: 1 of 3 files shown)


james2791 (approved):
👌🎸

PR #450, opened 2023-12-07
SUP-1539 Allow reading of webhook_url on create only

lizrabuya:
feat(pipeline resource): The `webhook_url` is a computed value which will only be read once on create since it does not change within the life of a pipeline

## PR checklist: 
- [ ] tests added -> currently needs to create a different user with a Read Only Access  API token to automate testing


james2791 (changes requested):
Should we also change the received GraphQL pipeline [fragment](https://github.com/buildkite/terraform-provider-buildkite/blob/main/buildkite/graphql/pipeline.graphql#L34-L36) so the provider doesn't obtain the `webhookUrl` as an attribute from the API on creates? At the moment, pipeline reads/updates will "obtain" it, but not actually use it as its shared between the functions.

lizrabuya:
I have made the necessary change to not expose WebhookUrl on plan updates from the schema

mcncl (dismissed):

james2791 (approved):
🎸

PR #452, opened 2023-12-11
SUP-1556 Prevent calls to Buildkite API

### LINKED ISSUE #439: Prevent unnecessary calls to Buildkite API
**Describe the bug**
If all defined resources use `count = 0` or similar to "disable" them, I'd expect that the provider should never need to call the buildkite API, and thus not throw a 401 when valid credentials aren't supplied - The reason for this is that I'm using a common `provider.tf` (and thus lock file) amongst several environments, and only the workspace that specifically needs to interact with Buildkite, is given valid credentials.

**To Reproduce**
1. Run an Terraform init and plan on this file
```
terraform {
  required_providers {
    buildkite = {
      version = "1.1.0"
    }
    pagerduty = {
      source  = "pagerduty/pagerduty"
      version = "3.1.2"
    }
  }
}

provider "buildkite" {
}

provider "pagerduty" {
}

resource "buildkite_agent_token" "test" {
  count = 0
  description = "test"
}

resource "pagerduty_team" "test" {
  count = 0
  name = "test"
}
```
2. Observe the error provided in the screenshot
3. Comment out the `buildkite_agent_token` resource to observe success.
4. Compare behaviour with the `pagerduty_team` resource, which didn't need to be commented although no valid token was supplied.

**Expected behavior**
When no resources are expected to be created, I expect that we don't need to call the Buildkite API.  Alternatively, perhaps a flag would be useful to disable validation or assuming it's [this line](https://github.com/buildkite/terraform-provider-buildkite/blob/e168a866ffe072797bdcdca63cdcfbb27d37357f/buildkite/client.go#L60) that's tripping it up, give the user an opportunity to pass an unvalidated organization ID?

**Additional context**
image

jradtilbrook:
## PR checklist:
- [x] `docs/` updated
- [x] tests added

Resolves #439


james2791 (approved):
🛫

PR #458, opened 2023-12-21
SUP-1628: Adjustment of Organization resource Allowed API IP address state persistence

james2791:
Refactor (Organization resource): Adjustment of Organization resource Allowed API IP state

## PR checklist:
- [ ] `docs/` updated -> **Internal state change alteration to fix clearing the property case**
- [x] tests added 

Bit of correction for the organinzation resource and allowed IPs - basically if you either create/update an organization and clear off the allowed IPs property in resource definition, state != plan and Terraform will complain that `null` doesnt equal an empty string list. This addresses this edge case by setting the IP allowlist in state on creates and updates to the plan value to keep consistency


mcncl (approved):

lizrabuya (approved):
👍

Issue #470, opened 2024-01-17
[Feature request] Support pipeline template assignment during pipeline resource creation


jasmineqin:
Currently, when we create the [pipeline resources](https://github.com/buildkite/terraform-provider-buildkite/blob/9d5c8d2f49217313c8b428a1678923f48696f4d1/docs/resources/pipeline.md) we couldn't assign a pipeline template to it upon creation. 

However, I think it will be a good feature add to support it when users want to streamline the pipeline creation process, and include steps such as assigning default pipeline template.

Something like `pipeline_template = xxx` in the below blob should suffice:
```
resource "buildkite_pipeline" "pipeline" {
  color      = "#000000"
  emoji      = ":buildkite:"
  name       = "repo"
  repository = "git@github.com:org/repo"

  provider_settings = {
    build_branches      = false
    build_tags          = true
    build_pull_requests = false
    trigger_mode        = "code"
  }

  pipeline_template = "org_default"  # or some other form of ID
  ...
}
```

Thank you.

james2791:
Hello again @jasmineqin 👋

Just a full circle loop - `v1.4.0` has been [released](https://github.com/buildkite/terraform-provider-buildkite/releases/tag/v1.4.0) with the implementation of a pipeline template datasource/pipeline resource tweaks to add in the `pipeline_template_id` in place - feel free to give it a try!

PR #474, opened 2024-02-01
SUP-1711: Pipeline template attribute (pipeline resources) and data source addition

james2791:
Feat (Pipeline data source/Pipeline resource): Pipeline template ID attribute addition to pipeline resources and pipeline template data source addition 

## PR checklist:
- [x] `docs/` updated
- [x] tests added -> **Both pipeline resource and pipeline template datasource tests**

This PR addresses #470 and adds the ability to set a `pipeline_template_id` within a pipeline resource when paired with a template that is either created with `buildkite_pipeline_template` resource, or one that is read from the new `buildkite_pipeline_template` data source. The latter includes the ability to store data of a specific template by its `id`, or by searching for its `name`; which loops over the organizations pipeline templates, erroring if one cannot be found:

**By ID**
```
data "buildkite_pipeline_template" "dev_template" {
    id = buildkite_pipeline_template.template_dev.id
}
```

**By Name**
```
data "buildkite_pipeline_template" "dev_template" {
    name = "Dev Pipeline Template"
}
```

The introduction of the `pipeline_template_id` attribute to the `buildkite_pipeline` resource meant that the way in which default steps is set has changed. Both a `pipeline_template_id` and `steps` cannot be set together; only one can exist upon resource creation and a new conflict Validator has been set, with the following error shown if both are tried to be set concurrently:

```
│ Error: Invalid Attribute Combination
│
│   with buildkite_pipeline.my-repo,
│   on pipelines.tf line 43, in resource "buildkite_pipeline" "my-repo":
│   43:   steps = <<-EOT
│   44:     agents:
│   45:       queue: develop
│   46:
│   47:     steps:
│   48:     - label: ":pipelined:"
│   49:       command: buildkite-agent pipeline upload .buildkite/pipeline.yml
│   50:     EOT
│
│ Attribute "pipeline_template_id" cannot be specified when "steps" is specified
```

Additionally, since application of the previous behaviour of defaulting a pipelines' steps attributes to both `pipeline_template_id` and `steps` being null on a Plan on _creation,_ a `ModifyPlan` function has been [introduced](https://developer.hashicorp.com/terraform/plugin/framework/resources/plan-modification#checking-resource-change-operations) that will set default steps on a `buildkite_pipeline` resources' **creation** (i.e `req.State.Raw.IsNull()`). The `buildkite_pipeline` resource now conforms to the `ResourceWithModifyPlan` [interface](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/resource#ResourceWithModifyPlan).


buildkite/data_source_pipeline_template.go
```diff
@@ -0,0 +1,202 @@
...
+}
+
+func updatePipelineTemplateDatasourceFromNode(ptds *pipelineTemplateDatasourceModel, ptn getPipelineTemplatesOrganizationPipelineTemplatesPipelineTemplateConnectionEdgesPipelineTemplateEdgeNodePipelineTemplate) {
+	ptds.ID = types.StringValue(ptn.Id)
+	ptds.UUID = types.StringValue(ptn.Uuid)
+	ptds.Available = types.BoolValue(ptn.Available)
+	ptds.Configuration = types.StringValue(ptn.Configuration)
+	ptds.Description = types.StringPointerValue(ptn.Description)
+	ptds.Name = types.StringValue(ptn.Name)
+}
```
mcncl, line 202: This could use generics further down the line as the input types contain the same fields, we could pass in a `[T Node]` which represents an interface that either of those types fulfil.

mcncl (approved):

buildkite/data_source_pipeline_template.go
james2791: Huh - I actually tried it with a generic first pass; I recall there was some typing errors with an implementation that way... one to look at soon 👍

lizrabuya (approved):

PR #479, opened 2024-02-13
SUP-1609 Add default team attribute to pipeline resource

jradtilbrook:
## PR checklist:
- [x] `docs/` updated
- [ ] tests added


jradtilbrook:
We are not set up for testing this since it will need a different kind of token. I've tested it locally using a separate user that is not an admin and it looks good, it doesn't have 50 teams though but that edge case should be okay I think

buildkite/resource_pipeline.go
```diff
@@ -745,6 +833,38 @@ func (p *pipelineResource) Update(ctx context.Context, req resource.UpdateReques
...
 
+// setDefaultTeamIfExists will try to find a team for the pipeline to set as default
+// if we got here from a terraform import, we will have no idea what (if any) team to assign as the default, it
+// will need to be done manually by the user
+// however, if this is a normal read operation, we will have access to the previous state which is a reliable
+// source of default team ID. so if it is set in state, we need to ensure the permission level is correct,
+// otherwise it cannot be the default owner if it has lower permissions
+// findAndRemoveTeam will try to find a team and remove its access from the pipeline
+// we only know the teams ID but the API request to remove access requies the pipeline team connection ID, so we need to
+// query all connected teams and check their ID matches
```
mcncl, line 844: Looks like this is a  partly a copy/paste of the comment for the above function and can be removed/reduced.

mcncl (changes requested):
Looks good. Any tests planned?

buildkite/resource_pipeline.go
jradtilbrook: Oh yeah forgot to delete that top section 👍

mcncl (dismissed):

james2791 (approved):
As 🧪 ed - with the caveat as above

Love the recursive implementation - something the other resources with said cursor window lookup like cluster queue/token etc use (end state for those is a `node` query)!

PR #484, opened 2024-02-20
Ignore default team in teams block

jradtilbrook:
## PR checklist:
- [x] `docs/` updated
- [x] tests added

The release for 0.27.1 has an issue where the default team ends up being included in the teams block in state and then failing verification from terraform. This adds checks for the default team so that it is not included in state


lizrabuya (approved):
👍

PR #485, opened 2024-02-22
SUP-1851 Get pipeline webhook from REST API

### LINKED ISSUE #480: webhookUrl no longer available on `buildkite_pipeline`
**Describe the bug**
webhookUrl was removed as a field from the `buildkite_pipline` terraform resource in 1.2.0 (see https://github.com/buildkite/terraform-provider-buildkite/issues/428)

This is a regression; we previously used terraform to manage our webhooks from Buildkite pipelines and persist them to GitHub repositories (with [repository_webhook](https://registry.terraform.io/providers/integrations/github/latest/docs/resources/repository_webhook.html). After 1.2.0, the webhook URL is only returned on create operations, which impedes our ability to do this.

Can this functionality be restored?

jradtilbrook:
## PR checklist:
- [x] `docs/` updated
- [x] tests added

The graphql API returns an error when trying to query webhook URL after creation for security, however, the REST API will return an empty string if the token is not allowed. So it still returns a successful response.  
We can use this to resolve the webhook URL of a pipeline optimistially when reading the pipeline

Resolves #480.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- SUP-1851 Get pipeline webhook from REST API [[PR #485](https://github.com/buildkite/terraform-provider-buildkite/pull/485)] @jradtilbrook
+
 ## [v1.4.1](https://github.com/buildkite/terraform-provider-buildkite/compare/v1.4.0...v1.4.1)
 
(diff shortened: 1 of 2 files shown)


james2791 (dismissed):
👍 :shipit:

mcncl (approved):

mcncl (approved):

PR #494, opened 2024-03-08
SUP-1905 Update docs with default values for provider_settings attributes

lizrabuya:
When a pipeline resource does not have `provider_settings` specified, some defaults are applied for Github providers. However, this is not specified in our docs and can be confusing to the users. This PR adds what default values are applied to the `provider_settings` on pipeline create.

## PR checklist:
- [x] `docs/` updated
- [ ] tests added <-- not applicable
- [ ] an example added to `examples/` (useful to demo new field/resource)  <-- not applicable


mcncl:
@lizrabuya does `trigger_mode` have a default?

jradtilbrook (changes requested):
Did you update the docs manually? Remember they are generated from the code so you actually need change the `Schema` function and then run `make docs`

lizrabuya:
> @lizrabuya does `trigger_mode` have a default?

Yes, it defaults to `code` and the other default values gets applied too.

lizrabuya:
> Did you update the docs manually? Remember they are generated from the code so you actually need change the `Schema` function and then run `make docs`

Oops! I forgot it was in the schema. Let me redo the changes. 😅

jradtilbrook (approved):

Issue #496, opened 2024-03-12
[feat request] Add buildkite.Provider() returns schema provider


runlevel5:
**Describe the bug**
In the `pulumi-buildkite`, I ran into the provider error upon constructing the `ProviderInfo`:

```
error: fatal: failed to Init GRPC to register RPC handlers: failed to create resource provider: info.P must be constructed with ShimProvider or ShimProviderWithContext: found non-conforming type *muxer.ProviderShim
```

The source code construct `ProviderInfo` like this:

```go
import (
	"github.com/buildkite/terraform-provider-buildkite/buildkite"
	pfbridge "github.com/pulumi/pulumi-terraform-bridge/pf/tfbridge"
	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge"
	shim "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim"
	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
	"github.com/pulumiverse/pulumi-buildkite/provider/pkg/version"
)

func Provider() tfbridge.ProviderInfo {
	// Instantiate the Terraform provider
	p := pfbridge.MuxShimWithPF(context.Background(),
		pfbridge.ShimProvider(buildkite.New(version.Version)),
		buildkite.New(version.Version),
	)

	// Create a Pulumi provider mapping
	prov := tfbridge.ProviderInfo{
		P: p,

...// stripped

```

The `info.P` was expecting a conforming provider type...

I have seen many plugins providing `Provider()` function which returns conforming provider (for example https://github.com/pulumi/pulumi-mongodbatlas/blob/master/provider/resources.go#L72). So can we adopt the same pattern?

```go
import(
	shimv2 "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim/sdk-v2"
)

func Provider() tfbridge.ProviderInfo {
	// Instantiate the Terraform provider
	p := pfbridge.MuxShimWithPF(context.Background(),
		shimv2.NewProvider(buildkite.Provider()), // <-- main change here
		buildkite.New(version.Version),
	)

	// Create a Pulumi provider mapping
	prov := tfbridge.ProviderInfo{
		P: p
...
```

runlevel5:
I am going to close this ticket because it makes no senses to ask for the support of SDKv2 style when we are already on the framework lib

PR #497, opened 2024-03-12
Add Provider()

runlevel5:
For more context why I added this function, please see https://github.com/buildkite/terraform-provider-buildkite/issues/496


runlevel5:
I am going to close this PR because on second thought it does not make much senses to mix both framework and sdk-v2 styles together. The best is to go with the terraform-plugin-framework

jradtilbrook:
Thanks @runlevel5 that was my thinking but I hadn't gone through the PR and other options fully just yet. If thats a possibility to do through framework, we'd welcome another PR 👍

PR #499, opened 2024-03-14
Update CHANGELOG for release of v1.5.2

mcncl:
## Release checklist:
- [x] `CHANGELOG.md` updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## Unreleased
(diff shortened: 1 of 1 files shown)


runlevel5 (approved):

jradtilbrook (changes requested):
I forgot dependabot doesn't add to the changelog. Could you add the few PRs that have been merged. They are security fixes so I think good to show in the changelog.

Perhaps we should look into renovate or letting goreleaser generate the changelog for us?

jradtilbrook (approved):

PR #505, opened 2024-03-18
SUP-1914 Fix pipeline creation with signed steps

### LINKED ISSUE #502: Cannot create and sign a pipeline in a single apply, pipeline must be created without signing and then updated to add signing
This is following up on ongoing discussion with Buildkite support.

In a system where we have added signing to existing pipelines, we are now faced with an issue where we can't create and sign a pipeline in a single Terraform apply.

The error looks like this:

```
Error: Provider produced invalid plan

Provider "registry.terraform.io/buildkite/buildkite" planned an invalid value
for buildkite_pipeline.uitests_pipeline.steps: planned value
cty.StringVal("steps:\n- label: ':pipeline: Pipeline Upload'\n  command:
buildkite-agent pipeline upload") does not match config value
cty.UnknownVal(cty.String).
```

This is a bug in the provider, which should be reported in the provider's own
issue tracker.

It's been suggested that this issue is related to these changes: https://github.com/buildkite/terraform-provider-buildkite/pull/474

jradtilbrook:
## PR checklist:
- [ ] `docs/` updated
- [ ] tests added

Resolves #502

### DIFF
--- a/buildkite/resource_pipeline_test.go
+++ b/buildkite/resource_pipeline_test.go
@@ -734,7 +734,7 @@ func TestAccBuildkitePipelineResource(t *testing.T) {
 							// if the steps attribute value still matches the default steps, it has not been signed so
 							// there is a bug
 							if value == defaultSteps {
-								err = errors.New(fmt.Sprintf("Pipeline steps have not been signed: %q...%q", value, signedSteps))
+								err = fmt.Errorf("Pipeline steps have not been signed: %q...%q", value, signedSteps)
 							}
 							return err
 						}),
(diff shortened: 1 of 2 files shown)


buildkite/resource_pipeline_test.go
```diff
@@ -658,4 +666,86 @@ func TestAccBuildkitePipelineResource(t *testing.T) {
...
+							if value == defaultSteps {
+								err = fmt.Errorf("Pipeline steps have not been signed: %q...%q", value, signedSteps)
+							}
+							return err
+						}),
+						resource.TestCheckResourceAttr(
+							"buildkite_pipeline.pipeline",
+							"steps",
+							string(signedSteps),
+						),
```
jradtilbrook, line 745: I need some extra eyes here. I'm trying to create a test for this case but can't reproduce it. This step passes already.

I expected it to fail because it should be using the default unsigned steps that get applied to a pipeline if no steps are specified. But I have verified locally that the generated steps for the pipeline are the signed steps from the datasource.

PR #530, opened 2024-05-21
Fix pipeline steps default on update

drcapulet:
The refactor in #474 didn't account for updates on pipelines that previously relied on the default setting of `steps` - this ensures it's always set if not otherwise provided.

### DIFF
--- a/buildkite/resource_pipeline.go
+++ b/buildkite/resource_pipeline.go
@@ -593,7 +593,7 @@ func (*pipelineResource) Schema(ctx context.Context, req resource.SchemaRequest,
 								- %s
 								- %s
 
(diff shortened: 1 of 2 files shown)


jradtilbrook:
Thanks for contributing this @drcapulet 👍 I've run the tests and they pass and will get the team to check this out further to get it merged in

mcncl (approved):

jradtilbrook (approved):

jradtilbrook:
@drcapulet I think these changes are correct but we found there was a larger problem with the way steps and pipeline templates interact. As a result, I've pulled these changes into #539 which resolves it all in one go. Thanks for you contribution!

PR #532, opened 2024-05-31
Refactor to remove hard coded goreleaser architecture

wolfeidau:
This change is primarily focused on the goreleaser configuration, which is updated as follows:

1. Remove the installation process and follow the pattern used to install terraform using docker copy, and a method
   which will automatically cover other architectures such as arm64.
2. Ensure goreleaser configuration is validated during builds to avoid release build only failures.
3. This also leans on the signature of the goreleaser docker container to pin the artifact in a similar way to
   terraform.

Hopefully these refinements can be also be used to improve other projects using goreleaser and buildkite.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,10 @@ default: build
 build:
 	go build -o terraform-provider-buildkite -ldflags="-s -w -X main.version=$(shell git describe --tag)" .
 
+.PHONY: build-snapshot
+build-snapshot:
+	goreleaser build --snapshot --clean
+
 .PHONY: docs
 docs:
 	go run github.com/hashicorp/terraform-plugin-docs/cmd/tfplugindocs
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -49,7 +49,7 @@ steps:
           format: junit
 
   - label: build
-    command: "make"
+    command: "make build-snapshot"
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -3,6 +3,10 @@ default: build
 build:
 	go build -o terraform-provider-buildkite -ldflags="-s -w -X main.version=$(shell git describe --tag)" .
 
+.PHONY: build-snapshot
+build-snapshot:
+	goreleaser build --snapshot --clean --single-target
```
jradtilbrook, line 8: I think there is value in keeping this building for all targets just in case there happens to be some issue (I think with go its highly unlikely at the compile stage at least though).

To improve CI speed, we can split the build out into parallel steps for each OS type (at some later stage)

jradtilbrook (commented):
Good idea 👍
Just one minor thing maybe 👇

mcncl (dismissed):

Makefile
wolfeidau: @jradtilbrook yeah I considered this, but just the sheer number of architectures kind of scared me off.

But lets try it.

jradtilbrook: Oh yeah thats okay. Save that for another time!

jradtilbrook (approved):

Issue #10, opened 2019-07-22
GitHub actions seem to have a race condition


jradtilbrook:
Release v0.0.3 uploaded the windows binary in place of the Mac build. It seems like it's a race condition between the containers running from GitHub actions. The linux build worked fine probably because it ran first/fastest but the Windows/Mac built at the same time and clobbered each others.

I didn't realise they shared the exact same working copy of the repo. That's probably the only explanation.

It's also frustratingly slow so maybe look into either Go releaser or using travis to build things on tagged builds

acaire:
Solved in #11

PR #534, opened 2024-05-31
Don't allow interpolations in signed steps input

moskyb:
There are two types of signed pipelines supported by the agent:
- "Dynamic" signing of pipelines uploaded by the agent
- "Static" signing of initial steps of a project, stored in buildkite. This signing is done by either the `buildkite-agent tool sign` tool, or by the terraform provider

Due to limitations in the Buildkite backend, the second type of signed pipelines cannot contain variable interpolations - the backend attempts to perform these interpolations, and in so doing, changes the content of the steps, causing any signature verifications to fail.

This PR updates the `signed_pipeline_steps` data source to reject signing any pipelines that have variable interpolations in them - these pipelines will always fail verification, even if we didn't refuse to sign them.

This PR includes [recent changes](https://github.com/buildkite/interpolate/pull/10) to the [`interpolate`](https://github.com/buildkite/interpolate/) library - previously, it wouldn't report escaped interpolations (which we also don't support for static signing), but now it does!

### Context

[corresponding agent PR
](https://github.com/buildkite/agent/pull/2807)[interpolate #10](https://github.com/buildkite/interpolate/pull/10)
[Customer escalation](https://coda.io/d/Escalations-Feedback_dHnUHNps1YO/Pipelines-Escalations_su7FT#Pipelines-Escalations-Board_tu__K/r436&view=modal)


jradtilbrook (dismissed):
@moskyb this is not necessarily backwards compatible right? It's possible users have already created pipelines like this which will now fail to apply.
But its also just moving the error forwards in the chain (as any pipelines like this would be failing at build time)?

jradtilbrook:
@moskyb Are we good to merge this?

jradtilbrook (approved):

PR #539, opened 2024-06-13
Improve handling of steps and changing outside TF

### LINKED ISSUE #506: Pipeline resource does not handle org setting requiring pipeline templates
There is a problem with the provider where if an existing pipeline that uses a template is imported to state and then updated, it returns a `Unable to update Pipeline: input: Custom step configuration is not permitted, instead it is derived from the associated pipeline template`

This is because it skips over the plan modifier because its an existing pipeline, and then tries to set the steps to an empty string in the graphql mutation. This likely instead needs to set the steps as null or just not include them in the mutation so that the API does not try to deal with a template and defined steps

### LINKED ISSUE #507: Buildkite producing invalid terraform plan
This seems to be similar to the error seen in https://github.com/buildkite/terraform-provider-buildkite/issues/502

I've been trying to run a Terraform apply for our org and have been hitting multiple instances of this error:
```
╷
│ Error: Provider produced invalid plan
│ 
│ Provider "registry.terraform.io/buildkite/buildkite" planned an invalid value for buildkite_pipeline.website.steps: planned value cty.StringVal("steps:\n- label:
│ ':pipeline: Pipeline Upload'\n  command: buildkite-agent pipeline upload") does not match config value cty.UnknownVal(cty.String).
│ 
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
╵
```

We're on version 1.4.1 of the Buildkite terraform provider.

jradtilbrook:
## PR checklist:
- [x] `docs/` updated
- [x] tests added

**Notable changes**

- This will no longer set steps when a pipeline is being used and vice versa. Doing this simplifies the plan modification downstream

Resolves #507 
Resolves #506


buildkite/resource_pipeline.go
```diff
@@ -717,22 +717,18 @@ func (p *pipelineResource) UpgradeState(ctx context.Context) map[int64]resource.
...
 		var template, steps types.String
 
 		// Load pipeline_template_id and steps (if defined)
 		req.Plan.GetAttribute(ctx, path.Root("pipeline_template_id"), &template)
 		req.Plan.GetAttribute(ctx, path.Root("steps"), &steps)
 
-		// Set default steps only if there is no template oe defined steps
+		// Set default steps only if there is no template or defined steps
 		if template.IsNull() && (steps.IsUnknown() || steps.IsNull()) {
 			resp.Diagnostics.Append(resp.Plan.SetAttribute(ctx, path.Root("steps"), defaultSteps)...)
```
tomowatt, line 730:
```suggestion
if template.IsNull() && steps.IsNull() {
			resp.Diagnostics.Append(resp.Plan.SetAttribute(ctx, path.Root("steps"), defaultSteps)...)
```

Wondering if it makes sense to drop `steps.IsUnknown()` from this condition, to allow for when the value of `steps` is still be be determined, as we've got a case where an user has the value of `steps` be created elsewhere and is subject to change. As currently, it would set the 'default steps' in-place and result in a difference in the planned value and error out.

jradtilbrook: Is this off the back of this [doc](https://github.com/hashicorp/terraform/blob/main/docs/resource-instance-change-lifecycle.md) specifying how this is called twice? I think you might be on the right track, although I'm wondering if the "unknown" check should be moved way up to the top - something like, if its unknown, do nothing straight away - and wait for the invocation where it is known

jradtilbrook: I completely removed the checks that were intended for "creating/updating" as it doesn't really matter.

At the end of it all, we care what the user has supplied (req.Config) and whether we are in "template" or "steps" mode (explicit or implicit).

I've updated the plan modifier and created a few new tests for each case. I think it covers it all. And they are all passing!! Have a look over and see if there are any branches I haven't covered in the tests

tomowatt (approved):
LGTM

I've ran a few manual tests, going back, forth and inbetween:
* explicity defined Steps
* implicit default Steps
* explicity defined Pipeline Template

And all works as expected, with the bonus of handling unknown Step values at the time of Plan/known at Apply

Issue #533, opened 2024-05-31
Seg fault when cluster queue is not found


patrobinson:
**Describe the bug**
I didn't capture the stack trace sorry, but I experienced a seg fault when doing a `terraform apply` that pointed to this location:

https://github.com/buildkite/terraform-provider-buildkite/blob/main/buildkite/resource_cluster_queue.go#L196

Upon reading the code, it seems impossible for `err` to be anything but nil, so anytime this code path is reached it will seg fault.

**To Reproduce**
I guess try and apply a stack that reads a non-existant cluster queue

**Expected behavior**
the provider doesn't seg fault.

PR #546, opened 2024-06-19
Fix nil reference error when err may be nil

mcncl:
Fixes https://github.com/buildkite/terraform-provider-buildkite/issues/533


buildkite/resource_cluster_queue.go
```diff
@@ -193,7 +193,7 @@ func (cq *clusterQueueResource) Read(ctx context.Context, req resource.ReadReque
 	// This is a tradeoff of the current getClusterQueues Genqlient query (searches for 50 queues via the cluster UUID in state)
 	resp.Diagnostics.AddError(
 		"Unable to find Cluster Queue",
-		fmt.Sprintf("Unable to find Cluster Queue: %s", err.Error()),
+		"Unable to find Cluster Queue",
```
jradtilbrook, line 196: Should probably use the uuid from state or something more helpful

jradtilbrook (approved):

PR #562, opened 2024-09-05
SUP-2506: Organization rule resource and data source implementation

james2791:
Feature(Organization rules resource/data source): Implemented both a resource and data source (`buildkiite_organization_rule`) for interaction with [Rules](https://buildkite.com/docs/pipelines/rules). As of typing, both types (`TRIGGER_BUIILD` and `ARTIFACTS_READ`) are supported - and IDed lookups for both's read takes in the respective node query found in `node.graphql`, where said Fragment is within `organization_rule.graphql`.

For creation of `buildkiite_organization_rule`s, both types support optional [conditions](https://buildkite.com/docs/pipelines/rules) (see the `Conditions` heading from those docs) within the rule's JSON document, along with an optional `description` field at the rule level. 

## Test Coverage

Test variations implement the successful cases for creating, deleting and importing both rule types via the `buildkite_organization_rule` resources (all tests invoke a deletion)
- Creating a `buildkite_organization_rule` (`TRIGGER_BUIILD` action) with required attributes
- Creating a `buildkite_organization_rule` (`TRIGGER_BUIILD` action) with all attributes
- Creating a `buildkite_organization_rule` (`ARTIFACTS_READ` action) with required attributes
- Creating a `buildkite_organization_rule` (`ARTIFACTS_READ` action) with all attributes
- Importing a `buildkite_organization_rule` (`TRIGGER_BUIILD` action) with required/all attributes
- Importing a `buildkite_organization_rule` (`ARTIFACTS_READ` action) with required/all attributes

For data sources through `data.buildkite_organization_rule`:
- Loading a `buildkite_organization_rule` (`TRIGGER_BUIILD` action) that has required attributes set into state
- Loading a `buildkite_organization_rule` (`TRIGGER_BUIILD` action) that has all attributes set into state
- Loading a `buildkite_organization_rule` (`ARTIFACTS_READ` action) that has required attributes set into state
- Loading a `buildkite_organization_rule` (`ARTIFACTS_READ` action) with has all attributes into state

Additionally various corner-cases that can occur when definining a rule's JSON `value` document:
- Invalid action specified in the `buildkite_organization_rule`'s `type` attribute.
- When an invalid conditional is specified in a `buildkite_organization_rule`'s `value` JSON document, specifically in the `conditions` optional array.
- When no `source_pipeline_uuid` attribute is specified in a `buildkite_organization_rule` resource's `value` JSON document, and a `target_pipeline_uuid` attribute is.
- When no `target_pipeline_uuid` attribute is specified in a `buildkite_organization_rule` resource's `value` JSON document, and a `source_pipeline_uuid` attribute is.
- When the `source_pipeline_uuid` attribute is excluded from a `buildkite_organization_rule` resource's `value` JSON document.
- When the `target_pipeline_uuid` attribute is excluded from a `buildkite_organization_rule` resource's `value` JSON document.
- When the `uuid` specified in the `source_pipeline_uuid` attribute doesn't exist in the organization.
- When the `uuid` specified in the `target_pipeline_uuid` attribute doesn't exist in the organization.

## PR checklist:
- [x] `docs/` updated **(These docs for rules are templated with a warning message on opting-in for using Rules before using both Organization rule resources/data sources.)**
- [x] tests added


buildkite/resource_organization_rule.go
```diff
@@ -0,0 +1,370 @@
...
+			"target_uuid": resource_schema.StringAttribute{
+				Computed:            true,
+				MarkdownDescription: "The UUID of the target resource that this organization rule allows or denies invocation its respective action. ",
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
+			"effect": resource_schema.StringAttribute{
+				Computed:            true,
+				MarkdownDescription: "Whether this organization rule allows or denys the action to take place between source and target resources. ",
```
mcncl, line 113: *denies

buildkite/resource_organization_rule.go
```diff
@@ -0,0 +1,370 @@
...
+		Otherwise, it will create and throw an error stating that it cannot obtain the source type from the returned API response.
+	*/
+	switch r.RuleCreate.Rule.SourceType {
+	case "PIPELINE":
+		if ruleCreateSourcePipeline, ok := r.RuleCreate.Rule.Source.(*OrganizationRuleFieldsSourcePipeline); ok {
+			sourceUUID = ruleCreateSourcePipeline.Uuid
+		} else {
+			return nil, nil, errors.New("Error obtaining source type upon creating the organization rule.")
+		}
+	}
```
mcncl, line 302: Without a default block, the code will continue to the next section. This means that if the case is NOT `PIPELINE` it will continue to the next block, which also checks if it's `PIPELINE`, which will then continue on and try to return a `&sourceUUID` and `&targetUUID` which I _think_ will cause a nil pointer error.

buildkite/resource_organization_rule.go
```diff
@@ -0,0 +1,370 @@
...
+	case "PIPELINE":
+		sourceUUID = nr.Source.(*OrganizationRuleFieldsSourcePipeline).Uuid
+	}
+
+	switch nr.TargetType {
+	case "PIPELINE":
+		targetUUID = nr.Target.(*OrganizationRuleFieldsTargetPipeline).Uuid
+	}
+
+	return sourceUUID, targetUUID
```
mcncl, line 334: This should be fine as it will essentially return `""`

docs/resources/organization_rule.md
```diff
@@ -0,0 +1,108 @@
...
+#             }            
+#           }
+#           action
+#         }
+#       }
+#     }
+#   }
+# }
+
+terraform import buildkite_organization_rule.artifact_read UnVsZS0tLTAxOTE5NmU2LWNiNjctNzZiZi1iYzAyLTVhYzFiNzhhMWMyOA==
```
mcncl, line 107: Would we not consider other methods of import too? UUID, slug of rule?

examples/resources/buildkite_organization_rule/import.sh
```diff
@@ -0,0 +1,47 @@
...
+#             }            
+#           }
+#           action
+#         }
+#       }
+#     }
+#   }
+# }
+
+terraform import buildkite_organization_rule.artifact_read UnVsZS0tLTAxOTE5NmU2LWNiNjctNzZiZi1iYzAyLTVhYzFiNzhhMWMyOA==
```
mcncl, line 47: Same here re: methods of import

mcncl (commented):
Added some thoughts

buildkite/resource_organization_rule.go
james2791: I really enjoy spelling, altered, cheers

buildkite/resource_organization_rule.go
james2791: I've added a `default` to both thees switch statements to catch when GraphQL doesn't return the type for TF to store in state, and thus will error 👍

templates/data-sources/organization_rule.md
```diff
@@ -0,0 +1,42 @@
...
+page_title: "buildkite_organization_rule Data Source - terraform-provider-buildkite"
+subcategory: ""
+description: |-
+  Use this data source to retrieve an organization rule by its ID.
+  More information on organization rules can be found in the documentation https://buildkite.com/docs/pipelines/rules/overview.
+---
+
+# buildkite_organization_rule (Data Source)
+
+~> Rules is a feature that is currently in development and enabled on an opt-in basis for early access. To have this enabled for your organization for utilizing this data source, please reach out to Buildkite's [Support Team](https://buildkite.com/support).
```
james2791, line 12: This is to align with what is currently shown on the [docs](https://buildkite.com/docs/pipelines/rules) - and a pre-requisite before folks can use this data source for their organization.

templates/resources/organization_rule.md
```diff
@@ -0,0 +1,108 @@
...
+page_title: "buildkite_organization_rule Resource - terraform-provider-buildkite"
+subcategory: ""
+description: |-
+  An Organization Rule allows specifying explicit rules between two Buildkite resources and the desired effect/action.
+  More information on organization rules can be found in the documentation https://buildkite.com/docs/pipelines/rules/overview.
+---
+
+# buildkite_organization_rule (Resource)
+
+~> Rules is a feature that is currently in development and enabled on an opt-in basis for early access. To have this enabled for your organization for utilizing this resource, please reach out to Buildkite's [Support Team](https://buildkite.com/support).
```
james2791, line 12: This is to align with what is currently shown on the [docs](https://buildkite.com/docs/pipelines/rules) - and a pre-requisite before folks can use this resource for their organization.

docs/resources/organization_rule.md
james2791: I think thats a good option - at the least if one knows a `uuid` of a rule, in conjuction to the the regular `id` way that the import passthrough works as of now. Something else I'll start looking at 👍

examples/resources/buildkite_organization_rule/import.sh
james2791: Same with `uuid`s - I think for data sources too this'll go well, similar to pipeline templates, though will need a more extensive lookup

buildkite/resource_organization_rule.go
james2791: 👍👍

mcncl (approved):

PR #563, opened 2024-09-13
Remove Jarryd from CODEOWNERS

mcncl:
OK, this one is actually kind of sad.

Thanks to @jradtilbrook for everything that he's done with this provider. We'll care for it with love 💚


jradtilbrook (approved):
:saluting_face:

tomowatt (approved):

Issue #509, opened 2024-03-24
Feature request: support pausing/resuming dispatch on cluster queues


james2791:
As of writing - the cluster queue resources support maintaining a queue through its various CRUD operations - where users are able to maintain its attributes such as its `key`, `description` and the `cluster_id` of the cluster that it is a part of.

Managing the paused state of a queue (the `dispatchPaused` attribute on a Cluster Queue GraphQL [object](https://buildkite.com/docs/apis/graphql/schemas/object/clusterqueue)) will be handy to have - as users would be able to programatically pause/resume their queues statefully with respective plan/applies. The cluster queue resource will need to make the respective [pause](https://buildkite.com/docs/apis/graphql/schemas/mutation/clusterqueuepausedispatch)/[resume](https://buildkite.com/docs/apis/graphql/schemas/mutation/clusterqueueresumedispatch) mutations if said attribute was added in - taking in the queues `id` as input.

If a user also tried to remove a queue that was paused - perhaps the provider could also stop this from happening in the case that the queue is legitimately paused. For example, the queue is attached to a agent fleet where no scaling is desired over the weekend, or perhaps scheduled downtimes on respective agent fleets are occurring during a scheduled upgrade etc.

mcncl:
Completed in https://github.com/buildkite/terraform-provider-buildkite/pull/573

PR #573, opened 2024-10-16
Support queue dispatch pause

mcncl:
feat: Support queue dispatch pausing

## PR checklist:
- [x] `docs/` updated
- [x] tests added

Addresses https://github.com/buildkite/terraform-provider-buildkite/issues/509


buildkite/resource_cluster_queue_test.go
```diff
@@ -113,7 +113,9 @@ func TestAccBuildkiteClusterQueueResource(t *testing.T) {
 				},
 				{
 					Config: configBasic(clusterName, queueKey, updatedQueueDesc),
-					Check:  ckecUpdated,
+					Check:  checkUpdated,
+					// Destroying a queue will pause its dispatch
+					ExpectNonEmptyPlan: true,
```
mcncl, line 118: So, this is kind of needed as the tear down of queues has an implicit pausing on the queue, thus setting `dispatch_paused` to `true` on delete

mcncl: This means the plan will not be empty, because the tear down causes the `false` to become `true`

tomowatt (approved):
:shipit:

PR #575, opened 2024-11-14
SUP-2709: clean go modules

tomowatt:
```
go mod verify
go mod tidy
go mod verify
```
Cleaning up unused modules for dependabot alerts


mcncl (approved):
🧹

PR #576, opened 2024-11-14
SUP-1918: Verify document generation is up-to-date

petetomasik:
* Pipeline will run `make docs` and exit `1` if `git diff` returns any modified files under `docs/` post-generation.
* Various tweaks to make `tfplugindocs` run happily inside `go` container
* Ran `terraform fmt` to cleanup `*.tf` files
* Ran `make docs`
* Go modules cleanup
* Fixed index digest hash for `hashicorp/terraform:1.4` for multi-platform (old SHA was incorrectly `linux_386`)


.buildkite/steps/documentation.sh
```diff
@@ -0,0 +1,19 @@
...
+make docs
+
+echo "--- :git: Checking for changes"
+git diff --exit-code docs &>/dev/null && true
+
+docs_changes="$?"
+
+if [ "${docs_changes:-0}" -ne 0 ] ; then
+	echo "+++ :bk-status-failed: Documentation changes detected!!!"
+	git status --short
```
mcncl, line 16: I wonder should we output this as an annotation on the build to make it more front and centre? Non-blocking, just a general thought.

mcncl (dismissed):

.buildkite/steps/documentation.sh
tomowatt: Yeah thinking adding this as an annotation would make it better over all as things can be missed when within Job Logs

petetomasik:
> @petetomasik Not 100% sure what's happening with the linter Job failing, might be due to the conflicts though running the linter locally is not returning the same results as the Job 🤔

This was seemingly caused by the inclusion of `toolchain` [here](https://github.com/buildkite/terraform-provider-buildkite/pull/576/commits/a925acdef033ae637e861c29ca2f37b916269597). Removed and linting once more.

tomowatt (approved):

PR #588, opened 2024-12-03
SUP-2388 - Update docs with templates, import blocks

petetomasik:
* Switch to templates
* Include both `terraform import` and `import` block syntax for resource imports
* Create template overrides for:
  * `buildkite_organization`
  * `buildkite_organization_rule`


tomowatt (approved):
🚀

Issue #527, opened 2024-05-14
feature request: Add datasources to get the list of organization members


christophe-scalepad:
Hey,

Hope you are doing well.

## Description

As a company, we would like to manage RBAC in buildkite (and everywhere else) as IaC using terraform. We are currently missing the possibility to fetch the members of an organization. This would allow us to map our's idp's groups to teams we create  in buildkite. Currently with what is available to us, our only option would be to hardcode the user ids in our IaC which we would like to avoid.

This feature is already present in [another buildkite provider](https://registry.terraform.io/providers/tink-ab/buildkite/latest/docs/data-sources/org_members). This would allow us to fetch the user ids dynamically from buildkite using the data source and create a dynamic map with the users that we got from our idp's group.

I haven't looked in the other providers implementation, but I'm guessing it relies on your API's as well.

wishing you well

mcncl:
Hey there @christophe-scalepad!

Thanks for raising this issue. Sounds like a useful thing we cold add in and shouldn't be too difficult to do; the queries are all available in the API as far as I can tell. We'll plan the work for this, but I can't see it being more that a few days to add in once it's picked up.


Cheers!

ianchesal:
Here to +1 this -- this is the missing link between teams and organizing them in Terraform at present. Fetching the UUID for users to add them to teams currently O(hard) to do in any way that isn't incredible fragile.

petetomasik:
Data sources for `organization_members` and `organization_member` have been implemented in https://github.com/buildkite/terraform-provider-buildkite/pull/590

PR #590, opened 2024-12-10
SUP-2214 - Create data sources for `organization_members` and `organization_member`

petetomasik:
Create new data sources:
#### `buildkite_organization_members`
Takes in no required arguments and returns a `members` object containing all Organization members:
```
data "buildkite_organization_members" "organization" {}
```
#### `buildkite_organization_member`
Requires `email` argument and returns the `id`, `uuid`, `name` and `email` of a matching user. Emits error message of `No organization member found` when no match found in Organization.
```
data "buildkite_organization_member" "a_smith" {
  email = "a.smith@company.com"
}

resource "buildkite_team" "developers" {
  name                = "Developers"
  privacy             = "VISIBLE"
  default_team        = false
}

resource "buildkite_team_member" "developers_a_smith" {
  team_id = buildkite_team.developers.id
  user_id = data.buildkite_organization_member.a_smith.id
  role    = "MEMBER"
}
```
Addresses: https://github.com/buildkite/terraform-provider-buildkite/issues/527


buildkite/graphql/organization_members.graphql
```diff
@@ -0,0 +1,24 @@
+query GetOrganizationMembers(
+    $slug: ID!,
+    # @genqlient(pointer: true)
+    $cursor: String
+) {
+    organization(slug: $slug) {
+        members(first: 100, after: $cursor) {
```
mcncl, line 7: The limit on a GQL query is 500, should we increase to that?

mcncl (changes requested):
Looks good, perhaps increase the return count of the query? If an org has less than 500 then it will be fine and won't need the cursor, but we can probably set the max here.

buildkite/graphql/organization_members.graphql
petetomasik: Increased results to `500` 👍

mcncl (approved):

Issue #591, opened 2024-12-11
When changing a description of the default queue of a cluster, the default queue of that cluster is dropped 


ntascii:
**Describe the bug**
When changing a description of a cluster's default queue, the cluster will no longer have its assigned default queue.
We have a cluster managed in terraform. The usage can be simplified like:
```
resource "buildkite_cluster_queue" "test_queue" {
  key         = "test_queue"
  description = "test description"
  cluster_id  = module.test_cluster.cluster_id
}

resource "buildkite_cluster_default_queue" "default" {
  cluster_id = module.test_cluster.cluster_id
  queue_id   = buildkite_cluster_queue.test_queue
}
```

**To Reproduce**
- Change description of the default queue 
- terraform apply 
- The description is updated, the state file still contains resource buildkite_cluster_default_queue however, there is no longer a default queue for the cluster 
- Hit error `No default queue is assigned to the cluster` whenever doing a `terraform plan` afterwards. 
- 
**Expected behavior**
Default queue should not be dropped when changing it's description. 

**Additional context**

tomowatt:
@ntascii 

We've found that there is an issue with the GraphQL API itself that results Default Queue being removed when the Queue is updated until this is fixed we recommend adding the `replace_triggered_by` lifecycle onto the `buildkite_cluster_default_queue` resources e.g.,

```
resource "buildkite_cluster" "local" {
  name        = "TF cluster"
  description = "Cluster for TF projects"
}

resource "buildkite_cluster_queue" "test_queue" {
  cluster_id  = buildkite_cluster.local.id
  key         = "test_queue"
  dispatch_paused = "true"
}

resource "buildkite_cluster_default_queue" "test_default_queue" {
  cluster_id = buildkite_cluster.local.id
  queue_id   = buildkite_cluster_queue.test_queue.id

  lifecycle {
    replace_triggered_by = [  buildkite_cluster_queue.test_queue ]
  }
}
```

the PR linked above includes other fixes for Cluster Queues.

tomowatt:
Hey @ntascii 
Thanks for your patience, following up as the Product Team have fixed the issue with GraphQL API and the workaround with the `lifecycle` is no longer required now so the Default Queue will stay as is when the Queu is updated.

PR #593, opened 2024-12-12
SUP-2813: correct logic for pausing/resuming Dispatch on Queues and remove error with Default Cluster Queues

tomowatt:
* removed `isPausible` as it would trigger Pausing when the Description was updated and State `DispatchedPaused = false`
* compare Plan and State values to pause/resume dispatch else do nothing, which stops changes to the `Description` causing the Queue to be paused
* `r.ClusterQueueUpdate.ClusterQueue.DispatchPaused` is nil at the time of the original logic resulting in `panic: runtime error: invalid memory address or nil pointer dereference` so changed to the Plan Value to be used for clarity else could be hardcoded as `state.DispatchPaused = types.BoolValue(true)` or `state.DispatchPaused = types.BoolValue(false)` (where appropritate)
* Remove error raised when a Cluster Queue doesn't have a Default Queue defined as there is an error with the API when updating a Queue resulting it in being removed as Default. But Clusters do not require a Default Queue being set so error not required.
* Add Warning as Queues cannot be created with `dispatch_paused = true` due to limitation with API.

Includes fixes for https://github.com/buildkite/terraform-provider-buildkite/issues/591 though there is a API issue resulting in the changes of Default Queues during updates to the Cluster Queue

### DIFF
--- a/buildkite/resource_cluster_queue.go
+++ b/buildkite/resource_cluster_queue.go
@@ -226,16 +226,10 @@ func (cq *clusterQueueResource) ImportState(ctx context.Context, req resource.Im
 
 func (cq *clusterQueueResource) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) {
 	var plan, state clusterQueueResourceModel
-	var description types.String
 	var planDispatchPaused, stateDispatchPaused bool
 
-	diagsState := req.State.Get(ctx, &state)
-	diagsDescription := req.Plan.GetAttribute(ctx, path.Root("description"), &description)
-
-	// Load state and obtain description from plan (singularly)
-	resp.Diagnostics.Append(diagsState...)
+	resp.Diagnostics.Append(req.State.Get(ctx, &state)...)
 	resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)
-	resp.Diagnostics.Append(diagsDescription...)
 
 	if resp.Diagnostics.HasError() {
 		return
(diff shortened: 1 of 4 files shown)


lizrabuya (dismissed):

buildkite/resource_cluster_queue.go
```diff
@@ -146,8 +146,15 @@ func (cq *clusterQueueResource) Create(ctx context.Context, req resource.CreateR
...
-	state.DispatchPaused = plan.DispatchPaused
 
+	// GraphQL API does not allow Cluster Queue to be created with Dispatch Paused
+	if plan.DispatchPaused.ValueBool() {
+		resp.Diagnostics.AddWarning(
+			"Cluster Queue cannot be created with Dispatch Paused",
+			"Setting dispatch_paused in State to 'true' but requires re-run of Apply",
+		)
+	}
+	state.DispatchPaused = plan.DispatchPaused
```
petetomasik, line 157: To better match the current functionality of the API, I think we should set `state.DispatchPaused = types.BoolValue(true)` here and (modify, but) retain the warning. We should also modify the wording in the Schema  for `dispatch_paused` to state that on resource creation this attribute will be set to `false`.

Alternatively, we can move the logic that calls `pauseDispatchClusterQueue()` and `resumeDispatchClusterQueue()` into its own `func` and call the function from both `Create()` and `Update()`. This would ensure that the "expected" state of a `cluster_queue` resource is correct after each (singular) TF run.

petetomasik: Actually, no. Setting `state.DispatchPaused = types.BoolValue(false)` will break Terraform with a `... dispatch_paused: was cty.True, but now cty.False` error if `dispatch_paused = true` on resource creation.

mcncl (dismissed):

mcncl (approved):

PR #596, opened 2024-12-19
Adding cluster queue support for hosted agents

CerealBoy:
Adding support for managing hosted agent cluster queues.

There are a couple of nuances that will go along with this capability (to begin with), so we will also capture those within the docs for any hints on their usage.

I have included a small QoL change to the `Makefile` that helped greatly with the local development. This _should_ be compatible with the original functionality provided by the command. There may be some additional pointers added to the `README.md` to reflect this change and other pointers that I encountered along the way.

## TODO

- [x] `go test ./...` passes
- [ ] `make testacc` passes
- [ ] Tidy documentation on new functionality
- [ ] Ensure validation covers the common expected paths


buildkite/resource_cluster_queue.go
```diff
@@ -95,6 +129,52 @@ func (clusterQueueResource) Schema(ctx context.Context, req resource.SchemaReque
...
+							"xcode_version": resource_schema.StringAttribute{
+								Optional: true,
+								Computed: true,
+							},
+						},
+					},
+					"linux": resource_schema.SingleNestedAttribute{
+						Optional: true,
+						Attributes: map[string]resource_schema.Attribute{
+							"image_agent_ref": resource_schema.StringAttribute{
```
matthewborden, line 151: The naming of this field in the API is `agent_image_ref`, we should be consistent with the naming if we're doing something similar.

CerealBoy: I've updated this attribute now to keep the naming consistent

mcncl (approved):

mcncl (approved):

mcncl (approved):

PR #598, opened 2024-12-20
SUP-2591: add PlanModifiers when Hosted Agent attributes changed

tomowatt:
* Queues cannot change from Self-Hosted to Hosted (and vice-versa) so Resource should be recreated
* instance_shape cannot be updated in-place so Resource should be recreated to change it


petetomasik (approved):
Looks good! Our custom `PlanModifiers` can be found under `internal/planmodifier/` and used like [this](https://github.com/buildkite/terraform-provider-buildkite/blob/main/buildkite/resource_pipeline.go#L541). We can take a small action item for this in the name of getting this fix merged in. 👍

Issue #586, opened 2024-11-29
Decouple pipeline slug from pipeline name


AliSoftware:
### Is your feature request related to a problem? Please describe.

The `slug` of a pipeline is derived from its name—`name` attribute in terraform.

This means that if you decide to change the display name of an existing pipeline at some point, this will also changes its `slug`. As a result:
 - The URL of the pipeline page on the web, so any existing reference to the old URL (e.g. in our team's doc) would have to be udated
 - Since the `slug` is used as an identifier for the pipeline in many context, all of those will need to be updated as well (e.g. scripts that use the API that needs the slug to identify the pipeline, GraphQL queries, etc)
 - In particular, the [elastic-ci-stack-s3-secrets-hooks](https://github.com/buildkite/elastic-ci-stack-s3-secrets-hooks) plugin uses the `slug` of the pipeline to fetch the proper secrets (SSH clone keys, env vars…) for a pipeline in S3, which means changing the name of a pipeline will break it and make it stop finding those secrets and thus stop working just because the `slug` changed under us.

### Describe the solution you'd like

It would be nice if the `slug` would only be derived from the `name` if not provided explicitly in the `terraform` resource, but would otherwise be read-write.

So in particular, creating the pipeline resource via this code would make the `slug` of the created pipeline be derived to `my-project`:

```terraform
resource "buildkite_pipeline" "my-project-pipeline" {
  name = "My Project"
  # no `slug = …` attribute provided explicitly, so it will be derived from `name` to be `my-project` implicitly
…
}
```

And changing the `name` terraform attribute later to `name = "My Awesome Project"` would update the `slug` to `my-awesome-project`. So far, just like today, no change there.

But specifying the `slug` a an explicit attribute in the terraform resource would allow us to override its derived value, so:
```terraform
resource "buildkite_pipeline" "my-project-pipeline" {
  name = "My Awesome Project"
  slug = "my-project"
…
}
```
would allow us to still reference that pipeline with the slug `my-project`, even if the human-readable display name used in the Buildkite web portal etc. would be "My Awesome Project".

This would allow to give the option to decouple the display name of a pipeline from its `slug`, all while keeping the current behavior of having the slug derived by default if not provided explicitly.

### Describe alternatives you've considered
N/A

### Additional context

This is a follow-up of this thread in the support forums: https://forum.buildkite.community/t/decouple-pipeline-name-from-logical-reference-slug/1414/6 cc @benmc

### Potential impact

[...]

PR #603, opened 2024-12-23
SUP-2777 - Allow configurable pipeline slugs

petetomasik:
Changes `slug` from a read-only attribute to an optional configurable attribute. If not set, `slug` is derived from the pipeline name and returned from the API. The `slug` will be updated when the `name` of the pipeline changes. If the `slug` is defined, a second REST API call is made to update the `slug` for the pipeline. The `slug` can be set or excluded at any time in the pipeline's lifecycle and the value will be set based on the user-defined input or via the API's response.

Utilizes Terraform [private state](https://developer.hashicorp.com/terraform/plugin/framework/resources/private-state) to track whether `slug` value in State was set via `api` or `user`. This extra value is required to correctly evaluate the transition of removing user-defined `slug` from an existing resource and re-trigger storing an updated `slug` from the API.

Addresses: https://github.com/buildkite/terraform-provider-buildkite/issues/586

### DIFF
--- a/buildkite/resource_pipeline.go
+++ b/buildkite/resource_pipeline.go
@@ -1019,10 +1019,11 @@ func setPipelineModel(model *pipelineResourceModel, data pipelineResponse) {
 	model.Tags = tags
 }
 
-// As of May 21, 2021, GraphQL Pipeline is lacking support for the following properties:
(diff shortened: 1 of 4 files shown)


mcncl (approved):

rajatvig:
This change has possibly caused a regression on our end as we use the `pipeline.slug` to pass to another TF resource which now does not have the said resource available.

Issue #606, opened 2025-01-17
Pipeline Slug cannot be used without explicitly setting it


rajatvig:
**Describe the bug**
Post the change in https://github.com/buildkite/terraform-provider-buildkite/pull/603 it seems we can no longer use the `slug` in another terraform resource.

**To Reproduce**
Create a new Terraform Resource using the `buildkite_pipeline` resource and then use the `pipeline.slug` in another resource and it fails.

**Expected behavior**
It should be set as the output of the pipeline resource.

petetomasik:
Hi @rajatvig! We're having issues re-producing this bug. With the change you referenced, the `slug` attribute was made configurable. It was previously a read-only attribute, but has existed as a `pipeline` attribute for some time.

Would you be able to provide us with the previous version of the provider you were using prior to updating to version `>=1.15.3`? Could you also provide us with a snippet of the Terraform code you are using to create the `buildkite_pipeline` resource and how you are referencing the `slug` attribute of this pipeline in another resource? What is the specific error that you are encountering? Thanks!

rajatvig:
What we roughly have is

```terraform
resource "buildkite_pipeline" "foo" {
  name           = "foo"
  description    = "Foo Pipeline"
  repository     = "git@github.com:org/foo.git"
  default_branch = "main"

  steps = file("pipelines/dynamic.yml")

  provider_settings = {
    build_pull_requests            = true
    build_branches                 = true
    publish_commit_status          = true
    publish_commit_status_per_step = true
    separate_pull_request_statuses = true
    trigger_mode                   = "code"
  }
}

module "foo-bindings" {
  source = "./modules/bindings"
  repository = buildkite_pipeline.foo.repository
  pipeline   = buildkite_pipeline.foo.slug
}
```

When the module `foo-bindings` uses the `pipeline` variable in string interpolation, it gets null.
The same code pre 1.15.4 worked.

petetomasik:
Hi @rajatvig - Thanks for providing this information. I was able to reproduce this issue of the `slug` attribute being set to `null`. There was already a [PR](https://github.com/buildkite/terraform-provider-buildkite/pull/607) created earlier today that was cleaning up some of the logic around the `slug` attribute that fixes the issue you are running into. We will work on getting this PR merged in and version `1.15.5` released shortly. Thanks again for reporting this issue!

petetomasik:
[...]

Issue #601, opened 2024-12-23
buildkite_pipeline provider settings docs appear to be incorrect concerning some default values


erikkrieg:
**Describe the bug**
I was trying out Buildkite for the first time and setting it up using the official Terraform provider. I got a cluster, agent and pipeline setup with access to my GitHub repo but was having trouble getting jobs to actually trigger on my pull request (only had success triggering new builds manually from the Buildkite web app).

When reading over [settings in the TF docs](https://registry.terraform.io/providers/buildkite/buildkite/latest/docs/resources/pipeline#nested-schema-for-provider_settings), I saw:
> - [build_branches](https://registry.terraform.io/providers/buildkite/buildkite/latest/docs/resources/pipeline#build_branches-6) (Boolean) Whether to create builds when branches are pushed. Defaults to true when trigger_mode is set to code.
> - [build_pull_requests](https://registry.terraform.io/providers/buildkite/buildkite/latest/docs/resources/pipeline#build_pull_requests-6) (Boolean) Whether to create builds for commits that are part of a pull request. Defaults to true when trigger_mode is set to code.

Based on this, I was under the impression that all I needed for my provider settings to enable my desired pipeline triggers was:
```hcl
  provider_settings = {
    trigger_mode = "code"
  }
```

After debugging for some time I discovering in the UI what the actual values were. The "trigger builds after pushing code" section was enabled, but no sub-options, like "build pull requests" were enabled.

So it appears that the `buildkite_pipeline` resource is actually setting these options to `false` as the default.

**To Reproduce**
1. Create a `buildkite_pipeline` with `provider_settings.trigger_mode = "code"`, omitting other options
2. Apply... (I did go through several iterations of apply resources while learning about the available settings)
3. View terraform state for the created pipeline (i.e. `terraform state show buildkite_pipeline.example`)

```
   provider_settings                        = {
        build_branches                                = false
        build_pull_request_forks                      = false
        build_pull_request_labels_changed             = false
        build_pull_request_ready_for_review           = false
        build_pull_requests                           = false
        build_tags                                    = false
        cancel_deleted_branch_builds                  = true
        filter_condition                              = ""
        filter_enabled                                = false
        prefix_pull_request_fork_branch_names         = false
        publish_blocked_as_pending                    = false
        publish_commit_status                         = false
        publish_commit_status_per_step                = false
        pull_request_branch_filter_configuration      = ""
        pull_request_branch_filter_enabled            = false
[...]

PR #607, opened 2025-01-17
Implement default `provider_settings`, adjust `slug` attribute logic

petetomasik:
* All of the attributes under `provider_settings` now default to individually documented values.
  * Addresses https://github.com/buildkite/terraform-provider-buildkite/issues/601
* Additional tweaks for determining `slug` source as api- vs. user-defined to prevent requiring extra `terraform apply` commands to bring state into parity
  * Addresses https://github.com/buildkite/terraform-provider-buildkite/issues/606


buildkite/resource_pipeline.go
```diff
@@ -628,102 +664,121 @@ func (*pipelineResource) Schema(ctx context.Context, req resource.SchemaRequest,
 							"`fork` will create builds when the GitHub repository is forked.",
 							"`none` will not create any builds based on GitHub activity.",
 							"`trigger_mode`",
-							"`code`",
+							"`none`",
```
tomowatt, line 667: Believe this shouldn't be changed as there is `none` defined two lines above and `code` is valid

petetomasik: This is changing the substitution in this [line](https://github.com/buildkite/terraform-provider-buildkite/blob/9d0e4b3363328ecfcc570b1d937aac5e96b0edc9/buildkite/resource_pipeline.go#L660) that describes what the default `trigger_mode` is in the description.

tomowatt: gotcha!
reading without the fully expanded code above 🤦

tomowatt (approved):
🚀

Issue #866, opened 2025-01-30
Provider fails to remove base image from hosted cluster queue and produces inconsistent state.


matthewborden:
When removing the linux stanza (or agent_image_ref) from a `buildkite_cluster_queue`, `terraform apply` produces an inconsistent result:

```
 Error: Provider produced inconsistent result after apply
│
│ When applying changes to buildkite_cluster_queue.hosted_large, provider "provider[\"registry.terraform.io/buildkite/buildkite\"]" produced an unexpected new value: .hosted_agents.linux: was
│ null, but now cty.ObjectVal(map[string]cty.Value{"agent_image_ref":cty.StringVal("docker.io/node:latest")}).
│
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
╵
```

Terraform and provider version:
```
🕙 [ 12:57:53 ] ❯ terraform --version
Terraform v1.8.2
on darwin_arm64
+ provider registry.terraform.io/buildkite/buildkite v1.15.5
```

To reproduce the inconsistent result apply:

For Buildkite folks, I've setup a reproduction in https://github.com/buildkite/terrakite/commit/b5512bdf61c140881c4f0c19c706e5e0c9f25681, you'll need to setup an org to apply it to. 

Create a hosted queue with the following and set a custom base image (note: this required this feature to enabled by support) with:
```
resource "buildkite_cluster" "hosted" {
  name = "hosted"
}
resource "buildkite_cluster_queue" "hosted_large" {
  cluster_id = buildkite_cluster.hosted.id
  key       = "hosted_large"
  hosted_agents = {
    instance_shape = "LINUX_AMD64_4X16"
    linux = {
      agent_image_ref = "docker.io/node:latest"
    }
  }
}
```

Remove the linux stanza from the `buildkite_cluster_queue.hosted_large` block:
```
resource "buildkite_cluster_queue" "hosted_large" {
  cluster_id = buildkite_cluster.hosted.id
  key       = "hosted_large"
  hosted_agents = {
    instance_shape = "LINUX_AMD64_4X16"
  }
}
```

This should generate a diff of the following:

Terraform will perform the following actions:
  # buildkite_cluster_queue.hosted_large will be updated in-place
  ~ resource "buildkite_cluster_queue" "hosted_large" {
      ~ hosted_agents   = {
          - linux          = {
              - agent_image_ref = "docker.io/node:latest" -> null
            } -> null
            # (1 unchanged attribute hidden)
        }
[...]

PR #872, opened 2025-02-04
SUP-3317 - Fix hosted agent attributes state mismatch

petetomasik:
Fixes https://github.com/buildkite/terraform-provider-buildkite/issues/866
Also ensures that the value of `XcodeVersion` stored in state is sourced from the API response vs. what was provided via the plan. This was covering up another issue where the `XcodeVersion` was not actually being set via the API.


mcncl (approved):

mcncl:
Oh, looks like maybe a test for this @petetomasik ?

james2791 (dismissed):
Looks good - addition to the test mention, changelog entry add too 👍

tomowatt (approved):

PR #886, opened 2025-02-28
Simplified organization rule datasource reads via UUID

james2791:
Feat (Organization Rule Datasource): Optimised implementation of `buildkite_organization_rule` data sources's UUID read logic.

Previously, when a Terraform implementer would define a `buildkite_organization_rule` data source via lookup through its UUID - the provider would need to loop through the organizations currently defined rules to match responses, compare UUIDs, and store the corresponding found rules' data into Terraform state. 

With the introduction of the dedicated `rule` [query](https://buildkite.com/docs/apis/graphql/schemas/query/rule), organization rule data source reads by UUID utilise this for a direct lookup without the previously implemented looping strategy. 

## Test Coverage
Existing test cases implemented both ID and UUID based reads for an organization rule data source. There is no material change to this behaviour - more the obtaining of the data to store in Terraform state.

## PR checklist:
- [x] `docs/` updated <-- No need for a documentation update - provider logic alteration.
- [x] tests added <-- Tests remain consistent, no material change to definition of data sources.
- [x] an example added to `examples/` <-- Example organization rule data sources remain consistent, no change.
- [x] `CHANGELOG.md` updated with pending release information <--- Changes made


petetomasik (approved):
LGTM

PR #901, opened 2025-04-07
feat: add rate limit handling for REST API calls

scadu:
This PR implements rate limit handling for REST API calls using the `hashicorp/go-retryablehttp`. 
Includes:
  - Backoff using [RateLimit-Reset](https://buildkite.com/docs/apis/rest-api/limits#checking-rate-limit-details) header to time retry attempts
  - Retry configuration with 10 max attempts, 15-180s wait window defined now — can be exposed as configurable parameters later
  - Better error handling with response body content in error messages
  - Proper timeout context management for all requests

Also, added basic [Lefthook](https://github.com/evilmartians/lefthook) pre-commit check. 

Fixes https://github.com/buildkite/terraform-provider-buildkite/issues/477


PriyaSudip (commented):
Still reviewing but tests seem to have failed

scadu:
> Still reviewing but tests seem to have failed

@PriyaSudip ah yes, it went rouge after merging changes from main. I’ll work on it today, and remove draft status once tests are cleared.

petetomasik:
Small nit on alternating between short assignment statements and var declarations with implicit types with some of the API requests. Other than that, looks good! 👍

buildkite/client.go
```diff
@@ -53,33 +56,138 @@ func (client Client) GetOrganizationID() (*string, error) {
...
+						return max
+					}
+					return waitTime
+				}
+			}
+
+			// Fall back to Retry-After header if available
+			if retryAfter := resp.Header.Get("Retry-After"); retryAfter != "" {
+				if seconds, err := strconv.ParseInt(retryAfter, 10, 64); err == nil {
+					waitTime := time.Duration(seconds) * time.Second
```
PriyaSudip, line 126: Should we return waitTime here?

PriyaSudip, line 126: Should we return this waitTime?

PriyaSudip (commented):
Apart from the above, rest all looks good to me.

buildkite/client.go
scadu: Yes, we should! Fixed now.

scadu:
> Small nit on alternating between short assignment statements and var declarations with implicit types with some of the API requests. Other than that, looks good! 👍

@petetomasik aligned styling where possible (to not cause any scoping issues, e.g. reusing values of `ctx` .etc.)

petetomasik (approved):

PR #917, opened 2025-05-06
feat: support setting emoji on test suite

mcncl:
feat: allow setting emoji on test suite resource

## PR checklist:
- [x] `docs/` updated
- [x] tests added


buildkite/data_source_cluster_test.go
```diff
@@ -20,7 +20,7 @@ func TestAccBuildkiteClusterDatasource(t *testing.T) {
 					Config: fmt.Sprintf(`
 						provider "buildkite" {
 							timeouts = {
-								read = "0ms"
+								read = "0s"
```
mcncl, line 23: This test was failing because `ms` isn't an available time metric and so it essentially meant no timeout.

.buildkite/steps/documentation.sh
```diff
@@ -12,7 +12,7 @@ git diff --exit-code docs &>/dev/null && true
 docs_changes="$?"
 
 if [ "${docs_changes:-0}" -ne 0 ] ; then
-	echo "+++ :bk-status-failed: Documentation changes detected!!!"
+	echo "+++ :bk-status-failed: Documentation changes detected!!! Run `make docs` locally, commit the changes and re-push."
```
mcncl, line 15: I just changed this because it was a little frustrating to fix (had to read the file)

buildkite/resource_test_suite.go
```diff
@@ -17,6 +17,7 @@ import (
 type testSuiteModel struct {
 	ApiToken      types.String `tfsdk:"api_token"`
 	DefaultBranch types.String `tfsdk:"default_branch"`
+	Emoji         types.String `tfsdk:"emoji"`
```
mcncl, line 20: Even though we set it here, it isn't returned via the API so can't be in `Response`

PR #918, opened 2025-05-16
add ignore_default_branch_pull_requests to provider settings

patrobinson:
Tested locally:
image

But it gives me an error:
```
│ Error: Provider produced inconsistent result after apply
│
│ When applying changes to buildkite_pipeline.bash_pipeline, provider "provider[\"registry.terraform.io/buildkite/buildkite\"]" produced an
│ unexpected new value: .provider_settings.ignore_default_branch_pull_requests: was cty.True, but now null.
│
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

I dunno how to Terraform?


mcncl:
@patrobinson 👋 you'll need to ensure that the value is being set in `func updatePipelineResourceExtraInfo...`

You should then be able to copy over one of the other tests and make adjustments to check that your change is effective.

mcncl (approved):

PR #930, opened 2025-06-03
fix: destroy resources created during acceptance tests

scadu:


mcncl (approved):
Nice little clean up! Love the remove of the `Node` checks 🙂

PR #945, opened 2025-07-01
feat(pipeline): add retry logic for delete with active jobs

### LINKED ISSUE #504: When a group of pipelines are replaced with destroy/create, the create is skipped if any of the pipelines are running
Following up on discussions with Buildkite support

We ran a Terraform apply that resulted in attempted destroy and create to replace a group of pipelines.

Because a number of these pipelines were running, the apply errored out and failed partway through.

The behavior was:

Destroyed ALL pipelines
<-- Failed
Skipped Create ALL pipelines

We'd rather this apply did not fail, or if it fails then at least it should create all the pipelines that it has destroyed before stopping.

I'd suggest that the fatal error on running pipelines should be changed to a warning that doesn't stop the apply from recreating.

It would also make sense to issue each destroy/create as an atomic pair, one at a time, rather than destroying everything up front.

scadu:
Add retry mechanism for pipeline deletion when active jobs prevent immediate deletion.

To adjust timeout for `delete` operations, adjust `delete` attribute in `timeouts` settings for the provider – https://registry.terraform.io/providers/buildkite/buildkite/latest/docs#delete-1, example:
```HCL
provider "buildkite" {
  organization = "my-magic-org"
  timeouts = {
    delete = "10m"
  }
}
```

___
Expected output during `terraform apply` when deleting pipeline resource(s):
```
buildkite_pipeline.test: Still destroying... [id=UGlwZWxpbmUtLS0wMTk3YzVmMC1lM2QxLTQyYzItYmI5NS0xZWYzNTg4NDg3ZTE=, 08m00s elapsed]
buildkite_pipeline.test: Destruction complete after 8m8s
```

Closes #504

### DIFF
--- a/buildkite/resource_pipeline.go
+++ b/buildkite/resource_pipeline.go
@@ -329,17 +329,26 @@ func (p *pipelineResource) Delete(ctx context.Context, req resource.DeleteReques
 	err := retry.RetryContext(ctx, timeout, func() *retry.RetryError {
 		log.Printf("Deleting pipeline %s ...", state.Name.ValueString())
 		_, err := deletePipeline(ctx, p.client.genqlient, state.Id.ValueString())
+
 		if err != nil && isResourceNotFoundError(err) {
 			return nil
 		}
 
+		if err != nil && isActiveJobsError(err) {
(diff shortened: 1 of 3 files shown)


tomowatt (approved):
🚀

PR #947, opened 2025-07-02
chore: remove deprecated dependabot reviewers and annotation fixes

tomowatt:
Dependabot reviewers are being deprecated see - https://github.blog/changelog/2025-04-29-dependabot-reviewers-configuration-option-being-replaced-by-code-owners/ - and we already have CODEOWNERS set up so removing them to avoid further comments on PRs from Dependabot

Additionally, adding `context` to annotations as the 'docs' annotations were being overridden by using the `default` context which were also used by the 'testacc' annotations.

And committing the 'docs' changes from `make docs`


tomowatt:
For context the backticks resulted in the following in the log ouput, not ideal or expected
image

and the 'docs' annotations were overwritten by using the _default_ context resulting in only 1 annotation rather than 2 when the 'docs' failed
image

PR #949, opened 2025-07-03
Update the Graph with the latest definitions

CerealBoy:
This is primarily for the changes to macOS hosted agents shapes, however I think the version bump / change in the way the generator functions has changed how some of the `schema.graphql` is captured so the diff is ginormous. I'm expecting once this PR is opened there will be the ignore whitespace option which should simplify the review a bit.


scadu:
@CerealBoy did format of the comments generated for GraphQL schema changed? Looks like it breaks down single-line comments into three lines now which seems to be biggest contributing factor to the diff size (and the schema itself).

PR #962, opened 2025-07-28
Supporting the macOS version for Hosted macOS Cluster Queues

### LINKED ISSUE #959: Support MacOS version in buildkite_cluster_queue
Right now there's no way to specify the version of MacOS in the `buildkite_cluster_queue` resource. This is necessary because the version of XCode can depend on the OS used. Could you add support for it? This syntax would likely make sense:

```
resource "buildkite_cluster_queue" "hosted_agents_macos" {
  cluster_id = buildkite_cluster.primary.id
  key        = "hosted-agents-macos"

  # Pause dispatch after create
  dispatch_paused = true

  hosted_agents = {
    instance_shape = "MACOS_M2_4X7"

    mac = {
      macos_version = "15.5" # NEW
      xcode_version = "16.4"
    }
  }
}
```

CerealBoy:
## Description

~Pending the availability of the option via the GraphQL API, t~ This PR aims to make the `macos_version` parameter available for Hosted Agent Cluster Queues. I've used my local environment to update the generated GraphQL content.

Maybe it's too late in the today for me to see how to do so, how can we make the value entirely optional?

### Verification scenarios

The following scenarios have been tested, their checked value is for the success of that capability.

- [x] Creating a Cluster Queue with `macos_version`
- [x] Updating a Cluster Queue with `macos_version`
- [x] Updating a Cluster Queue without `macos_version`
- [x] Creating a Cluster Queue without `macos_version`
- [x] Add the `macos_version` on existing Queue without it set
- [x] Remove the `macos_version` from an existing Queue (state management)
- [ ] Update tests and add appropriate new ones

## Context

Resolves #959

### DIFF
--- a/buildkite/resource_cluster_queue.go
+++ b/buildkite/resource_cluster_queue.go
@@ -491,7 +491,7 @@ func (cq *clusterQueueResource) Update(ctx context.Context, req resource.UpdateR
 			}
 
 			if !plan.HostedAgents.Mac.MacosVersion.IsNull() {
-				var version HostedAgentMacOSVersion = HostedAgentMacOSVersion(plan.HostedAgents.Mac.MacosVersion.ValueString())
+				version := HostedAgentMacOSVersion(plan.HostedAgents.Mac.MacosVersion.ValueString())
 				hosted.PlatformSettings.Macos.MacosVersion = &version
 			}
 		}


schema.graphql
```diff
@@ -2763,6 +2763,10 @@ Configuration options for the base image of hosted agent instances on macOS plat
 """
 type HostedAgentMacOSSettingsType {
 """
+The macOS version to use for macOS hosted agent instances for this cluster queue.
+"""
+	macosVersion: String
```
matthewborden, line 2768:
```suggestion
	macosVersion: HostedAgentMacosVersion
```

I would have expected this to be a HostedAgentMacosVersion

scadu:
@CerealBoy looking good! I see you’ve initially meant to add new/update tests. Not sure if we need anything here, or it’s just a copy-pasta leftover?

PR #963, opened 2025-07-28
Allow for the InstanceShape of a Queue to not require re-creation

### LINKED ISSUE #960: Cluster requires recreate when changing instance shape
**Describe the bug**
When changing the instance shape on a hosted mac instance in the `buildkite_cluster_queue` resource, terraform wants to recreate the entire queue. But in the UI I can change the shape without needing to do that. Terraform should do the same.

**To Reproduce**
1. Create a hosted mac queue with one instance shape and apply
2. Change the instance shape to a different one and plan, terraform wants to recreate the queue

**Expected behavior**
The instance shape on the queue should be modified without recreating the queue

CerealBoy:
## Description

I think we had some nuances with the GraphQL API a little while ago where it wasn't possible to update the instance shape of a Queue without destroying and creating it again. This is no longer the case, or at the very least it is now possible to change this value with an update as expected.

I'm not sure if there are tests associated with this functionality, nor if there's anything else related to this that requires, in dev when I tried to change a Queue with this code diff it was updated rather than re-created.

## Context

Fixes #960

### DIFF
--- a/buildkite/resource_cluster_queue.go
+++ b/buildkite/resource_cluster_queue.go
@@ -217,9 +217,6 @@ func (clusterQueueResource) Schema(ctx context.Context, req resource.SchemaReque
 								append(MacInstanceShapes, LinuxInstanceShapes...)...,
 							),
 						},
-						PlanModifiers: []planmodifier.String{
-							stringplanmodifier.RequiresReplace(),
-						},
 					},
 				},
 			},


mcncl (approved):

matthewborden:
Sorry I missed the review train here, I did want to mention that:

> I think we had some nuances with the GraphQL API a little while ago where it wasn't possible to update the instance shape of a Queue without destroying and creating it again. This is no longer the case, or at the very least it is now possible to change this value with an update as expected.

The GraphQL API still prevents changing the Machine Type (Linux vs Mac) and the Architecture (AMD to ARM) of a hosted queue. We could relax the restrictions on changing the architecture, but changing the machine type probably justifies a re-creation since the parameters of a Mac queue are so different to the parameters of a linux queue.

CerealBoy:
> The GraphQL API still prevents changing the Machine Type (Linux vs Mac) and the Architecture (AMD to ARM) of a hosted queue. We could relax the restrictions on changing the architecture, but changing the machine type probably justifies a re-creation since the parameters of a Mac queue (MacOS Version, XCode Version) are so different to the parameters of a Linux queue (Base Image).

The Linux vs Mac and AMD vs ARM change still requires re-creation (there's separate validation for that), this is purely for changing the Shape specifically.

PR #982, opened 2025-09-19
Add support for merge queue settings to terraform provider

jonathanly:
Modifies the terraform provider to support the new Buildkite's enhanced support for GitHub merge queues.

The settings exposed:
- `build_merge_group_checks_requested`
- `cancel_when_merge_group_destroyed`
- `use_merge_group_base_commit_for_git_diff_base`

### DIFF
--- a/docs/resources/pipeline.md
+++ b/docs/resources/pipeline.md
@@ -173,9 +173,8 @@ resource "github_repository_webhook" "my_webhook" {
 Optional:
 
 - `build_branches` (Boolean) Whether to create builds when branches are pushed.
-- `build_merge_group_checks_requested` (Boolean) Whether to create builds when merge queues and checks are requested for a merge group.
-- `build_pull_request_base_branch_changed` (Boolean) Whether to create builds for pull requests when its base branch changes.
 - `build_merge_group_checks_requested` (Boolean) Whether to create merge queue builds for a merge queue enabled GitHub repository with required status checks
+- `build_pull_request_base_branch_changed` (Boolean) Whether to create builds for pull requests when its base branch changes.
 - `build_pull_request_forks` (Boolean) Whether to create builds for pull requests from third-party forks.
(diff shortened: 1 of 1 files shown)


go.mod
jonathanly, line 1: Hey @buildkite/support friends - are these changes to the `go.mod` and `go.sum` files to be expected?

tomowatt: @jonathanly I'd run the following locally:
```
go mod verify
go mod tidy
go mod verify
```
As I don't believe we need to be adding all those extras, as saw a few of them that we removed previously as weren't essential.

scadu: Yep, no need for those. I thought `make docs` also runs cleanup, but it looks like we only added it to the lefthook-managed [pre-commit hook ](https://github.com/buildkite/terraform-provider-buildkite/blob/main/lefthook.yml).

buildkite/resource_pipeline_test.go
```diff
@@ -731,6 +734,9 @@ func TestAccBuildkitePipelineResource(t *testing.T) {
 						resource.TestCheckResourceAttrPair("buildkite_pipeline.pipeline", "cluster_id", "buildkite_cluster.cluster", "id"),
 						resource.TestCheckResourceAttrPair("buildkite_pipeline.pipeline", "cluster_name", "buildkite_cluster.cluster", "name"),
 						resource.TestCheckResourceAttr("buildkite_pipeline.pipeline", "provider_settings.ignore_default_branch_pull_requests", "true"),
+						resource.TestCheckResourceAttr("buildkite_pipeline.pipeline", "provider_settings.build_merge_group_checks_requested", "true"),
```
scadu, line 737: We might also verify that settings are disabled with `false`, just to rule out potential overwrites of defaults (not necessarily in this PR, but I'd appreciate having a guard while working on the provider in general).

lizrabuya (approved):
Looks good to me!

scadu (approved):

PR #992, opened 2025-10-02
Add support for Portals

scadu:
Add support for Portals[1]

1. https://buildkite.com/docs/apis/portals


mcncl (approved):
This looks good to me, I don't see the endpoints documented anywhere, but the tests run and pass so they must exist. I assume they're just not in `/docs` yet.

PR #1060, opened 2026-02-23
Cluster secrets

stephanieatte:
Add validation to reject reserved BUILDKITE_ and BK_ key prefixes in cluster secrets. Secret keys must NOT start with:

`BUILDKITE` - Reserved for Buildkite system variables
`BK` - Reserved prefix shorthand


examples/resources/buildkite_cluster_secret/resource.tf
```diff
@@ -4,6 +4,8 @@ resource "buildkite_cluster_secret" "example" {
   value       = "super-secret-password"
   description = "Production database password"
   policy      = <<-EOT
+    pipeline_slug: my-pipeline
+    branch: main
```
mcncl: Maybe the same issue as before? Additionally, should this be `branch` or `build_branch` as in the example below it?

The entry below hasn't been deleted, if that was the intent?

go.mod
```diff
@@ -56,6 +56,8 @@ require (
 	github.com/yuin/goldmark-meta v1.1.0 // indirect
 	go.abhg.dev/goldmark/frontmatter v0.2.0 // indirect
 	golang.org/x/exp v0.0.0-20230626212559-97b1e661b5df // indirect
+	golang.org/x/sync v0.18.0 // indirect
+	golang.org/x/tools v0.38.0 // indirect
```
mcncl:
```suggestion
```
Remove these as they're superseded below

mcncl (changes requested):

PR #1091, opened 2026-03-13
Fix provider_settings not being read during pipeline import

ellsclytn:
During import, ImportStatePassthroughID only sets the id field, leaving ProviderSettings nil. The nil check caused Read to skip populating provider_settings entirely, resulting in Terraform planning to create the full block. Even after fixing that, any single config diff caused all unset Optional+Computed fields to show spurious planned changes because they lacked UseStateForUnknown modifiers.

Fixes #1089

### DIFF
--- a/buildkite/resource_pipeline.go
+++ b/buildkite/resource_pipeline.go
@@ -737,6 +737,9 @@ func (*pipelineResource) Schema(ctx context.Context, req resource.SchemaRequest,
 					"trigger_mode": schema.StringAttribute{
 						Computed: true,
 						Optional: true,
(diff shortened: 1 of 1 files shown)


petetomasik:
Superseded by #1191, merged on `main`. It fixes the same root cause (`ImportStatePassthroughID` only sets `id`, so `Read`'s `nil` check on `ProviderSettings` never fires right after import), using a private-state marker to force one `provider_settings` refetch on the read that follows import, plus `ImportStateVerify` test coverage. Different approach than the `UseStateForUnknown` modifiers here, but it resolves the same bug.

PR #1135, opened 2026-05-07
feat: adds write-only value option for buildkite secret resource

zscholl:
## Description

Adds a [write only argument](https://developer.hashicorp.com/terraform/language/manage-sensitive-data/ephemeral#write-only-arguments) for secret value on the `buildkite_cluster_secret` resource.

You can use this to ensure that the secret values don't end up in terraform state. This seemed like a natural thing to do given the BuildKite API treats the value as write only (there's no GET read secret API option).

Keeps the existing value for backwards compatability. 

Example usage
```hcl
variable "api_token" {
  type      = string
  sensitive = true
  ephemeral = true
}

variable "api_token_version" {
  type = string
}

resource "buildkite_cluster_secret" "write_only_example" {
  cluster_id       = "01234567-89ab-cdef-0123-456789abcdef"
  key              = "API_TOKEN"
  value_wo         = var.api_token
  value_wo_version = var.api_token_version
  description      = "API token"
}
```

## Testing

Added unit and acceptance tests, ran acceptance testing locally and they passed


buildkite/resource_cluster_secret.go
jeremybumsted, line 1: The write-only CRUD flow looks good. I think there is a possible edge case here worth documenting:

Imported secrets will not have either `value` or `value_wo_version` in state because the API can't return secret values and `ImportState` only seeds `cluster_id`/`id`. For `value_wo`, that means the first apply after import will re-send/update the secret because `value_wo_version` chagnes from null to the configured value. Could we call that out in the import docs?

buildkite/resource_cluster_secret_test.go
jeremybumsted, line 1: Test coverage looks good here and acceptance tests pass when I test locally, good stuff!

Could we add an acceptance step for migrating an existing resource from `value` to `value_wo` + `value_wo_version`? The current write-only acceptance test starts with `value_wo`, but existing users will likely migrate from the legacy `value` attribute. 

An end-to-end test would exercise the important behaviour here: the update is triggered by `value_wo_version`, while the secret value itself must come from `req.Config`

README.md
```diff
@@ -2,7 +2,7 @@
 
 [![Build status](https://badge.buildkite.com/7224047dadf711cab2facd75939ea39848850d7c5c5a765acd.svg?branch=main)](https://buildkite.com/buildkite/terraform-provider-buildkite-main)
 
-This is the official Terraform provider for [Buildkite](https://buildkite.com). The provider is listed in the [Terraform Registry](https://registry.terraform.io/) and supports Terraform >= 1.0.
+This is the official Terraform provider for [Buildkite](https://buildkite.com). The provider is listed in the [Terraform Registry](https://registry.terraform.io/) and supports Terraform >= 1.11.
```
jeremybumsted, line 5: This PR brings the minimum required version up to 1.11, do we need 1.11 in this case? That may introduce other breaking changes so if it's not required for this change, maybe we can keep it at 1.0?

jeremybumsted (commented):
Overall I think this looks good @zscholl thanks for raising this and apologies for the slow review!! A few comments/questions before I think we're good to go.

README.md
zscholl: The `write-only` argument is only supported in `1.11` and higher. Looks like you can still use the provider without setting the `wo` values on a version under `1.11`. Would you rather we document this as a condition of using the `wo` arguments or just generally bump the minimum version?

I ran this branch on terraform `1.10` and was able to plan/apply when not using the `wo` values.

https://developer.hashicorp.com/terraform/language/manage-sensitive-data/write-only#requirements

buildkite/resource_cluster_secret.go
zscholl: Yeah good call! Updated the docs

buildkite/resource_cluster_secret_test.go
zscholl: Yes also a good call! Added a test and also a couple others for ensuring that `value_wo_version` is non empty and that `value` was not present in state after migrating to `value_wo`

README.md
jeremybumsted: @zscholl had some discussions with the team and we want to ensure we're supporting folks that might still be using older versions of Terraform as well, so documenting the requirement for using this particular feels like the right direction. 

I'm also thinking it would make sense to add a check for the Terraform version, and perhaps fail out/print a warning that a minimum version is required to make use of `wo` arguments.

What do you think?

zscholl: Actually terraform already provides such an error!

If you attempt to run a plan or apply on a version `<1.11` and you use a `wo` argument you'll see something like

```
Error: WriteOnly Attribute Not Allowed

  with buildkite_cluster_secret.write_only,
  on main.tf line 18, in resource "buildkite_cluster_secret" "write_only":
  18:   value_wo         = "write-only-value"

The resource contains a non-null value for WriteOnly attribute value_wo.
Write-only attributes are only supported in Terraform 1.11 and later.
```

jeremybumsted (dismissed):
Looks good to me!

README.md
jeremybumsted: Ah amazing - TIL!

buildkite/resource_cluster_secret.go
scadu: One small thing: we'd need to add a template, so `make docs` doesn't remove it from the edited doc. A new template file should be created at `templates/resources/cluster_secret.md.tmpl`. Contents:

```
---
# generated by https://github.com/hashicorp/terraform-plugin-docs
page_title: "{{.Name}} {{.Type}} - {{.ProviderName}}"
subcategory: ""
description: |-
{{ .Description | plainmarkdown | trimspace | prefixlines "  " }}
---

# {{.Name}} ({{.Type}})

{{ .Description | trimspace }}

{{ if .HasExample -}}
## Example Usage

{{tffile .ExampleFile }}
{{- end }}

{{ .SchemaMarkdown | trimspace }}
{{- if .HasImport }}

## Import

Using `terraform import`, import resources using the `id`. For example:
{{codefile "shell" .ImportFile }}

Because Buildkite does not return secret values after they are set, imported secrets cannot populate `value` or `value_wo_version` in Terraform state. Configure either `value` or `value_wo` before the next plan. When using `value_wo`, the first apply after import will re-send the configured secret value because `value_wo_version` changes from no state value to the configured value.

In Terraform v1.5.0 and later, use an [`import` block](https://developer.hashicorp.com/terraform/language/import) to import instances using the `id`. For example:
{{ $tfImportBlock := print "examples/resources/" .Name "/import.tf" }}{{tffile $tfImportBlock }}
{{- end }}
```

scadu (approved):

scadu:
Thanks again for your contribution, @zscholl! Merging now 🚀

PR #1146, opened 2026-05-14
[SUP- 7066]: feat: add `use_step_key_as_commit_status` provider setting

### LINKED ISSUE #1145: `use_step_key_as_commit_status` setting not exposed
**Is your feature request related to a problem? Please describe.**
The Buildkite API has a `use_step_key_as_commit_status` option, but the Terraform provider does not expose this option.

**Describe the solution you'd like**
Expose `use_step_key_as_commit_status` in the Terraform provider so users can set that option for pipelines.

**Describe alternatives you've considered**
N/A

**Additional context**
`use_step_key_as_commit_status` is already in the API so just needs to be exposed in the provider.

**Potential impact**
Allows users to set `use_step_key_as_commit_status` via the Terraform provider.

omarjatoi:
This PR exposes the `use_step_key_as_commit_status` setting in the Terraform provider. Closes https://github.com/buildkite/terraform-provider-buildkite/issues/1145.


buildkite/resource_pipeline.go
```diff
@@ -949,6 +950,14 @@ func (*pipelineResource) Schema(ctx context.Context, req resource.SchemaRequest,
 							boolplanmodifier.UseNonNullStateForUnknown(),
 						},
 					},
+					"use_step_key_as_commit_status": schema.BoolAttribute{
+						Computed:            true,
+						Optional:            true,
+						MarkdownDescription: "Whether to use step keys as commit status names for per-step commit statuses. Defaults to false.",
```
stephanieatte: Could you update this to mention that publish_commit_status and publish_commit_status_per_step also need to be set to true for this to work? e.g. "Whether to use step keys as commit status names for per-step commit statuses. Requires publish_commit_status and publish_commit_status_per_step to also be enabled. Defaults to false."

buildkite/resource_pipeline_provider_settings_test.go
```diff
@@ -0,0 +1,48 @@
+package buildkite
```
stephanieatte: Could you move these tests into buildkite/resource_pipeline_test.go to keep consistent with the rest of the test suite?

stephanieatte (changes requested):
Thanks for the PR Omar! Two small things before we merge:

1. Could you update the MarkdownDescription for use_step_key_as_commit_status to mention that publish_commit_status and publish_commit_status_per_step also need to be set to true for this setting to have any effect.
2. Could you move the tests from resource_pipeline_provider_settings_test.go into resource_pipeline_test.go to keep consistent with the rest of the test suite.

Thanks again for the contribution! 🙌

omarjatoi:
Done, thanks for the review 🫡

stephanieatte (approved):
LGTM! changes addressed

PR #1153, opened 2026-05-22
Fix bug with empty `env` map

### LINKED ISSUE #1152: Setting `env = {}` in `resource "buildkite_pipeline_schedule` causes infinite drift
**Describe the bug**
Instead of just keeping `env = {}` in the state or being a no-op, we get infinite drift

**To Reproduce**
```hcl
resource "buildkite_pipeline_schedule" "test" {
  pipeline_id = buildkite_pipeline.pipeline.id
  branch      = ...
  cronline    = ...
  label       = ...
  env         = {}
}
```

Plan gives this, even right after an apply:
```hcl
  # buildkite_pipeline_schedule.test will be updated in-place
  ~ resource "buildkite_pipeline_schedule" "test" {
      + env         = {}
        id          = "..."
        # (8 unchanged attributes hidden)
    }
```

**Expected behavior**
No infinite drift; state should be stable

petetomasik:
## Summary

Fixes #1152. Setting `env = {}` on `buildkite_pipeline_schedule` caused infinite drift because `Read` coerced empty-map state to null, conflicting with the empty-map config on every refresh.

The fix preserves the prior state's empty-vs-null map shape in `updatePipelineScheduleNode` when the API returns no env vars. The API doesn't distinguish `env = {}` from omitted `env`, so we keep whichever shape the user wrote.

## Changes

- Skip overwriting `psState.Env` when the API returns no env vars and prior state was a non-null empty map (`buildkite/resource_pipeline_schedule.go`)
- Add regression test for `env = {}` with `PostApplyPostRefresh.ExpectEmptyPlan()`
- Add transition test exercising empty -> populated -> empty
- Document the empty-map equivalence on the `env` attribute

### DIFF
--- a/buildkite/resource_pipeline_schedule.go
+++ b/buildkite/resource_pipeline_schedule.go
@@ -103,11 +103,9 @@ func (ps *pipelineSchedule) Schema(ctx context.Context, req resource.SchemaReque
 				MarkdownDescription: "The message the builds show for builds created by this schedule.",
 			},
 			"env": resource_schema.MapAttribute{
-				ElementType: types.StringType,
-				Optional:    true,
-				MarkdownDescription: "The environment variables that scheduled builds should use. " +
-					"An empty map (`{}`) is equivalent to omitting this attribute at the API level, " +
-					"though Terraform preserves whichever shape you wrote in your configuration.",
+				ElementType:         types.StringType,
+				Optional:            true,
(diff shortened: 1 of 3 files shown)


mcncl (dismissed):
Looks good!

scadu (approved):

PR #1156, opened 2026-06-01
SUP-2308: Read pipeline provider_settings via GraphQL

tomowatt:
## Summary

`buildkite_pipeline`'s `Read` now sources `provider_settings` from the **GraphQL API** instead of a per-refresh REST `GET`, reducing REST API usage on refresh for organizations with lower REST rate limits (SUP-2308).

## Approach

- **Kept the existing `*providerSettingsModel` pointer model**, the `archived` attribute, and all `provider_settings` fields. This deliberately avoids the `types.Object` / `Computed` / custom-plan-modifier refactor from the earlier attempt, which caused ~266 "Provider returned invalid result object after apply" failures.
- **Dedicated fragment/query:** `repository.provider.settings` is fetched via a `RepositoryProviderSettingsFields` fragment + `getPipelineProviderSettings` query, kept **out of** the shared `PipelineFields` fragment. `PipelineFields` is used by `getNode` (the generic node read shared with other resources, e.g. `pipeline_team`) and the create/update mutations; the provider-settings subtree can error server-side for some pipelines, so requesting it on every pipeline read previously coupled unrelated reads to its resolution.
- `Read` calls the dedicated query **only when the user configured `provider_settings`**, and **surfaces any error** (never swallowed — no stale state).
- Create/Update provider-settings **writes** stay on REST (no GraphQL mutation input exists).
- `issue_comment_match_mode` is a GraphQL enum (`EXACT`/`CONTAINS`); it is lowercased to `exact`/`contains` to match the schema validator. Bitbucket's nullable `cancelDeletedBranchBuilds`/`skipBuildsForExistingCommits` are aliased to avoid a union field-merge shape conflict with GitHub's non-null versions.

## Depends on backend GraphQL changes

Reading `provider_settings` from the GraphQL API requires two changes to the Buildkite GraphQL API (tracked internally under SUP-2308) that must be **deployed to production before this provider release ships**. Until both are live, acceptance tests that read provider settings fail with *"An error occurred while executing your GraphQL query."*

1. **Expose `use_step_key_as_commit_status`** on the GitHub repository provider settings type (and GitHub Enterprise), bringing it to parity with the REST API.
2. **Coerce blank command-word match modes to null.** `issue_comment_match_mode` / `review_comment_match_mode` can be stored as `""`, which is not a valid value for the match-mode enum and caused the GraphQL field to raise during result coercion (surfaced as the generic error above), breaking reads of affected pipelines. The fix returns `null` for blank values.

## Reconciled with #1157

#1157 (merged via `main`) added 10 GitHub webhook-trigger `provider_settings` fields (`build_check_run_completed`, `build_create_event`, `build_deployment_status_created`, `build_pull_request_converted_to_draft`, `build_pull_request_review_{requested,dismissed,submitted}`, `build_release_{created,published,released}`) read via REST. These are now also mapped through the GraphQL read so `provider_settings` stays fully GraphQL-sourced (no drift). All 39 model fields are covered by the GraphQL mapping.

## Test plan

- [x] `go build ./...`, `go vet ./...`, `gofmt`, `go test ./...` (unit) pass; new unit test `TestMapProviderSettingsFromGraphQLGitHub`
- [x] Local e2e against a dev instance: `create_pipeline_setting_all_attributes` (provider_settings incl. `use_step_key_as_commit_status` + the #1157 fields round-trip via the GraphQL read with no error/drift)
- [x] Full acceptance suite in CI against the test org (the dev org used locally has teams disabled / mandatory clusters / the v0→v1 tests use the external 0.27.0 provider)

## PR checklist

- [x] tests added/updated
- [x] `docs/` (no schema/attribute changes in this provider PR)
- [ ] `CHANGELOG.md` (release-please managed; carried by conventional-commit messages)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


tomowatt:
This is ready for review but depends on the backend changes being made first and the schema being regenerated once again to ensure it's up to date

mcncl:
@tomowatt I keep meaning to ask what the reason for this would be? Are we seeing a lot of API rate limiting being inflicted on folks because of this? Does this just shift the burden to the mystery of GraphQL complexity?

tomowatt:
@mcncl this was initially raised as an issue when the REST API limits got introduced and we have customers hit issues due to the scale of the their TF usage of their BK resources - Pipelines - which meant they would get rate limited quickly or had to try alternative ways to beat the REST API rate limits.
The GraphQL Complexity shouldn't be problematic as it's targeted onto a specific Pipeline and only made if the Provider setting are in the Resource configuration. Though on the latter point of making the additional query if Provider Setting are configured will need to be reviewed, but kept backwards compatibility.
Given the mixture of REST and GraphQL needed and used elsewhere in the Pipeline Resource, it's a large change for a small improvement.
We could make the Read function use the REST API only which would be a smaller change but the issues people had were with the REST API limits rather than GraphQL.

mcncl (approved):
Looks fine. I think generally we'd opt to use REST but due to the varying limitations of it (different rate limits etc) we can roll this out and see how we go.

PR #1162, opened 2026-06-12
fix: archive pipeline after REST updates in Create and Update

inez:
The REST API rejects any update to an archived pipeline (422 "Cannot update an archived pipeline"). `Create` and `Update` archived the pipeline before calling `updatePipelineExtraInfo`, so setting `archived = true` on any pipeline with `provider_settings` always failed mid-apply (and left the pipeline archived but the apply errored).

Move the archive mutation to the end of both flows, after the slug and provider settings REST calls. Add acceptance tests covering `archived` together with `provider_settings` for both create and update.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/buildkite/resource_pipeline_test.go
+++ b/buildkite/resource_pipeline_test.go
@@ -1851,4 +1851,101 @@ func TestAccBuildkitePipelineResource(t *testing.T) {
 			},
 		})
 	})
+
+	t.Run("archives pipeline with provider_settings", func(t *testing.T) {
+		// Regression test: archiving must happen after the provider_settings REST
+		// PATCH, otherwise the API rejects it with "Cannot update an archived
+		// pipeline".
+		pipelineName := acctest.RandString(12)
+		configFor := func(archived bool) string {
+			return fmt.Sprintf(`
+				resource "buildkite_pipeline" "pipeline" {
+					name = "%s"
+					repository = "https://github.com/buildkite/terraform-provider-buildkite.git"
+					archived = %t
+
+					provider_settings = {
+						build_branches = true
+						build_pull_requests = false
+					}
+				}
+			`, pipelineName, archived)
+		}
+
+		resource.ParallelTest(t, resource.TestCase{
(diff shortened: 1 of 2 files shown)


buildkite/resource_pipeline_test.go
```diff
@@ -1851,4 +1851,90 @@ func TestAccBuildkitePipelineResource(t *testing.T) {
...
+		resource.ParallelTest(t, resource.TestCase{
+			PreCheck:                 func() { testAccPreCheck(t) },
+			ProtoV6ProviderFactories: protoV6ProviderFactories(),
+			CheckDestroy:             testAccCheckPipelineDestroyFunc,
+			Steps: []resource.TestStep{
+				{
+					Config: config,
+					Check: resource.ComposeAggregateTestCheckFunc(
+						resource.TestCheckResourceAttr("buildkite_pipeline.pipeline", "archived", "true"),
+					),
```
Megh03, line 1946: One small suggestion: for the "creates archived pipeline with provider_settings" test, could we add a direct API check alongside the Terraform state check, similar to what the update test above does? That way we're verifying the pipeline is actually archived in Buildkite and not just in  Terraform state :)


```suggestion
					Config: config,
					Check: resource.ComposeAggregateTestCheckFunc(
    resource.TestCheckResourceAttr("buildkite_pipeline.pipeline", "archived", "true"),
    func(s *terraform.State) error {
        slug := fmt.Sprintf("%s/%s", getenv("BUILDKITE_ORGANIZATION_SLUG"), pipelineName)
        resp, err := getPipeline(context.Background(), genqlientGraphql, slug)
        if err != nil {
            return err
        }
        if !resp.Pipeline.Archived {
            return fmt.Errorf("expected pipeline to be archived, got archived = false")
        }
        return nil
    },
),
```

inez: Good call. Done.

Megh03 (approved):
LGTM 🙂

PR #1173, opened 2026-07-01
fix: retry transient backend errors and stabilize parallel acceptance tests

petetomasik:
## Summary

Cleans up acceptance tests to run reliably in parallel and adds retry handling for a transient backend error that the HTTP client can't see.

## Retry transient "currently busy" errors

Some Buildkite mutations (notably `clusterCreate`) can return a transient throttle such as `Cluster creation is currently busy, please try again.`. This arrives in a GraphQL 200 response body, so the underlying `retryablehttp` client never retries it and the apply fails.

- Added `isTransientError` in `util.go`, which matches `currently busy` / `please try again` (case-insensitive) against both `gqlerror.List` messages and plain errors.
- `retryContextError` now returns a `RetryableError` for these transient cases and keeps everything else non-retryable, so `RetryContext` backs off within its existing timeout instead of failing immediately.
- Added `TestRetryContextError` covering nil, transient string/gqlerror inputs, and non-retryable cases (generic errors, not-found).

## Acceptance test cleanup

- Set `default_team = false` in team, pipeline_team, and team_member test configs so tests don't collide on the org's single default team.
- Randomized hardcoded team names in the team data source tests with `acctest.RandString(12)` to avoid conflicts between parallel runs.
- Switched cluster secret tests from `resource.Test` to `resource.ParallelTest`.
- Fixed the `ckeckUpdated` typo in the organization test.
- Bumped the test `-parallel` flag from 4 to 12 in the Makefile.

### DIFF
--- a/buildkite/retry_test.go
+++ b/buildkite/retry_test.go
@@ -1,7 +1,10 @@
 package buildkite
 
 import (
+	"errors"
 	"testing"
+
+	"github.com/vektah/gqlparser/v2/gqlerror"
 )
 
 // This tests retry logic in general. Both REST and GraphQL clients are wrapped with `retryablehttp.Client`.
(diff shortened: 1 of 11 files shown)


buildkite/util.go
```diff
@@ -19,6 +19,7 @@ var (
 	resourceNotFoundRegex = regexp.MustCompile(`(?i)(No\s+\w+(\s+\w+)*\s+found|not\s+found|no\s+longer\s+exists)`)
 	activeJobsRegex       = regexp.MustCompile(`(?i)(active\s+(builds|jobs)|running\s+(builds|jobs)|builds?\s+are\s+running|jobs?\s+are\s+running)`)
 	alreadyExistsRegex    = regexp.MustCompile(`(?i)(already\s+been\s+added|already\s+exists)`)
+	transientErrorRegex   = regexp.MustCompile(`(?i)(currently\s+busy|please\s+try\s+again)`)
```
scadu: Thought: currently busy alone matches the clusterCreate throttle. please try again also matches permanent REST 4xx (e.g. suite-update 422), which retryablehttp won't retry, so RetryContext spins until timeout. 429/503 are already retried at the transport layer.

```suggestion
	transientErrorRegex   = regexp.MustCompile(`(?i)currently\s+busy`)
```

buildkite/retry_test.go
```diff
@@ -67,3 +70,63 @@ func TestClientCreation(t *testing.T) {
...
+		{
+			name:          "cluster creation busy is retryable",
+			err:           errors.New("input:3:2: clusterCreate Cluster creation is currently busy, please try again."),
+			wantRetryable: true,
+		},
+		{
+			name:          "please try again is retryable",
+			err:           errors.New("something went wrong, please try again"),
+			wantRetryable: true,
+		},
```
scadu, line 95: Flip to non-retryable to match the narrowed regex and pin the boundary.

```suggestion
		{
			name:          "generic please try again is not retryable",
			err:           errors.New("something went wrong, please try again"),
			wantRetryable: false,
		},
```

scadu (changes requested):
I left some comments on the retry change only.

The clusterCreate throttle is the one GraphQL mutation returning this phrasing, raised on a failed lock, so retrying is safe.

Not in this diff, but now that closures can retry, Client.GetOrganizationID in client.go caches orgId before its error check, so a transient failure there returns a cached empty org ID on the retry. Worth caching only on success.

scadu (approved):
Looking great!

PR #1204, opened 2026-08-11
Add cluster network ranges data source

petetomasik:
Exposes the egress IP ranges a cluster's hosted agents connect out from, via
`GET /v2/organizations/{org}/clusters/{cluster}/network_ranges`. Customers allowlisting
Buildkite hosted agents in a firewall or security group currently copy CIDRs out of the UI by
hand and keep them in sync manually; this lets the allowlist live in the same Terraform config
as the infrastructure it protects. Read-only, so it is a data source rather than a resource.

```hcl
data "buildkite_cluster_network_ranges" "hosted" {
  cluster_uuid = data.buildkite_cluster.hosted.uuid
}
```

## Notable decisions

**`cluster_uuid`, not `cluster_id`.** The REST path segment must be a UUID, and `.id` is the base64 GraphQL ID.
The attribute is named `cluster_uuid` to match `buildkite_cluster_queue` and
`buildkite_cluster_maintainer`, and is validated against the UUID shape because the API answers
a GraphQL ID with an opaque `{"message": "Not found"}` 404. The validator is case-insensitive to
agree with the API, which is why it does not reuse the lowercase-only `isUUID` in `util.go`.

**`ranges` is a set, not a list.** The upstream service does not guarantee ordering, so a list
would produce spurious diffs for anyone referencing `ranges[0]` or interpolating the collection
into a resource argument. Worth settling now, since changing this after release would need a
state upgrader.

**Empty rather than null for non-hosted clusters.** Clusters without hosted agents return an
empty collection, and the model builds a non-nil slice so the attribute is a known empty set. A
nil slice would reach state as null and break `for` expressions over it.

**Null fields stay null.** `cidr_range` and `kind` decode into `*string`, so a null from
upstream does not silently become `""` and reach a firewall rule as an empty CIDR.

## Testing

`make test` covers `Read` end to end by pointing the provider at a stub with `rest_url`: two
ranges mapped into state, an empty collection becoming a known empty set, the 403 diagnostic,
the UUID validator rejecting a GraphQL ID, and JSON decoding including null and absent fields.

The acceptance test creates a cluster and asserts `ranges.#` is `0`, which exercises auth, the
path, and the empty-collection contract against the real API. It passes against a live org.

## PR checklist:
- [x] `docs/` updated <-- regenerated with `make docs`
- [x] tests added <-- unit tests for `Read` and decoding, plus an acceptance test
- [x] an example added to `examples/` <-- `examples/data-sources/buildkite_cluster_network_ranges/`


buildkite/data_source_cluster_network_ranges.go
```diff
@@ -0,0 +1,147 @@
...
+	resp.Schema = schema.Schema{
+		MarkdownDescription: heredoc.Doc(`
+			Use this data source to retrieve the egress network ranges that a cluster's hosted agents
+			connect out from. This is useful for allowing Buildkite hosted agents through a firewall or
+			security group from the same Terraform configuration.
+
+			The API token must have the read_clusters scope, and the user it belongs to must be able to
+			manage the cluster. A token that can only read clusters will receive a 403.
+
+			Clusters that do not use hosted agents have no egress ranges and return an empty list.
```
tomowatt: wondering if we should updated this to mention that it's a hosted agents queue that is required for the cluster to have the egress ranges as it's possible to have a cluster with a mixture of both hosted and self-hosted queues
```suggestion
			Clusters that do not have a hosted agents queue have no egress ranges and return an empty list.
```

tomowatt (dismissed):
One wee comment but LGTM

scadu (approved):

tomowatt (approved):

PR #1205, opened 2026-08-11
fix(rest client): surface API error bodies on retryable failures

petetomasik:
## Summary

`makeRequest` (`buildkite/client.go`) discarded the final response of a retried request, so both ways a REST call can fail after retrying reached the user with the API's own explanation missing. A sustained 5xx surfaced as retryablehttp's `giving up after 4 attempt(s)`, carrying no status code and no body. A read timeout that expired during a backoff wait surfaced as a bare `context deadline exceeded`, which does not even say a request was made. In both cases an operator had no way to tell a misconfigured tenant from an outage without reproducing the call by hand.

The same request now reports:

```
the Buildkite API request failed: GET https://api.buildkite.com/v2/organizations/acme/clusters (status: 503): {"message":"Could not load network ranges: tenant is not configured"} (after 4 attempts)
```

and, when the deadline is what ended it, the same line with `: context deadline exceeded` appended. `context.DeadlineExceeded` stays in the error chain either way.

## Change

Two mechanisms are needed, because retry exhaustion and deadline expiry are separate paths inside retryablehttp and only the first one reaches `ErrorHandler`. `restErrorHandler` covers exhaustion, where the library breaks out of the loop before draining the body and hands the response over intact. A small per-request capture installed on the context covers the deadline case, where the library returns `ctx.Err()`, drops the response, and never calls `ErrorHandler` at all. The capture keeps a bounded prefix of the body from `CheckRetry`, which is the last point the body is readable, and puts the prefix back so the response stays usable for whoever consumes it next.

The `ErrorHandler` is deliberately REST-only. Handing a non-200 back to genqlient produces an `*HTTPError` that carries the raw body and has no `Unwrap`, so `errors.As` for `gqlerror.List` fails and the GraphQL matchers in `util.go` fall through to running their regexes over that body. A 5xx page saying "not found" would then have removed a live resource from state. `TestGraphQLClientHasNoErrorHandler` pins the asymmetry so it does not get tidied away later.

Because the error message now carries an API-supplied body, matching on it became unsafe: a proxy error page mentioning "404" would have read as a missing resource. Failures are therefore a typed `*apiError` exposing `StatusCode`, and the six call sites that matched `strings.Contains(err.Error(), "status: 404")` now call `isAPIStatus(err, http.StatusNotFound)`. A transport or context failure reports `StatusCode` 0, so a dying connection can never be mistaken for a status the API returned, and a status held over from an earlier attempt is reported as context rather than as the cause. `isTransientError` gained an `errors.As` guard for the same reason: a 5xx body reading "currently busy" would otherwise have made `retryContextError` re-run a non-idempotent mutation on top of the attempts the HTTP client had already spent.

`Error()` reproduces the previous wording exactly, so nothing user-facing regressed for the paths that already worked.

The second commit is unrelated to the fix and only documents something the work turned up: the REST backoff waits accumulate past the default operation timeout well before the retry budget runs out, so raising `max_retries` on its own does not necessarily buy more attempts.

## Testing

`buildkite/retry_test.go` covers retry exhaustion, deadline expiry mid-wait, rate limiting, retry-then-succeed, unretried client errors, body truncation on both paths that build a message, a non-retryable status ending a retried request, and the message and `isAPIStatus` shapes in table form. Three of them were written against a deliberately broken build to confirm they can fail for the reason they claim: the capture was temporarily made package-level to check that both scoping tests detect it, and the `ErrorHandler` was removed to check that the exhaustion test notices.

`go build`, `go vet`, `gofmt`, `go test ./...`, `go test -race`, and `golangci-lint run ./...` are clean. Acceptance tests were not run; this changes error paths rather than resource behaviour.

## PR checklist:
- [x] `docs/` updated <-- `docs/index.md` regenerated for the `max_retries` description
- [x] tests added
- [ ] an example added to `examples/` <-- nothing to demonstrate, no schema or resource change
- [ ] `CHANGELOG.md` updated with pending release information <-- maintained in dedicated release PRs, and the last three releases are not in it yet


buildkite/provider.go
```diff
@@ -198,7 +198,7 @@ func (*terraformProvider) Schema(ctx context.Context, req provider.SchemaRequest
 			},
 			"max_retries": schema.Int64Attribute{
 				Optional:            true,
-				MarkdownDescription: "Maximum number of retry attempts for retryable HTTP requests. Defaults to 10.",
+				MarkdownDescription: "Maximum number of retry attempts for retryable HTTP requests. Defaults to 10. REST requests are also bounded by the operation timeout, which covers the waits between attempts as well as the requests themselves, so raising this value alone may not produce more attempts.",
```
scadu: Minor: this is specifically the configured read timeout used by `makeRequest`, not every operation's timeout. It also doesn't cover all REST calls: the registry resource and data source call `client.http.Do` directly. Could we narrow this to what the code guarantees?

scadu (dismissed):

petetomasik:
@buildsworth-bk review

scadu (approved):

PR #1207, opened 2026-08-13
fix(organization): skip allowlist and 2FA mutations when unchanged

dattallant:
`buildkite_organization` can't be created on an organization whose plan lacks the API IP allowlist: even an empty `resource "buildkite_organization" "settings" {}` fails with "The API IP allowlist feature is not available for your organization", because Create/Update/Delete always send `organizationApiIpAllowlistUpdate` (with `""` when the attribute is unset). The 2FA mutation is likewise sent on every apply, and Read turned an empty allowlist into `[""]`, so an unset attribute showed a perpetual diff.

Both mutations are now only sent when the value differs from state, `id`/`uuid` come from `getOrganization`, and an empty allowlist stays null when the attribute isn't configured. Dropped `ExpectNonEmptyPlan` from the "removing the allowed API IP address list" step since the plan is now empty, added a subtest without the allowlist attribute plus a table test for the read mapping, and ran the full `TestAccBuildkiteOrganizationResource` locally against a test organization.

## PR checklist:
- [x] tests added
- [ ] an example added to `examples/` <-- no new field

### DIFF
--- a/buildkite/resource_organization.go
+++ b/buildkite/resource_organization.go
@@ -110,7 +110,13 @@ func (o *organizationResource) Create(ctx context.Context, req resource.CreateRe
 	}
 
 	log.Printf("Creating settings for organization %s ...", *org)
-	if err := o.updateAllowedApiIpAddresses(ctx, *org, plan.AllowedApiIpAddresses, types.ListNull(types.StringType)); err != nil {
+	// compare with the organization's current allowlist so a matching one is left as it is
+	current, diags := allowedApiIpAddressesFromAPI(ctx, organization.Organization.AllowedApiIpAddresses, plan.AllowedApiIpAddresses)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	if err := o.updateAllowedApiIpAddresses(ctx, *org, plan.AllowedApiIpAddresses, current); err != nil {
 		resp.Diagnostics.AddError(
 			"Unable to create Organization settings",
 			fmt.Sprintf("Unable to create Organization settings: %s", err.Error()),
--- a/buildkite/resource_organization.go
+++ b/buildkite/resource_organization.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	"github.com/MakeNowJust/heredoc"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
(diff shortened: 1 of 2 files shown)


buildkite/resource_organization.go
```diff
@@ -102,31 +100,34 @@ func (o *organizationResource) Create(ctx context.Context, req resource.CreateRe
...
 	if err != nil {
+		resp.Diagnostics.AddError(
+			"Unable to obtain Organization",
+			fmt.Sprintf("Unable to obtain Organization: %s", err.Error()),
+		)
+		return
+	}
+
+	log.Printf("Creating settings for organization %s ...", *org)
+	if err := o.updateAllowedApiIpAddresses(ctx, *org, plan.AllowedApiIpAddresses, types.ListNull(types.StringType)); err != nil {
```
omehegan: It looks like here we are comparing the plan with a synthetic null value, even if getOrganization returns a non-empty IP allowlist from Buildkite. I think that would mean that if someone already has that value set, there will be some issues getting the plan to converge. Can we instead use whatever value we get back from getOrganization here, so that we're comparing and applying consistently?

omehegan (changes requested):

buildkite/resource_organization.go
dattallant: yep, good catch! create was comparing against a hard-coded null. It now uses the allowlist from the getOrganization call it already makes, so an existing list that matches the
config is left alone, and it still gets replaced/cleared when the config differs. fixed in the latest commit

buildkite/resource_organization.go
```diff
@@ -258,3 +266,14 @@ func (o *organizationResource) Delete(ctx context.Context, req resource.DeleteRe
 
 	resp.Diagnostics.AddAttributeWarning(path.Root("enforce_2fa"), "Enforce 2FA setting left intact", "Use the web UI if you wish to change the value")
 }
+
+// updateAllowedApiIpAddresses sets the API IP allowlist, skipping the mutation when it is unchanged
+func (o *organizationResource) updateAllowedApiIpAddresses(ctx context.Context, orgID string, planned, current types.List) error {
+	// the mutation is rejected for organizations without the allowlist feature, even for ""
+	if planned.Equal(current) {
```
omehegan: I think there is one more issue here - this equality test considers null, `[]`, and `[""]` to be different, but `createCidrSliceFromList` with `strings.Join` below is going to serialize all three of them to the same value, `""`. That means that for an org which has the feature disabled, the Delete is going to compare a null list from the plan with a `""` list in the state, and will consider them different. Can we instead use something like 

```
	plannedValue := strings.Join(createCidrSliceFromList(planned), " ")
	currentValue := strings.Join(createCidrSliceFromList(current), " ")
```

and then compare their values?

omehegan (changes requested):

buildkite/resource_organization.go
dattallant: yeah I think that's simpler. pushed `363afa5` with a table test for the `null / [] / [""]` combos.

omehegan (approved):

PR #1208, opened 2026-08-13
Add buildkite_notification_service resource

scadu:
Adds `buildkite_notification_service` for managing organization-level notification services through the REST API.

Supports:

- Webhook
- Slack Workspace
- Datadog Pipeline Visibility
- AWS EventBridge
- OpenTelemetry control-plane tracing
- Linear

The resource supports lifecycle management and imports, preserves masked secrets in Terraform state, and documents the limitations of OAuth-managed Slack Workspace and Linear services.

Also includes:

- EventBridge event bus and rule examples
- Plan availability and Enterprise entitlement guidance
- Required API token scopes
- JSON content type handling for PATCH requests

This is stacked on #1205, so target needs to be changed to main once that PR is merged.

## PR checklist

- [x] `docs/` updated
- [x] tests added
- [ ] `CHANGELOG.md`


buildkite/resource_notification_service.go
```diff
@@ -0,0 +1,1088 @@
...
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+			"aws_account_id": schema.StringAttribute{
+				Optional:            true,
+				Computed:            true,
+				Sensitive:           true,
+				MarkdownDescription: "The AWS account ID for the partner event source. Required on create, immutable, and masked by the API.",
+				PlanModifiers: []planmodifier.String{
```
petetomasik, line 313: Importing an EventBridge service and then setting `aws_account_id` destroys and recreates it. `initializeImportedSettings` (line 921) sets `AWSAccountID` to null on import because the API masks the value, and this plain `RequiresReplace()` fires on the null -> known transition. The result is that the first apply after an import plans a destroy-before-create, which tears down the AWS partner event source and requires the AWS side to be rewired.

Confirmed with a plan check against a mock returning the masked account ID (`"aws_account_id":"XXXXXXXX9012"`):

```go
plancheck.ExpectResourceAction("buildkite_notification_service.test",
    plancheck.ResourceActionDestroyBeforeCreate)
```

This passes, so the replacement is real. The import notes steer users straight into it. `templates/resources/notification_service.md.tmpl:32` says "Configure the matching provider settings after import if Terraform should manage an available setting", and doing exactly that for EventBridge destroys the service.

The equivalent Datadog path is fine: import-then-set `api_key` produces a clean in-place update, because `api_key` has no `RequiresReplace`.

buildkite/resource_notification_service.go
```diff
@@ -0,0 +1,1088 @@
...
+				"Notification filters are not supported",
+				fmt.Sprintf("%s services are used by pipeline or Test Engine configuration, so notification filters have no effect. Remove %s.", providerType, strings.Join(filters, ", ")),
+			)
+			return
+		}
+	}
+
+	if !req.State.Raw.IsNull() {
+		return
+	}
```
petetomasik: Removing a settings block from an existing resource is a silent no-op that permanently unmanages the settings. This `if !req.State.Raw.IsNull() { return }` guard sits above the "settings are required" check at lines 443-449, so the required-settings validation only ever runs on create. Deleting `webhook = { ... }` from a config whose `provider_type` is still `webhook` therefore plans and applies cleanly.

What happens on that apply:

- `settingsUpdatePayload` breaks out at lines 761-763 because `m.Webhook == nil`, so `settings` never enters the payload.
- `updatePayload` returns an empty map, so `Update` skips the PATCH entirely and falls through to the `r.get` at line 593.
- `applyWebhookAPISettings` returns early at lines 930-932, so state keeps `webhook = null`.

The net effect is that Terraform shows `webhook = { ... } -> null` in the plan, reports the apply as successful, and writes null into state, while the service in Buildkite keeps its URL and token and keeps delivering. It does not self-heal either: on every later refresh `imported` is false and `m.Webhook` is nil, so `applyWebhookAPISettings` returns early again and the plan stays empty forever. The plan promised a change, the apply reported success, nothing happened, and drift detection is permanently off for those settings.

templates/resources/notification_service.md.tmpl
```diff
@@ -0,0 +1,35 @@
...
+{{codefile "shell" .ImportFile }}
+
+In Terraform v1.5.0 and later, use an [`import` block](https://developer.hashicorp.com/terraform/language/import). For example:
+{{ $tfImportBlock := print "examples/resources/" .Name "/import.tf" }}{{tffile $tfImportBlock }}
+
+Imported OAuth-managed `linear` and `slack_workspace` services can manage their description, enabled state, and deletion. They cannot manage OAuth settings or notification filters, which do not apply to these integrations.
+
+Legacy `slack` notification services are not supported. Migrate them to a Slack Workspace notification service before managing them with Terraform.
+
+Buildkite does not return Datadog API keys, OpenTelemetry headers, or AWS account IDs. Those values remain unset when importing their services. Configure the matching provider settings after import if Terraform should manage an available setting; changing the AWS region or account ID replaces an EventBridge service.
```
petetomasik: "Buildkite does not return Datadog API keys, OpenTelemetry headers, or AWS account IDs" is only true for one of the three. From the notification services [API docs](https://buildkite.com/docs/apis/rest-api/organizations/notification-services#notification-service-data-model-settings-and-secret-handling):

- "Masked fields, such as the Datadog `api_key` and Amazon EventBridge `aws_account_id`, show only their final characters."
- "Slack incoming webhook URLs and encrypted OpenTelemetry headers are omitted."

So `api_key` and `aws_account_id` are returned, masked to something like `XXXXXXXXXXXXcdef`. Only the OpenTelemetry headers are genuinely withheld.

The schema descriptions in this PR get it right, which means the generated page contradicts itself:

- `resource_notification_service.go:301` says the AWS account ID is "masked by the API"
- `resource_notification_service.go:328` says the Datadog API key is "masked by the API"

The conclusion the sentence draws is still correct, but for a different reason: those values are unset after import because the provider deliberately discards the masked value rather than storing it (`preserveNotificationServiceString`, lines 1016-1024), not because the API withholds it. That distinction matters to a reader deciding whether they can trust state to reflect the real secret.

buildkite/resource_notification_service.go
```diff
@@ -0,0 +1,1088 @@
...
+			resp.Diagnostics.AddError("Unable to update notification service", err.Error())
+			return
+		}
+	}
+
+	if !plan.Enabled.Equal(state.Enabled) {
+		result, err = r.setEnabled(ctx, state.ID.ValueString(), plan.Enabled.ValueBool())
+		if err != nil {
+			resp.Diagnostics.AddError("Unable to change notification service enabled state", err.Error())
+			return
```
petetomasik, line 674: If the PATCH at line 577 succeeds and `setEnabled` then fails, this returns without ever calling `resp.State.Set`. The API now holds the updated settings, but Terraform state still has the old ones, so state and reality disagree until the next refresh reconciles them.

`Create` deliberately avoids this exact shape. At line 511 it writes state immediately after the create, before attempting the disable, precisely so the resource stays tracked when the second call fails, and `TestNotificationServiceCreateTracksResourceBeforeDisableError` pins that behavior. `Update` should agree with it.

It is self-healing rather than destructive, which is why this is a suggestion and not a blocker, but the asymmetry is easy to remove: persist the PATCH result before calling `setEnabled`, then report the toggle error.

```go
if result != nil {
    resp.Diagnostics.Append(plan.applyAPIResponse(ctx, result, state)...)
    resp.Diagnostics.Append(resp.State.Set(ctx, &plan)...)
}

if !plan.Enabled.Equal(state.Enabled) {
    result, err = r.setEnabled(ctx, state.ID.ValueString(), plan.Enabled.ValueBool())
    if err != nil {
        resp.Diagnostics.AddError("Unable to change notification service enabled state", err.Error())
        return
    }
}
```

buildkite/client.go
```diff
@@ -445,8 +445,8 @@ func (client *Client) makeRequest(ctx context.Context, method string, path strin
 		return fmt.Errorf("failed to create request: %w", err)
 	}
 
-	// Add content-type header for POST/PUT requests with body
-	if (method == http.MethodPost || method == http.MethodPut) && bodyBytes != nil {
+	// JSON request bodies use the same media type regardless of the HTTP method.
+	if bodyBytes != nil {
```
petetomasik, line 452: This silently changes the wire format for three other resources, and nothing exercises them. Broadening the guard from `(POST || PUT) && body != nil` to `body != nil` is right for the new PATCH, but it is not scoped to it. Four existing callers previously sent a JSON body with no `Content-Type` at all and now send one:

- `resource_pipeline.go:1556`
- `resource_pipeline.go:1613`
- `resource_test_suite.go:433`
- `resource_cluster_queue.go:831`

All four spell the method as the string `"PATCH"` rather than `http.MethodPatch`. Worth normalizing them to `http.MethodPatch`, since the new code uses the constant. GET and DELETE are unaffected: every one of those callers passes `nil` for `postData`, so `bodyBytes` stays a nil interface and the header is not set.

Those PATCHes demonstrably work today, so the API already parses unlabeled JSON, and sending the correct header should be a no-op or a latent-bug fix. But it is a behavioral change to three resources that this PR otherwise does not touch, the only new assertion is on the notification service's own PATCH (`resource_notification_service_test.go:659`), and nothing covers pipelines, test suites, or cluster queues.

buildkite/resource_notification_service_test.go
```diff
@@ -0,0 +1,740 @@
...
+}
+
+func notificationServiceTestResponse(providerType string) notificationServiceAPIResponse {
+	response := notificationServiceAPIResponse{
+		ID:                  notificationServiceTestID,
+		Enabled:             true,
+		Scope:               "all",
+		ScopeUUIDs:          []string{},
+		BranchConfiguration: "",
+		Settings:            json.RawMessage(`{}`),
```
petetomasik, line 1286: The two providers whose secrets the API masks have no plan/apply coverage. `notificationServiceTestResponse` only fills `settings` for `webhook` and `open_telemetry_tracing`. Datadog and EventBridge fall through to `json.RawMessage(`{}`)`, so every harness-driven test that sets those providers exercises an empty settings object.

That leaves the masked-secret preservation logic (`resource_notification_service.go:950-991`) covered only by direct calls to `applyAPIResponse` in `TestNotificationServiceApplyAPIResponsePreservesWriteOnlySettings`. That test is good as far as it goes, but calling the function directly cannot surface the failures that actually matter here, because they live in the framework's plan/state consistency checks rather than in the function's return value:

- `api_key` and `aws_account_id` are Optional + Computed + Sensitive, so a preserved value that does not exactly match the plan surfaces as "Provider produced inconsistent result after apply".
- `aws_account_id` additionally carries `RequiresReplace`, so a mismatch can plan a destroy and recreate of a live EventBridge partner source rather than just erroring.

It is worth closing anyway, because this is the highest-blast-radius behavior in the resource and it is the least tested. The separate `aws_account_id` blocker I filed is exactly the class of bug these tests would have caught.

buildkite/resource_notification_service.go
```diff
@@ -0,0 +1,1088 @@
...
+	}
+
+	err := r.delete(ctx, state.ID.ValueString())
+	if err != nil && !isAPIStatus(err, http.StatusNotFound) {
+		resp.Diagnostics.AddError("Unable to delete notification service", err.Error())
+	}
+}
+
+func (r *notificationServiceResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
+	if !notificationServiceUUIDPattern.MatchString(req.ID) {
```
petetomasik: Importing an unsupported provider type succeeds and writes state the schema rejects. `ImportState` validates only the UUID shape, and nothing downstream checks the provider the API actually returns. `applyAPIResponse` writes `m.ProviderType = types.StringValue(result.Provider.ID)` at line 882 with no case for `slack`, `event_log_api`, or anything else outside the `OneOf` list at lines 184-193, and `initializeImportedSettings` (line 916) has no default branch, so an unrecognized provider silently gets no settings model at all.

The user is then stuck, and the way they get unstuck is the problem. Writing `provider_type = "slack"` to match state fails the `OneOf` validator with "Attribute provider_type value must be one of", which says nothing about migrating. Guessing `provider_type = "webhook"` instead is accepted, and since `provider_type` carries `RequiresReplace`, the next plan destroys the live Slack notification service and creates a webhook in its place.

The docs already promise better than this. `templates/resources/notification_service.md.tmpl:32` says legacy Slack services "are not supported. Migrate them to a Slack Workspace notification service before managing them with Terraform." That guidance never reaches the user, because nothing fails at import time when it would be actionable.

This also future-proofs the resource: any provider Buildkite adds later hits the same trap otherwise. Worth a test using `api.setProvider("slack")`.

buildkite/resource_notification_service.go
```diff
@@ -0,0 +1,1088 @@
...
+			},
+		},
+	}
+}
+
+func notificationServiceOpenTelemetryTracingSchema() schema.SingleNestedAttribute {
+	return schema.SingleNestedAttribute{
+		Optional:            true,
+		MarkdownDescription: "Settings for the `open_telemetry_tracing` provider.",
+		Attributes: map[string]schema.Attribute{
```
petetomasik, line 363: `tracestate` is a fully manageable OTLP setting and it is missing here. The OpenTelemetry settings block exposes `endpoint`, `service_name`, `headers` and `resource_attributes`, but not `tracestate`. It is also absent from `notificationServiceOpenTelemetryTracingAPISettings` (lines 121-125), so it is neither read nor written.

The commit message frames the omissions as "ignoring API fields the organization cannot configure", which is accurate for `headers` but not for `tracestate`. It is gated per organization, with a dedicated 422 (`` `tracestate` isn't available for this organization ``), but that is the same shape as EventBridge's `include_build_meta_data`, which the docs also describe as available "when available for the organization" and which this resource does expose at line 310. So the resource is inconsistent: one org-gated setting is managed, the other is not.

Consequence today is mild, since PATCH leaves omitted settings alone and an existing `tracestate` survives. But an org with the feature cannot set or change it through Terraform, and a value configured in the UI is invisible with no drift signal.

buildkite/resource_notification_service.go
```diff
@@ -0,0 +1,1088 @@
...
+				Computed:            true,
+				MarkdownDescription: "How the webhook token authenticates deliveries.",
+				Validators: []validator.String{
+					stringvalidator.OneOf("token", "signature"),
+				},
+			},
+			"version": schema.Int64Attribute{
+				Optional:            true,
+				Computed:            true,
+				MarkdownDescription: "The webhook payload version. New services must use the API's latest version.",
```
petetomasik: Nit. The description covers the create behavior but not the update behavior. "New services must use the API's latest version" is accurate for create. The API also constrains updates, and that constraint is not mentioned anywhere in the schema or the docs page:

- On create, the submitted version must equal `LATEST_VERSION`, otherwise `` `version` must be N for a new webhook notification service ``.
- On update, it must equal either the service's current version (a no-op) or `LATEST_VERSION`, otherwise `` `version` can only be upgraded to N, not changed to M ``.

So the attribute is effectively write-once-plus-one-way-upgrade, not freely settable. Someone pinning `version` to an older value in config to hold a payload shape steady will get a 422 on the next apply, and the schema gives no hint that this is expected.

The impact is low. The failure is a clear, actionable API error rather than corruption or a silent no-op, and a config that simply omits `version` never trips it (the attribute is Optional + Computed, so Terraform proposes the prior state value and no PATCH is sent).

Perhaps word this as something like:
```suggestion
				MarkdownDescription: "The webhook payload version. New services must use the API's latest version, and existing services can only be upgraded to it, never downgraded.",
```

buildkite/resource_notification_service_test.go
```diff
@@ -0,0 +1,740 @@
...
+
+	resource.UnitTest(t, resource.TestCase{
+		ProtoV6ProviderFactories: protoV6ProviderFactories(),
+		CheckDestroy: func(*terraform.State) error {
+			if !api.isDeleted() {
+				return fmt.Errorf("notification service was not deleted")
+			}
+			return nil
+		},
+		Steps: []resource.TestStep{
```
petetomasik, line 270: Several lifecycle paths with real consequences are never exercised. The lifecycle test is solid for what it covers: create, disable-on-create, description PATCH, re-enable, import. What it does not cover are the paths where a mistake is expensive rather than merely wrong. None of these are blocking on their own. Grouping them because they share a cause: the suite tests the webhook happy path thoroughly and the other providers only through direct model-function calls.

#### Replacement on immutable EventBridge fields
`aws_region` and `aws_account_id` both carry `RequiresReplace` (lines 293-304). Nothing asserts what a change to either actually plans, so the destroy-and-recreate behaviour is entirely unpinned. This matters more than a normal coverage gap because replacing an EventBridge service tears down the AWS partner event source, and it is the mechanism behind the import blocker I filed separately. `plancheck` is already used elsewhere in this package, so the assertion is cheap:

```go
ConfigPlanChecks: resource.ConfigPlanChecks{
    PreApply: []plancheck.PlanCheck{
        plancheck.ExpectResourceAction("buildkite_notification_service.test",
            plancheck.ResourceActionDestroyBeforeCreate),
    },
},
```

See `resource_pipeline_schedule_test.go:242` and `resource_pipeline_test.go:1070` for the existing pattern.

#### `scope` and `scope_uuids` are never successfully applied
Grepping the file, the only Terraform config that mentions either is the OAuth rejection case at lines 279-283, which asserts they are refused. No test creates or updates a service with `scope = "some_projects"`, so neither the UUID validator at lines 225-229 nor the `scope_uuids` diffing in `updatePayload` (lines 664-672) has ever run in a plan. Given the API silently narrows or rejects scope combinations, this is the filter most worth pinning.

#### `enabled` going true to false on update
Line 171-192 covers false-on-create and then false to true. The reverse transition on an existing service is not covered. I ran it locally and it behaves correctly, including a clean no-op re-plan, so this is genuinely just a missing test rather than a latent bug.

buildkite/resource_notification_service.go
```diff
@@ -0,0 +1,1088 @@
...
+	}
+}
+
+func addNotificationServiceInt64(payload map[string]any, key string, value types.Int64) {
+	if !value.IsNull() && !value.IsUnknown() {
+		payload[key] = value.ValueInt64()
+	}
+}
+
+func addNotificationServiceBoolAny(payload map[string]any, key string, value types.Bool) {
```
petetomasik: Nit. Naming of this func is the odd one out in this helper family. Every sibling names the Terraform type it accepts and nothing else: `addNotificationServiceString`, `addNotificationServiceInt64`, `addNotificationServiceSet`, `addNotificationServiceMap`. This one appends `Any`, and so does its counterpart `addChangedNotificationServiceBoolAny` at line 825.

All five write into a `map[string]any`, so the suffix does not distinguish this function from the others. If anything it points the wrong way, since `Any` reads as "accepts any type" while the signature takes a `types.Bool` specifically.

buildkite/resource_notification_service.go
```diff
@@ -0,0 +1,1088 @@
...
+				Optional:            true,
+				Computed:            true,
+				ElementType:         types.StringType,
+				MarkdownDescription: "Additional OpenTelemetry resource attributes.",
+			},
+		},
+	}
+}
+
+func notificationServiceEventsAttribute(description string) schema.SetAttribute {
```
petetomasik: Nit. This func is a single-use helper with a parameter that only ever takes one value. This is called once, from the webhook schema at line 274, with a literal description. Every other attribute in that map is declared inline, so the indirection makes the webhook block the only one where you have to jump elsewhere to see what an attribute is.

Perhaps inline it instead?

```go
"events": schema.SetAttribute{
    Optional:            true,
    Computed:            true,
    ElementType:         types.StringType,
    MarkdownDescription: "The webhook event names to deliver.",
},
```

buildkite/resource_notification_service.go
```diff
@@ -0,0 +1,1088 @@
...
+const (
+	notificationServiceProviderWebhook                   = "webhook"
+	notificationServiceProviderAWSEventBridge            = "aws_event_bridge"
+	notificationServiceProviderDatadogPipelineVisibility = "datadog_pipeline_visibility"
+	notificationServiceProviderOpenTelemetryTracing      = "open_telemetry_tracing"
+	notificationServiceProviderLinear                    = "linear"
+	notificationServiceProviderSlackWorkspace            = "slack_workspace"
+)
+
+var notificationServiceUUIDPattern = regexp.MustCompile(`^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$`)
```
petetomasik: Nit/follow-up item. This represents a third UUID matcher in the package, and the only one named `Pattern`. Two things, both cosmetic.

Naming - Every other package-level compiled regex in `buildkite/` is suffixed `Regex`: `clusterUUIDRegex` (`data_source_cluster_network_ranges.go:20`), and `resourceNotFoundRegex`, `activeJobsRegex`, `alreadyExistsRegex`, `transientErrorRegex` (`util.go:19-22`). This is the sole `Pattern`. Rename to `notificationServiceUUIDRegex`, or drop the prefix entirely if it moves (below).

Duplication - There are now three UUID matchers with three different behaviours:

| | Case handling |
|---|---|
| `notificationServiceUUIDPattern` (line 35) | `[0-9a-fA-F]`, mixed case |
| `clusterUUIDRegex` (`data_source_cluster_network_ranges.go:20` - I'm guilty for adding this one 🤦 ) | `(?i)`, mixed case |
| `isUUID` (`util.go:165`) | `[a-f0-9]` only, lowercase |

The tidy version is a single shared `uuidRegex` with `(?i)` that this resource, `clusterUUIDRegex`, and `isUUID` all use, since the schema validator genuinely needs a `*regexp.Regexp` and cannot use a bool-returning helper. That is a wider cleanup than this PR should carry and should become a follow-up item for the consolidation if you think it is worth it.

buildkite/resource_notification_service.go
scadu: Addressed in 7affe6b

buildkite/resource_notification_service.go
scadu: @petetomasik `tracestate` support is indeed gated, but it's not public preview that we advertise now, so I decided to not include it for the time being.

buildkite/resource_notification_service.go
scadu: Addressed in c171c0f

templates/resources/notification_service.md.tmpl
scadu: Addressed in 65c6f5c

buildkite/resource_notification_service.go
scadu: Addressed in 5b885f7

buildkite/client.go
scadu: Addressed in d93017d. The existing callers now use `http.MethodPatch` as well.

buildkite/resource_notification_service.go
scadu: Fixed in 22bf9a0. Import now stops with migration guidance when Buildkite returns an unsupported provider.

buildkite/resource_notification_service_test.go
scadu: Covered both EventBridge replacement paths and scope creation, removal, and the following no-op plan. Removing scope UUIDs now sends an empty list.

buildkite/resource_notification_service_test.go
scadu: Covered both masked-secret paths. EventBridge and Datadog imports now adopt configured values in place and verify the following plan is empty.

buildkite/resource_notification_service.go
scadu: Dropped the `Any` suffix from both boolean helpers.

buildkite/resource_notification_service.go
scadu: Inlined the single-use events schema.

buildkite/resource_notification_service.go
scadu: Renamed the matcher to `notificationServiceUUIDRegex`. I left cross-resource consolidation out of this PR.

buildkite/resource_notification_service.go
```diff
@@ -0,0 +1,1129 @@
...
+		Optional:            true,
+		MarkdownDescription: "Settings for the `aws_event_bridge` provider.",
+		Attributes: map[string]schema.Attribute{
+			"aws_region": schema.StringAttribute{
+				Optional:            true,
+				Computed:            true,
+				MarkdownDescription: "The AWS region for the partner event source. Required on create and immutable.",
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
```
petetomasik: `aws_region`, `aws_account_id` (:308) and `event_source_name` (:318) are Computed with no `UseStateForUnknown()`. The framework marks a Computed attribute unknown in the plan whenever its config value is null and the plan otherwise differs from prior state, and it resolves paths inside a `SingleNestedAttribute`, so these nested attributes are in scope. `enabled` and `scope` escape it only because they carry defaults, and `id`/`created_at` because they already have the modifier.

Two consequences, both reproduced against this PR's own mock harness.

**The first plan after importing an EventBridge service fails.** Writing exactly what `templates/resources/notification_service.md.tmpl:34` instructs, settings block present with the account ID added and `aws_region` left to the API:

```hcl
aws_event_bridge = {
  aws_account_id = "123456789012"
}
```

`aws_region` is null in config so it plans as unknown; the `RequiresReplace()` below skips only when plan and state are equal, so an unknown triggers replacement; Terraform re-plans the create half with null prior state; and `ModifyPlan`'s create-only validation fires:

```
Error: Missing notification service setting

  with buildkite_notification_service.test,
  on terraform_plugin_test.tf line 18, in resource "buildkite_notification_service" "test":

aws_event_bridge.aws_region is required when creating this notification service.
```

The user is performing an update, and nothing in the docs says `aws_region` has to appear in config. That the create-validation happens to catch this is what makes it a loud error rather than a silent teardown of the AWS partner event source.

**`event_source_name` plans as "(known after apply)" on every update.** Confirmed with a plan that changes only `description`:

```go
plancheck.ExpectUnknownValue("buildkite_notification_service.test",
    tfjsonpath.New("aws_event_bridge").AtMapKey("event_source_name"))
```

`examples/resources/buildkite_notification_service/resource.tf:33-35` feeds that value into `aws_cloudwatch_event_bus.name` and `.event_source_name`, both ForceNew in the AWS provider. So editing a Buildkite service description plans a replacement of the customer's event bus, cascading to `aws_cloudwatch_event_rule.buildkite_builds` (ForceNew `event_bus_name`) and anything else bound to that bus outside this config. I reproduced the unknown value directly; the AWS-side cascade is standard unknown-into-ForceNew behaviour that I did not run.

Adding the modifier to all three fixes both symptoms, and `go test ./buildkite/ -run NotificationService` still passes with it applied:

```suggestion
				PlanModifiers: []planmodifier.String{
					stringplanmodifier.UseStateForUnknown(),
					stringplanmodifier.RequiresReplace(),
				},
```

Two related items while you are here. `graphql_id` (:178) has the same gap and goes unknown on every update, unlike `resource_cluster_queue.go:132-135`; that one is only cosmetic. And `RequiresReplaceIf` at :309 still fires on a genuinely unknown plan value, for instance an account ID sourced from a resource created in the same apply, so an early `if req.PlanValue.IsUnknown() { return }` closes the remaining case.

buildkite/resource_notification_service.go
```diff
@@ -0,0 +1,1129 @@
...
+			"scope": schema.StringAttribute{
+				Optional:            true,
+				Computed:            true,
+				MarkdownDescription: "Which resources the service applies to. Defaults to `all`. Only `all` is supported for `linear` and `slack_workspace` services.",
+				Default:             stringdefault.StaticString("all"),
+				Validators: []validator.String{
+					stringvalidator.OneOf("all", "some_projects", "some_teams", "some_clusters"),
+				},
+			},
+			"scope_uuids": schema.SetAttribute{
```
petetomasik, line 227: `scope_uuids` accepts mixed-case UUIDs but the API returns them lowercased, so an uppercase UUID in config fails every apply.

The validator here uses `notificationServiceUUIDRegex` (`:35`), which is `[0-9a-fA-F]` and so accepts `4D1F0B4E-...`. The API accepts mixed case too. But these values are stored in a UUID column and come back lowercased, and `applyAPIResponse:923-929` writes the response set straight into state. `scope_uuids` is Optional without `Computed`, so a case difference is a hard framework error rather than drift:

```
Error: Provider produced inconsistent result after apply

… .scope_uuids: planned set element cty.StringVal("4D1F0B4E-…") does not correlate
with any element in actual.
```

The service is created or updated remotely first, so the operator gets a "report this to the provider" error after the change has already landed. Anyone pasting a UUID out of a source that upcases them hits it every time.

Two options. Either normalize on the way in, which keeps the permissive validator:

```go
addNotificationServiceSet(ctx, settings, "scope_uuids", ...)  // lowercase each element first
```

or tighten the validator to lowercase only (`[0-9a-f]`) so the failure happens at plan time with an actionable message instead of after apply. I lean toward normalizing, since rejecting a UUID the API itself accepts is surprising.

Worth doing before release either way: tightening the validator later rejects configs that used to apply.

Separately, and much smaller: `scope = "all"` with `scope_uuids` set is rejected by the API with a clear message about needing a `some_*` scope, so that combination fails cleanly at apply rather than corrupting anything. A `ModifyPlan` check would still move it to plan time and save a round trip, but that is a UX nicety rather than a correctness issue. `scope = "some_projects"` with no `scope_uuids` is deliberately allowed by the API, so no validation is wanted there.

buildkite/resource_notification_service.go
```diff
@@ -0,0 +1,1129 @@
...
+		addChangedNotificationServiceString(settings, "endpoint", m.OpenTelemetryTracing.Endpoint, state.OpenTelemetryTracing.Endpoint)
+		addChangedNotificationServiceString(settings, "service_name", m.OpenTelemetryTracing.ServiceName, state.OpenTelemetryTracing.ServiceName)
+		addChangedNotificationServiceMap(ctx, settings, "headers", m.OpenTelemetryTracing.Headers, state.OpenTelemetryTracing.Headers, &diags)
+		addChangedNotificationServiceMap(ctx, settings, "resource_attributes", m.OpenTelemetryTracing.ResourceAttributes, state.OpenTelemetryTracing.ResourceAttributes, &diags)
+	}
+
+	return settings, diags
+}
+
+func addChangedNotificationServiceString(payload map[string]any, key string, plan, state types.String) {
```
petetomasik: Every `plan.IsNull()` branch in this helper family is unreachable for Optional+Computed attributes, which means those settings cannot be unset once written.

The branch here sends explicit JSON `null` to clear a value. For an attribute that is both `Optional` and `Computed`, a null config value causes Terraform to carry the prior state value into the proposed new state, so `plan` is either equal to `state` (returns at `:837`) or unknown (also returns at `:837`). The null branch only fires for Optional-without-Computed attributes.

Confirmed for `aws_account_id`: deleting the line as the only change produces a `no-op` plan, not a diff.

Net effect is that once set, none of these can be removed through Terraform, and deleting the attribute from config produces no plan diff at all:

- `webhook.token_mode` (`:262`), `webhook.tls_verify` (`:283`)
- `aws_event_bridge.include_build_meta_data` (`:322`)
- `datadog_pipeline_visibility.datadog_site` (`:341`)
- `open_telemetry_tracing.service_name` (`:364`), `headers` (`:369`), `resource_attributes` (`:377`)

Raising it now because dropping `Computed` later is a breaking state-shape change. The question is per-attribute: which of these does the API actually assign a value to when you omit it? `service_name` genuinely defaults to `buildkite`, and `token_mode`/`tls_verify` plausibly have API defaults, so `Computed` is right for those and the behaviour just needs documenting. `include_build_meta_data`, `datadog_site`, `headers`, and `resource_attributes` look more like plain optional settings, in which case dropping `Computed` makes them clearable and is better done before anyone has state.

Either way, the dead null branches in the five `addChanged*` helpers can go once the flags are settled, or stay if some attributes lose `Computed`.

buildkite/resource_notification_service.go
```diff
@@ -0,0 +1,1129 @@
...
+func (m *notificationServiceResourceModel) applyAPIResponse(ctx context.Context, result *notificationServiceAPIResponse, previous notificationServiceResourceModel) diag.Diagnostics {
+	imported := previous.ProviderType.IsNull()
+	if imported {
+		m.initializeImportedSettings(result.Provider.ID)
+	}
+
+	m.ID = types.StringValue(result.ID)
+	m.GraphQLID = types.StringPointerValue(result.GraphQLID)
+	m.ProviderType = types.StringValue(result.Provider.ID)
+	m.Description = types.StringPointerValue(result.Description)
```
petetomasik: `description` is missing the null-versus-empty-string guard that `branch_configuration` gets on the very next line. `description` is Optional without `Computed`, so the planned value is written verbatim and state has to match it exactly. Lines `:913-917` implement precisely the guard this shape needs for `branch_configuration`. `description` has none.

`buildkite/resource_cluster_secret.go:264-271` is the established pattern, with a comment naming the hazard:

```go
// Handle description - preserve null vs empty string
if secret.Description == nil {
	state.Description = types.StringNull()
} else if *secret.Description == "" && state.Description.IsNull() {
	state.Description = types.StringNull()
} else {
	state.Description = types.StringValue(*secret.Description)
}
```

Against this PR's own stub, a config with `description = ""` fails:

```
Error: Provider produced inconsistent result after apply

… .description: was cty.StringVal(""), but now null.
```

That failure is driven by `resource_notification_service_test.go:986` (`if api.description != ""`), which models the API as returning `null` for an empty description. So one of the two artifacts in this PR is wrong about the contract, and I could not determine which: does the API store a submitted `""` as `""`, or coerce it to `null`? A single `POST` with `"description": ""` against a real org settles it. You almost certainly know the answer already.

A three-branch guard would be correct:

```suggestion
	if result.Description == nil {
		m.Description = types.StringNull()
	} else if *result.Description == "" && m.Description.IsNull() {
		m.Description = types.StringNull()
	} else {
		m.Description = types.StringValue(*result.Description)
	}
```

Same exposure exists unguarded for `datadog_tags` (`:347`, mapped at `:1030`), `webhook.url` (`:252`), and `open_telemetry_tracing.endpoint` (`:355`) if the API normalizes any of them, for instance a trailing slash on a URL or whitespace in a tag list. Those are all Optional-without-Computed and written straight from the response.

buildkite/resource_notification_service.go
```diff
@@ -0,0 +1,1129 @@
...
+		return notificationServiceSettingsDiagnostic(notificationServiceProviderAWSEventBridge, err)
+	}
+
+	previousAccountID := types.StringNull()
+	if previous.AWSEventBridge != nil {
+		previousAccountID = previous.AWSEventBridge.AWSAccountID
+	}
+	accountID := preserveNotificationServiceString(m.AWSEventBridge.AWSAccountID, previousAccountID)
+	var diags diag.Diagnostics
+	if settings.AWSAccountID != nil && !accountID.IsNull() && !accountID.IsUnknown() {
```
petetomasik, line 1087: This check fails closed on a guess about the mask format, and it duplicates a server-side guard that is already stricter.

`strings.TrimLeft(*settings.AWSAccountID, "X")` returning `""` is treated as a mismatch, so a fully masked response errors, as does any mask not built from `X`. Because this runs inside `Read`, the symptom is `terraform plan` and `terraform refresh` erroring on an already-managed resource with no way forward except editing config, and per the plan-modifier thread, removing `aws_account_id` to escape is itself hazardous.

What makes me want to fail open here is that the client-side check is not what protects the resource. `aws_account_id` is create-only and enforced as such by the API: a genuine change is rejected with a message saying it cannot be changed after creation, while resending either the stored value or the mask is accepted as a no-op. So the server already refuses the dangerous case, and this check's only unique contribution is turning an unrecognized mask format into a refresh failure.

Suggested shape, keeping the hard error only for a suffix that is present and genuinely does not match:

```go
if maskedSuffix == "" {
	// Fully masked, or a mask format we do not recognise. Nothing to compare against.
	return diags
}
```

In the other direction, `strings.HasSuffix` accepts any 12-digit account ending in the same final digits, so a typo is adopted into state as truth and, the field being create-only, never corrected. Not fixable in code, but `templates/resources/notification_service.md.tmpl:34` currently says adding the account ID "adopts it without replacing the service" without noting that only the trailing digits are checked. Worth a clause.

Minor, same block: `accountID = previousAccountID` at `:1001` runs after `AddError`, and every caller of `applyAPIResponse` discards state when diagnostics carry an error, so it is dead.

buildkite/resource_notification_service.go
```diff
@@ -0,0 +1,1129 @@
...
+						"Changing a known AWS account ID replaces the notification service. Setting it after import adopts the existing masked account ID.",
+						"Changing a known AWS account ID replaces the notification service. Setting it after import adopts the existing masked account ID.",
+					),
+				},
+			},
+			"event_source_name": schema.StringAttribute{
+				Computed:            true,
+				MarkdownDescription: "The AWS partner event source created for this service.",
+			},
+			"include_build_meta_data": schema.StringAttribute{
```
petetomasik, line 327: `include_build_meta_data` is `Computed` but the API assigns it no default, so once set it can never be cleared.

Every `plan.IsNull()` branch in the `addChanged*` helpers (`:836-900`) is unreachable for an Optional+Computed attribute: a null config value makes Terraform carry the prior value into the proposed new state, so `plan` either equals `state` or is unknown, and both return early at `:837`. The null branch that would send JSON `null` to clear the value only ever fires for Optional-without-Computed attributes.

Confirmed for `aws_account_id`: deleting the line as the only change produces a `no-op` plan rather than a diff.

For most of the settings this is harmless, because the API really does assign a default and `Computed` is the right flag. Those just need the behaviour documented, since deleting the attribute from config produces no diff at all rather than reverting to the default:

- `webhook.version`, `token`, `token_mode`, `events`, `tls_verify` all have API defaults
- `datadog_pipeline_visibility.datadog_site` defaults to `datadoghq.com`
- `open_telemetry_tracing.service_name` defaults to `buildkite`, and `headers` / `resource_attributes` default to empty maps
- `aws_event_bridge.event_source_name` is API-generated, and `aws_region` / `aws_account_id` are create-only, so `Computed` earns its place there for import adoption

`include_build_meta_data` (`:322`) is the odd one out. The API gives it no default, so it is a plain optional setting, and marking it `Computed` means a build meta-data filter can be set once and then never removed through Terraform. Dropping `Computed` makes it clearable:

```suggestion
			"include_build_meta_data": schema.StringAttribute{
				Optional:            true,
				MarkdownDescription: "A build meta-data pattern to include in EventBridge payloads.",
			},
```

Raising it now because dropping `Computed` later is a breaking state-shape change. `datadog_tags` is already Optional-only, which is the right call and matches this.

templates/resources/notification_service.md.tmpl
```diff
@@ -0,0 +1,35 @@
...
+{{codefile "shell" .ImportFile }}
+
+In Terraform v1.5.0 and later, use an [`import` block](https://developer.hashicorp.com/terraform/language/import). For example:
+{{ $tfImportBlock := print "examples/resources/" .Name "/import.tf" }}{{tffile $tfImportBlock }}
+
+Imported OAuth-managed `linear` and `slack_workspace` services can manage their description, enabled state, and deletion. They cannot manage OAuth settings or notification filters, which do not apply to these integrations.
+
+Legacy `slack` notification services are not supported. Migrate them to a Slack Workspace notification service before managing them with Terraform.
+
+Imported non-OAuth services still need their matching settings block. Buildkite masks Datadog API keys and AWS account IDs, and omits OpenTelemetry headers. Terraform does not store the masked values, so all three remain unset after import. Add a secret value if Terraform should manage it. Adding the matching AWS account ID after import adopts it without replacing the service. Changing a known AWS region or account ID replaces the EventBridge service.
```
petetomasik: Two gaps in this paragraph, both of which cost an operator real money if they follow it as written.

**`enabled`, `scope`, and `scope_uuids` are not mentioned, and omitting them reverts the remote service.** Those carry framework defaults (`resource_notification_service.go:211` and `:218`), and framework defaults are applied on update, not just create, so a null config value beats prior state.

Confirmed: importing a service the API reports as `enabled: false, scope: "some_projects"`, with a config that omits both, plans an Update and applies `enabled = true`, `scope = "all"`.

So importing an intentionally disabled EventBridge or Datadog service and applying starts it delivering, and importing a project-scoped webhook widens it to every pipeline in the org. The API's enable endpoint also clears a previous disabled or broken state, so Terraform can silently un-break a service someone disabled on purpose. The diff does show in the plan, which keeps this out of blocker territory, but import is the only supported path for `linear` and `slack_workspace` and the recommended one for EventBridge and Datadog, and this paragraph is where a reader looks.

**The Datadog `api_key` sentence describes the wrong behaviour.** "Add a secret value if Terraform should manage it" is followed only by the AWS-specific note about adoption, which reads as though all three secrets adopt. They do not. The API treats a resubmitted *mask* as a no-op, but the provider never sends the mask: it discards the masked value and sends whatever is in config. So for Datadog, adding `api_key` after import overwrites the live key, as `TestNotificationServiceDatadogLifecycle` shows with its `{"settings":{"api_key":"supersecret"}}` PATCH. Same for OpenTelemetry `headers`. A user who follows this guidance with a placeholder silently breaks their Datadog ingestion.

Suggested rewrite:

```suggestion
Imported non-OAuth services still need their matching settings block. Buildkite masks Datadog API keys and AWS account IDs, and omits OpenTelemetry headers. Terraform does not store the masked values, so all three remain unset after import.

Also declare `enabled`, `scope`, and `scope_uuids` to match the existing service. These attributes default to `true` and `all`, and the defaults take precedence over the imported values, so omitting them makes the first apply enable a disabled service and widen a scoped one to the whole organization.

Adding the matching AWS account ID after import adopts it without replacing the service; only its trailing digits are checked against the masked value. Setting a Datadog API key or OpenTelemetry headers after import overwrites the stored secret rather than adopting it, so supply the real value or leave the attribute unset. Changing a known AWS region or account ID replaces the EventBridge service.
```

buildkite/resource_notification_service.go
```diff
@@ -0,0 +1,1129 @@
...
+			Terraform meta-argument.
+
+			Webhook and Slack Workspace notification services are available on every plan.
+			New Amazon EventBridge, Datadog Pipeline Visibility, OpenTelemetry control-plane
+			tracing, and Linear notification services require Enterprise entitlements. Existing
+			organizations may retain access based on their entitlements. Terraform cannot check
+			these entitlements during planning; the Buildkite API rejects unavailable services
+			during apply with an upgrade error.
+
+			Secret settings that the Buildkite API masks or omits are preserved from Terraform
```
petetomasik, line 169: This paragraph reads as though masked secrets are the only thing a refresh cannot see, and there is a larger blind spot beside them.

`build_states` is a first-class top-level field on the notification service object, accepted on create and update and returned on every read, covering `build_passed`, `build_fixed`, `build_failed`, `build_blocked`, `build_canceled`, `build_failing`, and `job_activated`. It controls which states actually trigger a notification, and the API exposes it even for providers that ignore it so a read-then-update round trip is lossless.

This resource neither reads nor writes it, and `resource_notification_service_test.go:640-647` pins it as an unsupported argument, so I read the omission as a deliberate scope call rather than an oversight. Not sending it is safe, since an update leaves it alone.

It is invisible to users though, and `grep` across `docs/`, `templates/`, and `examples/` returns nothing. Two consequences: a service this resource creates always takes the API defaults with no way to change them, and a service whose build states were customised in the web UI keeps them with no drift signal, so two identical Terraform configs can deliver different notifications depending on how the service came into existence.

It matters most for the import-only providers. For `linear` and `slack_workspace` the settings are unmanaged and notification filters are rejected in `ModifyPlan`, which leaves `build_states` as the only substantive knob, and `templates/resources/notification_service.md.tmpl:30` lists what those services "cannot manage" without mentioning it.

Adding a `build_states` block later is purely additive, so the modelling can wait. The documentation gap should not:

```suggestion
			Secret settings that the Buildkite API masks or omits are preserved from Terraform
			state during refresh. Changes made to those values outside Terraform cannot be detected.

			This resource does not manage build state selection. Services it creates use the
			API defaults, and build states changed in the Buildkite interface are not detected
			as drift.
```

Worth a short comment beside the pinning test as well, so the next person does not read the rejection as a bug.

buildkite/resource_notification_service.go
scadu: Addressed in 9912396

buildkite/resource_notification_service.go
scadu: Addressed in 6c5224e and d71508b

buildkite/resource_notification_service.go
scadu: Addressed in 9358fb8

petetomasik (approved):

PR #1181, opened 2026-07-10
feat(pipeline): add build_pull_request_edited provider setting

gempesaw:
# context

hey folks, i saw the blog post that [pipelines support triggering when a PR's title or description is edited](https://buildkite.com/resources/changelog/352-trigger-pipelines-from-more-github-events/). we'd like to control that from our IaC!

it does seem to already be in the schema, but i didn't see support for it in the terraform provider https://github.com/buildkite/terraform-provider-buildkite/blob/600205e1623d71f0261b71e64e60cf592e901223/schema.graphql#L9359-L9362

so this PR attempts to wire the requisite support for `build_pull_request_edited`

# testing

i manually ran the acceptance test that sets the edited flag against a test org, and it passed 🎉 ! ( i couldn't run that whole file cuz other tests were complaining about `cluster` in my test org, which i believe is unrelated to my change?)

```shell
TF_ACC=1 go test ./buildkite/ -v -run \
  'TestAccBuildkitePipelineResource/create_pipeline_setting_all_attributes'
```

image

PR #1210, opened 2026-08-14
Add more GitHub trigger settings to pipeline provider_settings

dattallant:
Adds the remaining GitHub / GitHub Enterprise [trigger settings](https://buildkite.com/docs/apis/rest-api/pipelines) that can be set in the UI and REST API but not yet in `provider_settings`:

- `build_pull_request_review_comment_created`
- `review_comment_command_word`
- `review_comment_match_mode` (`exact` / `contains`, like `issue_comment_match_mode`)
- `build_pull_request_dequeued`
- `build_pull_request_reopened`

They are written through the existing REST `provider_settings` PATCH and read back through the GraphQL fragment like their siblings; `generated.go` is regenerated from the vendored schema, no `schema.graphql` change. Left out: `build_pull_request_edited` (covered by #1181) and `build_pull_request_stacks` (accepted by REST but not in the vendored schema yet, so it can't be read back without a schema refresh).

Extended the "setting all attributes" acceptance subtests and `TestMapProviderSettingsFromGraphQLGitHub`, and ran create + refresh + import locally against a test organization; the update subtest starts from a pipeline without `cluster_id`, which my organization rejects, so I couldn't run that one.

## PR checklist:
- [x] `docs/` updated
- [x] tests added

### DIFF
--- a/buildkite/resource_pipeline.go
+++ b/buildkite/resource_pipeline.go
@@ -1079,7 +1079,7 @@ func (*pipelineResource) Schema(ctx context.Context, req resource.SchemaRequest,
 					"build_pull_request_review_comment_created": schema.BoolAttribute{
 						Computed:            true,
 						Optional:            true,
-						MarkdownDescription: "Whether to create builds when an inline review comment is created on a pull request.",
+						MarkdownDescription: "Whether to create builds when an inline review comment is created on a pull request. Requires `build_pull_requests` to be enabled.",
 						PlanModifiers: []planmodifier.Bool{
 							boolplanmodifier.UseNonNullStateForUnknown(),
 						},
@@ -1106,7 +1106,7 @@ func (*pipelineResource) Schema(ctx context.Context, req resource.SchemaRequest,
 					"build_pull_request_dequeued": schema.BoolAttribute{
 						Computed:            true,
 						Optional:            true,
-						MarkdownDescription: "Whether to create builds when a pull request is removed from a merge queue.",
+						MarkdownDescription: "Whether to create builds when a pull request is removed from a merge queue. Requires `build_pull_requests` to be enabled.",
 						PlanModifiers: []planmodifier.Bool{
 							boolplanmodifier.UseNonNullStateForUnknown(),
 						},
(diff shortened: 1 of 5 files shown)


buildkite/resource_pipeline.go
```diff
@@ -1071,6 +1076,49 @@ func (*pipelineResource) Schema(ctx context.Context, req resource.SchemaRequest,
...
+							stringplanmodifier.UseNonNullStateForUnknown(),
+						},
+						Validators: []validator.String{
+							stringvalidator.OneOf("exact", "contains"),
+						},
+					},
+					"build_pull_request_dequeued": schema.BoolAttribute{
+						Computed:            true,
+						Optional:            true,
+						MarkdownDescription: "Whether to create builds when a pull request is removed from a merge queue.",
```
PriyaSudip: This also needs build_pull_requests to be "true". Can we update the description?

buildkite/resource_pipeline.go
```diff
@@ -1071,6 +1076,49 @@ func (*pipelineResource) Schema(ctx context.Context, req resource.SchemaRequest,
...
+						PlanModifiers: []planmodifier.Bool{
+							boolplanmodifier.UseNonNullStateForUnknown(),
+						},
+					},
+					"build_pull_request_reopened": schema.BoolAttribute{
+						Computed:            true,
+						Optional:            true,
+						MarkdownDescription: "Whether to create builds when a pull request is reopened.",
```
PriyaSudip: Similarly, requires build_pull_requests to be true.

buildkite/resource_pipeline_test.go
PriyaSudip, line 879: Can we also include the coverage for build_pull_request_review_comment_created and build_pull_request_reopened?

PriyaSudip (commented):
Nice work @dattallant , left few comments for your review :)

buildkite/resource_pipeline.go
dattallant: both descriptions now say they need `build_pull_requests` enabled, and I added the same note to `build_pull_request_review_comment_created` since it's a pull request event too (the API happily stores all three either way, they just don't fire without PR builds).

buildkite/resource_pipeline.go
dattallant: same as above

buildkite/resource_pipeline_test.go
dattallant: done - `build_pull_request_review_comment_created` and `build_pull_request_reopened` are now asserted in the refresh step there and in `TestMapProviderSettingsFromGraphQLGitHub`

PriyaSudip (approved):
Thanks for the changes, looks good now.

PR #1211, opened 2026-08-14
Add API access token settings to buildkite_organization

dattallant:
Adds the two [API access token settings](https://buildkite.com/docs/apis/rest-api/organizations/api-settings) to `buildkite_organization`, so restricting token creation to admins and revoking inactive tokens can be declared and drift-detected instead of set in the UI. Follow-up to #1207.

```hcl
resource "buildkite_organization" "settings" {
  restrict_user_api_token_creation = true
  revoke_inactive_tokens_after     = "DAYS_90"
}
```

Both attributes are optional and computed: when omitted the current value is only read into state, so upgrading the provider never changes an organization, and only values that differ are sent in the PATCH. `revoke_inactive_tokens_after` mirrors the `RevokeInactiveTokenPeriod` enum (`NEVER`, `DAYS_30` … `DAYS_365`) and is translated to the endpoint's nullable `revoke_inactive_tokens_after_days`; happy to switch to the REST integer if you prefer. The endpoint needs the `read_organization_settings`/`write_organization_settings` scopes, so a 403 on read only warns and keeps the last known values (as `retry_agent_affinity` does), and changing the revocation period on a plan without that feature returns the API error plus a one-line hint. When the api-settings read is forbidden but the write isn't, every configured value is sent rather than compared, so an explicit `false`/`NEVER` still applies.

Added table tests for the enum mapping and the PATCH payload, extended `TestAccBuildkiteOrganizationResource` (create, update, import, attribute removed, reset) and ran it locally against a test organization.

## PR checklist:
- [x] `docs/` updated
- [x] tests added

### DIFF
--- a/buildkite/resource_organization.go
+++ b/buildkite/resource_organization.go
@@ -4,24 +4,30 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"net/http"
 	"strings"
 
 	"github.com/MakeNowJust/heredoc"
+	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
 	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/booldefault"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/boolplanmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
 	"github.com/hashicorp/terraform-plugin-framework/types"
 )
 
 type organizationResourceModel struct {
-	AllowedApiIpAddresses types.List   `tfsdk:"allowed_api_ip_addresses"`
-	ID                    types.String `tfsdk:"id"`
-	UUID                  types.String `tfsdk:"uuid"`
-	Enforce2FA            types.Bool   `tfsdk:"enforce_2fa"`
+	AllowedApiIpAddresses        types.List   `tfsdk:"allowed_api_ip_addresses"`
+	ID                           types.String `tfsdk:"id"`
+	UUID                         types.String `tfsdk:"uuid"`
+	Enforce2FA                   types.Bool   `tfsdk:"enforce_2fa"`
+	RevokeInactiveTokensAfter    types.String `tfsdk:"revoke_inactive_tokens_after"`
+	RestrictUserApiTokenCreation types.Bool   `tfsdk:"restrict_user_api_token_creation"`
 }
 
 type organizationResource struct {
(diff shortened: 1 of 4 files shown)


buildkite/resource_organization.go
```diff
@@ -257,4 +301,135 @@ func (o *organizationResource) Delete(ctx context.Context, req resource.DeleteRe
...
+func (o *organizationResource) updateAPISettings(ctx context.Context, plan, prior, state *organizationResourceModel, diags *diag.Diagnostics) {
+	current, err := o.client.getOrganizationAPISettings(ctx)
+	readable := err == nil
+	if err != nil {
+		if !strings.Contains(err.Error(), "status: 403") {
+			diags.AddError("Unable to read organization API settings", fmt.Sprintf("Unable to read organization API settings: %s", err.Error()))
+			return
+		}
+		// without the read scope, compare against the previous state instead
+		current = apiSettingsFromModel(prior)
```
omehegan, line 426: When we create, `prior` will be nil. I think if this returns a 403, `apiSettingsFromModel(nil)` substitutes the zero values `NEVER` and `false`. That can cause an explicitly configured value to be silently skipped. Can we distinguish between a known value and a fallback? If the read fails, we might want to include all explicitly configured values in the PATCH, so we're not incorrectly assuming things based on the defaults. E.g.:

```go
currentKnown := err == nil
if isAPIStatus(err, http.StatusForbidden) {
    current = apiSettingsFromModel(prior)
}

payload := apiSettingsPatch(config, plan, current, currentKnown)
```

and then:

```go
if revokeConfigured &&
    !plan.RevokeInactiveTokensAfter.IsUnknown() &&
    (!currentKnown ||
        plan.RevokeInactiveTokensAfter.ValueString() !=
            revokePeriodFromDays(current.RevokeInactiveTokensAfterDays)) {
    payload["revoke_inactive_tokens_after_days"] =
        revokePeriodToDays(plan.RevokeInactiveTokensAfter.ValueString())
}
```

The same would apply to `restrict_user_api_token_creation`. This ensures an explicit `false` or `NEVER` is still enforced when the provider has write scope but cannot read the current value.

It would also be useful to add unit cases for “unknown current value + explicit `false`” and “unknown current value + explicit `NEVER`”.

It may also be worth emitting the same warning here as `readAPISettings`; otherwise the fallback is invisible to the user.

buildkite/resource_organization.go
```diff
@@ -257,4 +301,135 @@ func (o *organizationResource) Delete(ctx context.Context, req resource.DeleteRe
...
+		settings.RestrictUserApiTokenCreation = model.RestrictUserApiTokenCreation.ValueBool()
+	}
+	return settings
+}
+
+// apiSettingsPatch returns the configured api-settings that differ from current
+func apiSettingsPatch(plan *organizationResourceModel, current *organizationAPISettings) map[string]any {
+	payload := map[string]any{}
+	revoke := plan.RevokeInactiveTokensAfter
+	if !revoke.IsNull() && !revoke.IsUnknown() && revoke.ValueString() != revokePeriodFromDays(current.RevokeInactiveTokensAfterDays) {
```
omehegan: Since this attribute is `Optional` and `Computed` with `UseStateForUnknown`, a non-null plan value does not necessarily mean it was configured, it may have been copied from state. Could we inspect `req.Config` and use that to decide whether the setting is managed, while continuing to use `req.Plan` for its desired value? Otherwise an unrelated update can PATCH a stale computed value for an attribute the user omitted.

Same at line 378.

buildkite/resource_organization_test.go
```diff
@@ -168,8 +279,168 @@ func TestAccBuildkiteOrganizationResource(t *testing.T) {
...
+					Check: resource.ComposeAggregateTestCheckFunc(
+						resource.TestCheckResourceAttr("buildkite_organization.let_them_in", "restrict_user_api_token_creation", "false"),
+						testAccCheckOrganizationAPISettingsRemoteValues("NEVER", false),
+					),
+				},
+			},
+		})
+	})
+
+	t.Run("manages inactive API token revocation", func(t *testing.T) {
```
omehegan, line 476: This API call happens before `resource.Test`, so it runs even when `TF_ACC` is unset. Could we move the entitlement probe into `PreCheck`? `resource.Test` will then perform its normal `TF_ACC` gating before any live API call.

buildkite/resource_organization.go
```diff
@@ -257,4 +301,135 @@ func (o *organizationResource) Delete(ctx context.Context, req resource.DeleteRe
...
+	if !restrict.IsNull() && !restrict.IsUnknown() && restrict.ValueBool() != current.RestrictUserApiTokenCreation {
+		payload["restrict_user_api_token_creation"] = restrict.ValueBool()
+	}
+	return payload
+}
+
+func (o *organizationResource) readAPISettings(ctx context.Context, state *organizationResourceModel, diags *diag.Diagnostics) {
+	settings, err := o.client.getOrganizationAPISettings(ctx)
+	if err != nil {
+		if !strings.Contains(err.Error(), "status: 403") {
```
omehegan: I think you could use `http.StatusForbidden` here instead of the string `status: 403` to avoid the chance of matching on some other API response string. Same with line 404 below.

omehegan (changes requested):
In addition to the inline comments, now that we've merged #1207 you'll want to rebase on main.

buildkite/resource_organization.go
dattallant: fair point.  `updateAPISettings` will look at whether the GET worked; when it didn't, every configured value is sent regardless of the comparison and the same "keeping the last known values" warning is raised. The `TestApiSettingsPatch` has "unknown current and explicit false" / "explicit never" cases.

buildkite/resource_organization.go
dattallant: yeah agreed. `Create/Update` now read req.Config and apiSettingsPatch takes config for "is it managed" and plan for the value

buildkite/resource_organization_test.go
dattallant: moved into the step's PreCheck after `testAccPreCheck`

buildkite/resource_organization.go
dattallant: updated

omehegan (approved):

PR #1220, opened 2026-08-19
Add github_webhooks_enabled to buildkite_pipeline

dattallant:
Adds `github_webhooks_enabled` so pausing GitHub webhook processing for a pipeline (the [pause/resume GitHub webhooks](https://buildkite.com/docs/apis/rest-api/pipelines) endpoints, `DELETE`/`PUT …/pipelines/{slug}/github-webhooks`) can be declared instead of toggled in the UI:

```hcl
resource "buildkite_pipeline" "pipeline" {
  name                    = "example"
  repository              = "https://github.com/my-org/example.git"
  github_webhooks_enabled = false
}
```

It is a top-level optional attribute rather than part of `provider_settings` because it isn't a `provider_settings` key in the REST API. It is only written when the value changes and only read (via `GET …/github-webhooks`) while it is configured, so pipelines that don't set it make no extra calls.

I used the small REST endpoint instead of `githubWebhooksDisabled` in GraphQL to keep it out of the provider settings subtree that read-only queries conditionally. It only applies to GitHub / GitHub Enterprise repositories on organizations enrolled in the newer webhook triggers. 

Happy to make it computed and always read if you'd rather see drift for unmanaged pipelines.

Added an acceptance subtest (create disabled, enable, import) and ran it locally against a test organization.

## PR checklist:
- [x] `docs/` updated
- [x] tests added
- [ ] an example added to `examples/` <-- snippet above; can add to examples/ if useful


scadu (approved):
Looks good. Thanks for the contribution!

PR #1221, opened 2026-08-19
Add expires_at to buildkite_cluster_agent_token

dattallant:
`clusterAgentTokenCreate` accepts an `expiresAt` and `ClusterToken` exposes it, but tokens created through the provider could only ever be non-expiring. This adds an optional `expires_at` (RFC 3339) so short-lived agent tokens can be minted from Terraform:

```hcl
resource "buildkite_cluster_agent_token" "expiring" {
  cluster_id  = buildkite_cluster.primary.id
  description = "Token that expires"
  expires_at  = "2030-01-01T00:00:00Z"
}
```

There is no update input for the expiry, so changing or removing it replaces the token (`RequiresReplace`). The value is validated as RFC 3339 at plan time, the variable is omitted from the mutation when unset (the API rejects an explicit null), and on read the configured string is kept when it names the same instant as the API value so a different offset spelling doesn't show as drift. `generated.go` is regenerated from the vendored schema for the new fragment field and mutation variable; no `schema.graphql` change.

Added table tests for the validator and the read mapping plus an acceptance subtest (create with an expiry, then change it and expect a replace), and ran `TestAccBuildkiteClusterAgentTokenResource` locally against a test organization.

## PR checklist:
- [x] `docs/` updated
- [x] tests added


buildkite/resource_cluster_agent_token.go
```diff
@@ -293,3 +316,33 @@ func (ct *clusterAgentToken) Delete(ctx context.Context, req resource.DeleteRequ
...
+
+// expiresAtFromAPI keeps the configured value when it names the same instant as the API value
+func expiresAtFromAPI(remote *time.Time, current types.String) types.String {
+	if remote == nil {
+		return types.StringNull()
+	}
+	if configured, err := time.Parse(time.RFC3339, current.ValueString()); err == nil && configured.Equal(*remote) {
+		return current
+	}
+	return types.StringValue(remote.UTC().Format(time.RFC3339))
```
scadu: Important: this drops fractional seconds, even though `expires_at` accepts them. The API stores six digits, so higher-precision values never converge and Terraform keeps replacing the token. Could we reject higher precision and use `RFC3339Nano` when reading the API value? A round-trip test should catch this.

scadu (changes requested):

buildkite/resource_cluster_agent_token.go
dattallant: good catch, thanks! the read now formats with RFC3339Nano (keeping the "configured spelling wins if it's the same instant" bit) and the validator rejects anything finer than the API hands back; one wrinkle: through GraphQL expiresAt comes back with milliseconds (a .123456Z expiry reads back as .123Z), so I capped it at millisecond rather than microsecond precision - happy to bump it to six digits if the field starts returning them. Added round-trip cases for Z / +00:00 / .5Z / .123Z plus the rejection

scadu: Ah, indeed, the database stores six fractional digits, but GraphQL only returns three. I verified this against the API: `.123456Z` came back as `.123Z`. 
Capping the provider at millisecond precision is the right fix, thanks!

scadu (approved):

PR #1225, opened 2026-08-21
SUP-7780 Enable terraform support for org level pipeline default settings

dahtey-bk:
This PR relates to https://linear.app/buildkite/issue/SUP-7780/add-organization-pipeline-settings-resource
This is to ensure the ability to manage and modify org-level pipeline default settings via Terraform (Buildkite REST API already supports it so this closes the gap).

## Changes
- New resource in `buildkite/resource_organization_pipeline_settings.go`, registered in `provider.go`
- Docs, a usage example, and an import example
- Unit tests, acceptance tests, and a lifecycle test against a fake API

## Example
```hcl
resource "buildkite_cluster" "primary" {
  name = "Primary"
}
resource "buildkite_organization_pipeline_settings" "settings" {
  # defaults that new pipelines are created with
  default_branch     = "main"
  default_cluster_id = buildkite_cluster.primary.uuid
  # time jobs out after an hour, and stop any pipeline asking for more than two
  default_timeout_in_minutes = 60
  maximum_timeout_in_minutes = 120
  # expire scheduled jobs that have not started within a day (60 to 43200)
  scheduled_job_expiry_in_minutes = 1440
  # keep every pipeline private, and leave hosted agents reachable for debugging
  public_pipeline_creation_enabled      = false
  hosted_agents_terminal_access_enabled = true
  # export build data to a bucket, named without a scheme; configured as a pair
  build_export_location    = "my-build-data-bucket"
  build_export_strategy_id = "s3"
}
# read-only attributes, useful because build export is plan-gated
output "build_export_available" {
  value = buildkite_organization_pipeline_settings.settings.build_export_available
}
output "build_export_supported_strategies" {
  value = buildkite_organization_pipeline_settings.settings.build_export_supported_strategies
}
```
Every attribute is optional. Omitting one reads the organization's current value into
state and leaves it alone, so a minimal configuration that manages only the default branch
is valid:
```hcl
resource "buildkite_organization_pipeline_settings" "settings" {
  default_branch = "main"
}
```
Settings that can be emptied are cleared with an empty string rather than by removing the
attribute, since removal means "stop managing this":
```hcl
resource "buildkite_organization_pipeline_settings" "settings" {
  default_cluster_id = "" # new pipelines get no default cluster
  build_export_location    = "" # stop exporting build data
  build_export_strategy_id = ""
}
```
The organization already has these settings, so adopt them with an import rather than
letting Terraform treat them as new. The ID is the organization slug:
```hcl
import {
  to = buildkite_organization_pipeline_settings.settings
  id = "my-buildkite-organization"
}
```

## Notes
- **Emptiable settings are cleared with `""`**, not by removal: `default_cluster_id`, and
  `build_export_location` with `build_export_strategy_id` (`RequiredTogether`). The job
  timeouts have no empty value, so they can be changed here but only removed in the UI.
- **Advanced queue metrics are excluded** — the API can only turn them on, so neither a
  destroy nor a rollback could undo it.
- Requires an organization administrator and the `read_organization_settings` /
  `write_organization_settings` scopes.


buildkite/resource_organization_pipeline_settings_test.go
```diff
@@ -0,0 +1,448 @@
...
+				}
+			},
+			ProtoV6ProviderFactories: protoV6ProviderFactories(),
+			CheckDestroy:             testCheckOrganizationPipelineSettingsRemain,
+			Steps: []resource.TestStep{
+				{
+					Config: config(fmt.Sprintf(`
+						build_export_location    = %q
+						build_export_strategy_id = "s3"
+					`, bucket)),
```
scadu: CI fails here because this constructs an S3 bucket name that doesn't exist, and the API checks the bucket before enabling exports. Could we use a real bucket configured for Buildkite exports, or leave this mutation covered by the mock API test?

buildkite/resource_organization_pipeline_settings_fake_test.go
```diff
@@ -0,0 +1,297 @@
...
+
+		resource "buildkite_organization_pipeline_settings" "settings" {
+			%s
+		}
+		`, server.URL, settings)
+	}
+
+	const name = "buildkite_organization_pipeline_settings.settings"
+
+	resource.Test(t, resource.TestCase{
```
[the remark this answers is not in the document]
omehegan: Fixed in 3dba75d. The fake lifecycle and its validation helper now use `resource.UnitTest`, and the test was renamed from `TestAcc…` to `TestUnit…` so it runs in the normal unit suite.

buildkite/resource_organization_pipeline_settings_test.go
omehegan: Fixed in 3dba75d. I removed the live build-export mutation that referenced an unprovisioned bucket. The enable/disable lifecycle remains covered by the fake API unit test, including DELETE-based clearing and the plan-gated 403 diagnostic.

scadu (approved):

PR #1226, opened 2026-08-23
fix(registry): keep team_ids as configured on read

dattallant:
`team_ids` on `buildkite_registry` is create-only, but Read replaces it with the API's `team_ids`, which lists every team that currently has access. Once another team is granted access in the UI, state holds two teams, the config holds one, and every following plan fails with "Team IDs change detected" from `ModifyPlan`.

Read now keeps the configured `team_ids` and only takes the API's list when state has none (right after an import), so out-of-band grants no longer break the resource while an edit of `team_ids` is still rejected. Added a table test for `handleTeamIDs` and an acceptance subtest that grants a second team access through the API and expects an empty plan (it fails with the old behaviour); ran it locally against a test organization.

## PR checklist:
- [x] tests added
- [ ] an example added to `examples/` <-- no new field


scadu:
Thanks for the contribution @dattallant! Looking good.

scadu (approved):

PR #1227, opened 2026-08-23
Add buildkite_team_registry resource

dattallant:
Team access to a package registry can only be set at creation today (`team_ids` on `buildkite_registry`, which the API can't change), so later grants aren't visible to Terraform. This adds `buildkite_team_registry`, the registry counterpart of `buildkite_test_suite_team`, on the `teamRegistryCreate`/`Update`/`Delete` mutations, read through the shared node query:

```hcl
resource "buildkite_team_registry" "gems_everyone" {
  registry_id  = buildkite_registry.gems.id
  team_id      = buildkite_team.everyone.id
  access_level = "READ_ONLY" # READ_ONLY | READ_AND_WRITE | READ_WRITE_AND_ADMIN
}
```

`team_id`/`registry_id` replace on change since the API only updates the access level. `generated.go` is regenerated from the vendored schema (new fragment, three mutations, one node case); no `schema.graphql` change. Depends on #1226, which stops `buildkite_registry` from reading granted teams back into `team_ids`; without it a registry managed next to this resource fails to plan after the grant.

Added acceptance tests (create, update the level, import, recreate when removed) and ran them locally against a test organization.

## PR checklist:
- [x] `docs/` updated
- [x] tests added


scadu (approved):

scadu:
Done some end to end testing with this PR combined with the #1226 – works as expected. Thanks for the contribution @dattallant!

PR #1231, opened 2026-08-24
docs: fix import examples for team and pipeline_team, add registry im…

dattallant:
The sample IDs in the `buildkite_team` and `buildkite_pipeline_team` import examples decode to a `Pipeline---…` and a `Team---…` ID respectively, so I swapped them for IDs of the right type, and `buildkite_registry` supports import by slug but had no import example, so I added one. 

Also one sentence on `buildkite_pipeline` noting that optional attributes left out of the configuration after an import (e.g. `color`, `emoji`) are cleared on the next apply. Docs regenerated with `make docs`.
  
## PR checklist:
- [x] `docs/` updated
- [ ] tests added <-- docs only


buildkite/resource_pipeline.go
```diff
@@ -632,6 +632,8 @@ func (*pipelineResource) Schema(ctx context.Context, req resource.SchemaRequest,
 			More information on pipelines can be found in the [documentation](https://buildkite.com/docs/pipelines).
 
 			-> **Note:** When creating a new pipeline, the Buildkite API requires at least one team to be associated with it. You must use the 'default_team_id' attribute to specify this initial team. The 'buildkite_pipeline_team' resource can then be used to manage team access for existing pipelines.
+
+			-> **Note:** After importing a pipeline, optional attributes that are set on the pipeline but left out of the configuration (for example 'color' or 'emoji') are cleared on the next apply.
```
scadu: Minor: This says optional attributes are cleared generally, but several are deliberately preserved when omitted, including `github_webhooks_enabled` and the optional/computed attributes. 

Could we narrow this to the affected fields, for example:
> After importing a pipeline, 'color' and 'emoji' values left out of the configuration are cleared on the next apply

The generated `docs/resources/pipeline.md` should then be refreshed with `make docs`.

dattallant: narrowed it to color/emoji and regenerated the docs

scadu: Thanks @dattallant!

scadu (approved):

PR #1235, opened 2026-08-25
Import pipelines, teams, schedules, pipeline teams etc by slug

### LINKED ISSUE #1230: Import pipeline schedules, pipeline teams and cluster queues by slug/key
`buildkite_pipeline_schedule`, `buildkite_pipeline_team` and `buildkite_cluster_queue` can only be imported by GraphQL ID today (buildkite_cluster_queue as <id>,<cluster-uuid>), and unlike pipelines or teams there is no data source to look those IDs up, so adopting an organization with a few hundred pipelines means harvesting IDs with a script first. #143 originally asked to "consistently support both" forms.

I'd like to propose additional, human-readable import IDs next to the existing ones: 
- `buildkite_pipeline_schedule` <pipeline-slug>/<schedule-uuid>,
- `buildkite_pipeline_team` <pipeline-slug>/<team-slug>, 
- `buildkite_cluster_queue` <cluster-uuid>/<queue-key>, 
- and for consistency `buildkite_pipeline` <slug> and `buildkite_team` <slug>.

GraphQL IDs keep working and stay the documented primary form; the new forms are detected by shape, id in state stays the GraphQL ID, and anything that doesn't resolve is a hard error. The lookups mostly exist already (getPipelineScheduleBySlug is generated but unused since #339, GetTeamFromSlug, getClusterQueues), and a lean pipeline(slug){id} query keeps the complexity cost trivial.

Happy to send the PR. Would you take this, or would you rather see the lookup side extended instead (e.g. schedules and teams lists on data.buildkite_pipeline) so import blocks can reference them?

dattallant:
Closes #1230. Adds human-readable import IDs next to the GraphQL ones, so a large organization can be imported without collecting GraphQL IDs first:

- `buildkite_pipeline`: `<pipeline slug>`
- `buildkite_team`: `<team slug>`
- `buildkite_pipeline_schedule`: `<pipeline slug>/<schedule uuid>`
- `buildkite_pipeline_team`: `<pipeline slug>/<team slug>`
- `buildkite_cluster_queue`: `<cluster uuid>/<queue key>` (the `id,cluster_uuid` form is unchanged)

It is purely additive: GraphQL IDs keep working and stay the first form in the docs, the new forms are recognised by shape in each `ImportState` (one small helper shared by all five) and resolved to the GraphQL ID, which is what is stored in `id`, and anything that doesn't resolve is a hard error naming the input. Since slugs and keys can change, the examples note that the GraphQL ID is the stable identifier. Lookups reuse `getPipelineScheduleBySlug`, `GetTeamFromSlug` and `getClusterQueues` plus two lean queries (`pipeline(slug){id}` and the pipeline's teams with slugs); `generated.go` is regenerated from the vendored schema.

Added a table test for the detector and an `ImportStateVerify` step with the new form to each resource's acceptance test; ran the team and cluster queue ones locally against a test organization and checked all five forms by hand with `terraform import` (the pipeline-based subtests create cluster-less pipelines, which my organization rejects).

## PR checklist:
- [x] `docs/` updated
- [x] tests added


buildkite/util.go
```diff
@@ -198,3 +198,23 @@ func retryContextError(err error) *retry.RetryError {
 	}
 	return retry.NonRetryableError(err)
 }
+
+var (
+	importSlugRegex = regexp.MustCompile(`^[a-z0-9][a-z0-9-]*$`)
```
scadu: Could we parse these forms per resource rather than apply one slug regex to every component? Buildkite accepts uppercase pipeline slugs ([published grammar](https://buildkite.com/docs/apis/rest-api/pipelines#deriving-a-pipeline-slug-from-the-pipelines-name)) and queue keys with underscores; the [REST and GraphQL examples](https://buildkite.com/docs/agent/queues/managing#create-a-self-hosted-queue) both use `linux_small_amd`.

With the current detector, `My-Pipeline` and `<cluster-uuid>/linux_small_amd` aren't recognized. The queue form then reaches the [existing comma parser](https://github.com/buildkite/terraform-provider-buildkite/blob/5e7df56f42378473005a4ccc7ae00f50f8082219/buildkite/resource_cluster_queue.go#L466-L474) and fails. The simplest queue parser might only validate the cluster UUID and require a non-empty key, then let the lookup decide whether that key exists. Could we add regression tests for both examples as well? The [current table test](https://github.com/buildkite/terraform-provider-buildkite/blob/5e7df56f42378473005a4ccc7ae00f50f8082219/buildkite/util_test.go#L227-L259) explicitly treats `My-Pipeline` as passthrough.

buildkite/resource_pipeline_team.go
```diff
@@ -193,9 +193,38 @@ func (tp *pipelineTeamResource) Read(ctx context.Context, req resource.ReadReque
...
+		if err != nil {
+			resp.Diagnostics.AddError("Unable to import pipeline team", fmt.Sprintf("Could not find team %q on pipeline %q: %s", parts[1], parts[0], err.Error()))
+			return
+		}
+		req.ID = id
+	}
 	resource.ImportStatePassthroughID(ctx, path.Root("id"), req, resp)
 }
 
+// findPipelineTeamID pages through the pipeline's teams for the one with the given slug
```
scadu: Could we use the connection's [`search` argument](https://github.com/buildkite/docs/blob/c30803ba6fd4303c66d44fe29c1d4d3d01410530/data/graphql/schema.graphql#L8376-L8409) rather than fetching every assignment? The existing [`GetTeamFromSlug` query](https://github.com/buildkite/terraform-provider-buildkite/blob/5e7df56f42378473005a4ccc7ae00f50f8082219/buildkite/graphql/team.graphql#L1-L21) already returns the team name. We could resolve that first, pass the name to:

```
teams(
  first: 50
  after: $cursor
  search: $teamName
  order: RELEVANCE
)
```

We should still keep the exact slug comparison, since the search may return partial matches. This avoids restarting the [full assignment scan](https://github.com/buildkite/terraform-provider-buildkite/blob/5e7df56f42378473005a4ccc7ae00f50f8082219/buildkite/graphql/pipeline.graphql#L475-L496) for every import. With 50 assignments, those scans alone cost roughly 5,200 complexity points before the resource reads.

buildkite/resource_cluster_queue.go
```diff
@@ -469,6 +479,26 @@ func (cq *clusterQueueResource) ImportState(ctx context.Context, req resource.Im
 	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("cluster_uuid"), importComponents[1])...)
 }
 
+// findClusterQueueID pages through the cluster's queues for the one with the given key
```
scadu, line 482: We could trim the existing `getClusterQueues` query to the fields its only caller needs. `findClusterQueueID` only reads `id` and `key`, and I couldn’t find any other consumer of `getClusterQueues`. 
We can leave the shared `ClusterQueueValues` fragment unchanged for the create, update, and node-read operations, then select the two fields inline here:

```grapqhl
query getClusterQueues(
  $orgSlug: ID!
  $id: ID!
  $cursor: String
) {
  organization(slug: $orgSlug) {
    cluster(id: $id) {
      queues(order: KEY, first: 100, after: $cursor) {
        pageInfo {
          endCursor
          hasNextPage
        }
        edges {
          node {
            id
            key
          }
        }
      }
    }
  }
}
```

This should reduce query complexity points by more than 4x.

scadu (changes requested):

buildkite/util.go
dattallant: Done in `f7a7586`: each resource parses its own form now (pipeline slugs use the published `[a-zA-Z0-9]+[a-zA-Z0-9-]*` grammar, the queue form only checks the cluster UUID plus a non-empty key and lets the lookup decide), with My-Pipeline and `<uuid>/linux_small_amd` in the table test. 

One wrinkle that came with allowing uppercase: an unpadded GraphQL ID has the same shape as a slug, so GraphQL IDs are now recognised by decoding them (<Type>---<uuid>) before the slug grammar is tried, which keeps them the default.

buildkite/resource_pipeline_team.go
dattallant: yeah good call, f7a7586 resolves the team with GetTeamFromSlug first and then queries teams(search: <name> order: RELEVANCE) on the pipeline, still comparing the slug exactly since the search can partial-match.

buildkite/resource_cluster_queue.go
dattallant: Trimmed in `f7a7586` - getClusterQueues now selects just id and key inline and the ClusterQueueValues fragment is left as it was for create/update/read.

scadu (approved):
Looking good. Thanks again for your contribution @dattallant!

PR #1239, opened 2026-08-27
refactor(organization)!: manage allowed_api_ip_addresses via the REST API

petetomasik:
## Summary

`allowed_api_ip_addresses` on `buildkite_organization` was written through the GraphQL
`organizationApiIpAllowlistUpdate` mutation, while `restrict_user_api_token_creation` and
`revoke_inactive_tokens_after` already went through `GET`/`PATCH /v2/organizations/{org}/api-settings`.
One setting was being managed through two APIs.

This routes the allowlist through `api-settings` for both the resource and the data source, then
removes the mutation and the `allowedApiIpAddresses` field from the GraphQL operations, since nothing
reads them any more. `enforce_2fa` stays on GraphQL because the REST endpoint has no equivalent.

## Behaviour changes

The allowlist and the token-creation restriction now travel in the same `PATCH`, which is how the API
applies the two atomically.

Plan-gated refusals are reported usefully. The API answers 403 with its own explanation for a setting
the organization's plan does not include, and the provider surfaces that message alongside the name of
the missing feature. The previous code appended its plan hint on any failed write, including a 500; it
is now gated on an actual 403.

An unchanged allowlist still stays out of the request. An organization without the feature is refused
even a no-op write, so the resource only sends the value when it differs from what the API reports.

Managing `buildkite_organization` now needs an API token carrying the `read_organization_settings`
scope whatever the configuration sets, plus `write_organization_settings` to change any attribute other
than `enforce_2fa`, and the token's user must be an organization administrator. The resource reads
api-settings on every apply so it can record the values it adopts rather than invent them, which is why
an `enforce_2fa`-only configuration needs the read scope too. `README.md` and the provider's `api_token`
description recommended a token carrying neither scope; both now list them.

The data source degrades rather than failing. A 403 leaves `allowed_api_ip_addresses` unset with a
warning, so `id` and `uuid` still resolve for a caller who cannot see the settings. It also reports an
organization with no allowlist as no addresses: splitting the empty string the API answers with used to
report a single blank one, so anything indexing `allowed_api_ip_addresses[0]` on such an organization
will now find nothing there.

Failure paths keep the organization and state in step. Destroy asks the organization for the current
allowlist before deciding whether to clear it, since state alone cannot say whether there is one to
clear, and a read it cannot make stops the destroy rather than guessing, naming the ways forward. A
create that writes the settings and then fails on 2FA records no state, so the replacement apply that
would otherwise destroy first cannot clear an allowlist that did land; it warns about what was left
behind only when something actually was.

## Docs

The scopes are now stated once in the `buildkite_organization` resource description instead of being
repeated on three attributes, and each attribute keeps only what is specific to it. `README.md` and the
provider's `api_token` description carry both organization settings scopes, noting they are only needed
for `buildkite_organization`.

## Testing

New `buildkite/resource_organization_fake_test.go` stands up an httptest fake serving `api-settings`
plus the two GraphQL queries the resource and data source need for their identifiers, so the lifecycle
runs without an organization to change. Its GraphQL side no longer offers the allowlist at all, so
there is only one source for it. It covers:

- the full allowlist lifecycle including import, clearing with `[]`, and clearing by removing the attribute
- that a configuration saying nothing about the allowlist writes nothing, verified by making the fake
  refuse every write
- that a plan-gated 403 reports both the API's message and the missing feature
- that an `enforce_2fa`-only configuration still needs the settings read, and gives up before touching 2FA
- that the settings write reports whether it wrote, so a failed create only warns about what landed
- that a destroy which cannot read the settings names the way out, on a status the scope hint does not fit
- the data source reading the allowlist over REST, its separator handling, and leaving it unset on a 403
  while still returning `id` and `uuid`

`TestApiSettingsPatch` gained six allowlist cases, and the acceptance helpers preset and verify the
allowlist over REST. The data source acceptance test asserted the blank address the old split produced,
and now checks the identifiers a default lookup is for.

## PR checklist:
- [x] `docs/` updated
- [x] tests added
- [ ] an example added to `examples/` (useful to demo new field/resource) <-- No new field; the existing example already covers `allowed_api_ip_addresses`

@buildsworth-bk approve up to L2


buildkite/resource_organization.go
```diff
@@ -383,6 +350,12 @@ func apiSettingsFromModel(model *organizationResourceModel) *organizationAPISett
 // apiSettingsPatch returns the configured api-settings that differ from current, or all of them when current isn't known
 func apiSettingsPatch(config, plan *organizationResourceModel, current *organizationAPISettings, currentKnown bool) map[string]any {
 	payload := map[string]any{}
+	// the allowlist is owned rather than adopted: dropping it from the configuration clears it, so
+	// the plan alone says what it should be. Unchanged values stay out of the request because
+	// organizations without the allowlist feature are refused even those.
+	if allowed := allowedApiIpAddressesValue(plan.AllowedApiIpAddresses); allowed != current.AllowedIpAddresses {
+		payload["allowed_ip_addresses"] = allowed
+	}
```
scadu, line 372: We could handle permission issues better here (although I'd say this one would be on an edge case side).
When the token lacks `read_organization_settings`, the settings GET returns 403 and falls back to prior state. That fallback predates this PR, but moving the allowlist into this path creates a new failure mode: on create there is no prior state, so both the fallback and an explicit `allowed_api_ip_addresses = []` serialize to `""`. The `PATCH` is skipped and Terraform records `[]`, even if the live allowlist is still set.

I reproduced this against my test org. Since the resource now requires both REST scopes, could we fail on a forbidden GET rather than use the fallback for this field? 🤔

petetomasik: Fixed in e5ec2b9. `updateAPISettings` now errors on a forbidden GET instead of falling
back to prior state, naming the missing scope, with a regression test for the create case you
reproduced.

Refresh still tolerates a 403 and keeps its last known values, so losing the scope doesn't fail
every plan, but drift stays invisible until the next apply.

petetomasik:
@buildsworth-bk perform new review

buildkite/resource_organization.go
```diff
@@ -138,36 +141,27 @@ func (o *organizationResource) Create(ctx context.Context, req resource.CreateRe
...
 
+	state.Enforce2FA = types.BoolValue(organization.Organization.MembersRequireTwoFactorAuthentication)
 	if !plan.Enforce2FA.IsNull() && !plan.Enforce2FA.IsUnknown() && plan.Enforce2FA.ValueBool() != organization.Organization.MembersRequireTwoFactorAuthentication {
-		_, err = setOrganization2FA(ctx, o.client.genqlient, *org, plan.Enforce2FA.ValueBool())
-		if err != nil {
+		if _, err := setOrganization2FA(ctx, o.client.genqlient, *org, plan.Enforce2FA.ValueBool()); err != nil {
 			resp.Diagnostics.AddError("Unable to set 2FA", err.Error())
+			// the settings above are the organization's now, so they are recorded even though the
+			// apply failed. 2FA is left to show as drift rather than the resource going untracked.
+			resp.Diagnostics.Append(resp.State.Set(ctx, state)...)
```
scadu: When the API settings write succeeds but this 2FA mutation fails, could we avoid setting state here? Terraform marks the resource as tainted, so the next apply replaces it and `Delete` clears the live allowlist before retrying. I reproduced this: the retry cleared the allowlist between the initial and replacement writes. Since this resource manages an existing organization, I think it'd safer to return the error without state and warn that the REST settings were already applied.

scadu (changes requested):

buildkite/resource_organization.go
petetomasik: Yep, agreed. Addressed in a86efdd

PR #1250, opened 2026-09-02
fix: persist what a failed organization apply changed

petetomasik:
A failure part way through `Update` left Buildkite holding settings that state did not record. State was written once, at the end, so any error path after a mutation had already applied returned without reaching `resp.State.Set`. Under `-refresh=false` the next plan then saw no diff and the change was never reconciled.

`Update` now persists in a defer registered at the point the first mutation applies, so a new early return cannot forget it. The attributes are seeded from the prior state before that defer is registered and each is overwritten by the step that owns it, so a step that is never reached cannot write a null over a setting the organization still has. That matters for the api-settings pair in particular, since a non-403 failure on the GET returns before `updateAPISettings` assigns either attribute.

`Create` cannot do the same. Terraform taints an instance whose `Create` returns both state and an error, and a tainted prior becomes a replace on the next plan. This resource configures an organization that already exists rather than creating one, so being replaced runs `Delete`, which clears the API IP allowlist. Every step compares before it mutates, so recording nothing and letting the next apply re-run it converges on its own. What is lost is any record that the earlier steps took effect, so `Create` reports them in a warning naming each one, also from a defer, since that warning is the only place those changes are written down.

One correction fell out of this. `updateAPISettings` copied the planned `revoke_inactive_tokens_after` and `restrict_user_api_token_creation` into state ahead of the PATCH that applies them, so now that state is always persisted, a failed PATCH would leave state claiming a setting that never applied. It records the organization's values on that path instead. That case is worst when the GET was also refused, because `readAPISettings` falls back to state on a 403, so a wrong value put there once is re-adopted on every refresh and never shows in a plan.

Tests drive `Create` and `Update` directly against a stubbed API rather than through the acceptance harness: that `Create`'s warning names the allowlist and 2FA changes, and only the ones this apply made; that `Create` records no state at all; that an enforced 2FA survives a failing api-settings PATCH in `Update`; that a failed api-settings read keeps the prior values rather than nulling them; and that a failed PATCH leaves both attributes describing the organization. The shared scaffolding for these lives in `buildkite/helpers_test.go`.

## PR checklist:
- [ ] `docs/` updated <-- no schema or attribute changes, so `make docs` is a no-op; the general partial-failure guidance from #1249 is not part of this PR
- [x] tests added <-- unit tests over Create's warning, Create recording nothing, and the three Update failure paths
- [ ] an example added to `examples/` (useful to demo new field/resource) <-- no new field or resource
- [ ] `CHANGELOG.md` updated with pending release information <-- entries are compiled in the release PR, matching current practice on this repo


petetomasik:
@buildsworth-bk review and approve up to L2

PR #1252, opened 2026-09-02
fix: persist what a failed cluster queue apply applied

petetomasik:
Pausing or resuming dispatch is a mutation of its own, applied before the queue mutation that follows it. When that later call failed, Update returned without recording state, so Terraform went on believing dispatch was running on a queue that was really paused, and nothing in the next plan said otherwise until state was refreshed.

Both Create and Update now persist from the point the queue exists, registered in a defer so a new early return cannot forget it, and record the pause as soon as it applies rather than waiting for a later response to confirm it. This replaces two error paths in Create that persisted with a bare resp.State.Set whose diagnostics went nowhere.


petetomasik:
@buildsworth-bk perform a review and approve up to L2

PR #1254, opened 2026-09-02
fix: persist what a failed test suite apply applied

petetomasik:
Update PATCHes the suite and then swaps its owner team over two GraphQL mutations, and returning without recording state dropped everything the PATCH had changed. The slug is the damaging part: state kept naming the slug the suite answered to before the rename, and that is the slug the next update PATCHes, so under -refresh=false the request went to a URL the suite no longer had.

State is now persisted in a defer registered where the PATCH applies, and the new owner is recorded only once the previous one is detached, since both own the suite until then.

Attaching an owner that already owns the suite is an error, which made the detach unreachable on a re-run and left both teams owning the suite with no way forward. That case is tolerated so the detach is retried, and because the attach response carries no teams on that path, the suite's owners are read back through getTestSuite instead.


petetomasik:
@buildsworth-bk review and approve up to L2

tomowatt (approved):

PR #1255, opened 2026-09-02
fix: stop recording read values the API never returned

petetomasik:
Three read paths answered a failed API call by warning and then writing a value of their own into state. The warning scrolls past; the value persists and reads back as though Buildkite had returned it.

`buildkite_cluster_queue` is the damaging one. A failed `retry_agent_affinity` read recorded the schema default, so a queue configured `prefer-warmest` whose real setting is `prefer-different` agreed with its own configuration, no plan ever showed the drift, and the queue stayed wrong. It now keeps the last known value. The `buildkite_cluster` and `buildkite_clusters` data sources did the same thing with an empty maintainers list.

The line is between a refusal and a failure. A 403 is an answer: the caller may not read that attribute, and reporting no maintainers is a fair reading of it, so those keep their warning. Every other failure means the API never answered, which is an error rather than a value. `resource_organization.go` already draws this line for its api-settings.

Keeping the last known value assumes there is one. An import records only the id and the cluster uuid, and the schema default applies whenever the config value is null regardless of the prior, so a refused affinity read against imported state planned `prefer-warmest` against a null prior and then failed the apply on the same refusal, every time. That case errors and names the `read_clusters` scope the token is missing.

`buildkite_clusters` now stops at the first cluster whose maintainers read fails, rather than sweeping the remaining pages to fail the same way.

Behavior change worth a reviewer's attention: a token that can read a queue over GraphQL, but not over REST, used to get a warning and a fabricated affinity on every read. It still gets a warning where state has a value to keep, but against imported state it now gets an error. That is the intent of the change, not a side effect, but it is visible to restricted tokens.

Tests are table-driven unit tests over each branch of the three read paths (readable, refused, gone, and any other failure), plus one asserting the clusters data source stops at the first failure. The shared framework scaffolding for those tests lives in `buildkite/helpers_test.go`.

## PR checklist:
- [ ] `docs/` updated <-- no schema or attribute changes, only diagnostics behavior
- [x] tests added <-- unit tests covering every branch of the three read paths
- [ ] an example added to `examples/` (useful to demo new field/resource) <-- no new field or resource
- [ ] `CHANGELOG.md` updated with pending release information <-- entries are compiled in the release PR, matching current practice on this repo


petetomasik:
@buildsworth-bk provide a new review, with risk approval up to L2.

PR #1256, opened 2026-09-03
Add buildkite_pipelines data source

dattallant:
There's a data source for one pipeline by slug but no way to list them, so "which pipelines exist in this cluster / for this repository / are archived" still needs a script. This adds `data "buildkite_pipelines"` in the shape of `buildkite_teams`, passing the GraphQL `organization.pipelines` filters straight through and exposing its `count` as `total`:

```hcl
data "buildkite_pipelines" "deploys" {
cluster_id = buildkite_cluster.primary.id
archived   = false
search     = "deploy"
}
```

Filters are `search`, `repository` (URL), `cluster_id`, `archived` and `tags`, all optional and omitted from the query when unset; each pipeline has `id`, `uuid`, `slug`, `name`, `description`, `repository`, `default_branch`, `cluster_id` and `archived`. It is one paginated GraphQL query at 500 per page (about 3 complexity points per pipeline, so roughly 1.5k requested per full page), which is why the docs point large organizations at the filters; the `team` filter is left out for now because `TeamSelector` isn't bound in `genqlient.yaml`. `generated.go` is regenerated from the vendored schema.

Added acceptance tests (unfiltered, and filtered by search/cluster/tags) and ran them locally against a test organization.

## PR checklist:
- [x] `docs/` updated
- [x] tests added


buildkite/data_source_pipelines.go
```diff
@@ -0,0 +1,189 @@
...
+
+	var repository *PipelineRepositoryInput
+	if !state.Repository.IsNull() {
+		repository = &PipelineRepositoryInput{Url: state.Repository.ValueString()}
+	}
+
+	state.Pipelines = []pipelinesModel{}
+	var cursor *string
+	for {
+		res, err := GetOrganizationPipelines(ctx, p.client.genqlient, p.client.organization, cursor, state.Search.ValueStringPointer(), repository, state.ClusterId.ValueStringPointer(), state.Archived.ValueBoolPointer(), state.Tags)
```
petetomasik, line 152: A null `organization` reaches this call as a success, so the read reports zero pipelines instead of failing.

`Query.organization(slug:)` is nullable, and Buildkite doesn't return a GraphQL error when the slug is unknown or the token can't see it. Both cases come back as HTTP 200 with no `errors` key:

```
POST graphql.buildkite.com/v1 -> 200
{"data":{"organization":null}}
```

`GetOrganizationPipelinesResponse.Organization` is generated as a value struct, and unmarshalling `null` into a struct leaves it zero, so genqlient returns `err == nil` and the `if err != nil` below never fires. I confirmed the decode directly against the generated type: `err = nil, Count = 0, len(Edges) = 0, HasNextPage = false`. The loop then runs once, sets `total = 0`, appends nothing, breaks on `!HasNextPage`, and calls `resp.State.Set` with no diagnostic.

The trigger is mundane: a typo in `BUILDKITE_ORGANIZATION_SLUG`, or a token that lost org access or lacks GraphQL scope. Because it surfaces as a successful empty inventory rather than an error, a consumer doing

```hcl
for_each = { for p in data.buildkite_pipelines.all.pipelines : p.slug => p }
```

gets an empty map and Terraform plans to destroy everything keyed off it.

Worth flagging that the three sibling list data sources all fail loudly here (`data_source_teams.go:142`, `data_source_clusters.go:170`, `data_source_organization_members.go:97`), so this is a behaviour change relative to `buildkite_teams`, which this file otherwise follows closely. Their `len(edges) == 0` guard won't transplant, though, since an empty result is a legitimate answer for a filtered query and conflating the two would be its own bug.

The check needs to be on the organization rather than the edges, which is the pattern already at `data_source_organization.go:56`. Add `id` to the selection in `graphql/organization_pipelines.graphql`:

```graphql
organization(slug: $slug) {
    id
    pipelines(first: 500, after: $cursor, ...) {
```

then guard before accumulating:

```go
if res.Organization.Id == "" {
    resp.Diagnostics.AddError(
        "Unable to find organization",
        fmt.Sprintf("Could not find organization with slug \"%s\"", p.client.organization),
    )
    return
}
```

For context on why this is worth blocking rather than filing: the same failure mode was fixed in `buildkite_organization` in #1175 ("SUP-7447: Fix buildkite_organization data source silently returning empty on unresolved org slug"), so it's a known class here.

petetomasik (changes requested):
Thanks for the contribution @dattallant ! There's one additional guard we might want to add to this new data source.

buildkite/data_source_pipelines.go
dattallant: Done in `d123553`: the query now selects the organization id and the read errors with "Could not find organization" when it comes back empty, the same way buildkite_organization has since #1175; a filtered query that matches nothing still returns total = 0 and an empty list.

petetomasik (approved):

PR #1258, opened 2026-09-04
feat: add cluster cache registry resource

blaknite:
## PR checklist:
- [x] `docs/` updated
- [x] tests added

## Description

Adds the `buildkite_cluster_cache_registry` Terraform resource with GraphQL create, read, update, and delete support, import by GraphQL CacheRegistry ID, Relay node refresh, and structured JSON policy handling.

The policy attribute validates JSON objects and uses semantic normalization so API formatting and key ordering do not produce perpetual diffs. Omitted policy receives and retains the API default. Terraform Plugin Framework cannot distinguish removing an optional+computed policy from retaining its state, so clearing all rules requires explicitly setting empty `save.scopes`, `restore.scopes`, and `rules` values as documented by the resource.

## Context

The provider API schema was generated from the local `buildkite/buildkite` Cache Registry GraphQL branch because this API is not yet deployed publicly. The committed schema was then minimized against provider `main` to contain only the Cache Registry types, inputs, payloads, cluster field, mutations, Relay node, and audit union additions required by these operations.

Release and deployment depend on both upstream API changes landing and the GraphQL API becoming publicly available:

- REST API: https://github.com/buildkite/buildkite/pull/33745
- GraphQL API: https://github.com/buildkite/buildkite/pull/33753

## Changes

- Add `buildkite_cluster_cache_registry` with full lifecycle and GraphQL ID import.
- Add generated genqlient Cache Registry operations and Relay node lookup support.
- Add schema/model, policy semantics, lifecycle, rename, import, and out-of-band deletion coverage.
- Add resource examples, generated documentation, provider registration, and changelog entry.
- Add HashiCorp's `terraform-plugin-framework-jsontypes` for normalized JSON semantics.

## Verification

- `go tool genqlient` passed.
- `TF_ACC=1 BUILDKITE_GRAPHQL_URL=http://graphql.buildkite.localhost:3100/v1 BUILDKITE_API_TOKEN=bkua_development_api_token BUILDKITE_ORGANIZATION_SLUG=buildkite mise exec terraform@1.14.3 -- go test ./buildkite -run '^TestAccBuildkiteClusterCacheRegistryResource$' -v -count=1` passed (`ok github.com/buildkite/terraform-provider-buildkite/buildkite 3.658s`).
- `mise exec terraform@1.14.3 -- make test` passed.
- `make vet` passed.
- `go vet ./...` passed.
- `make testfmt` passed.
- `make build` passed.
- `git diff --check` passed.
- Documentation generation passed with Terraform 1.14.3.

Targeted acceptance testing ran against the local Buildkite instance. Production acceptance tests did not run.

## Deployment

Risk L2. Do not release this provider resource until buildkite/buildkite REST PR #33745 and GraphQL PR #33753 have landed and the Cache Registry GraphQL API is publicly available. There are no migrations or provider-side feature flags.

@buildsworth-bk permission to approve if risk L2

## Rollback

Revert this change before release if the upstream API is delayed or its contract changes. After release, reverting removes the resource from the provider schema but does not delete existing Cache Registries from Buildkite; users would need to retain the prior provider version or remove affected resources from Terraform state before upgrading.


buildkite/resource_cluster_cache_registry.go
```diff
@@ -0,0 +1,370 @@
...
+			},
+			"emoji": resource_schema.StringAttribute{
+				Optional:            true,
+				MarkdownDescription: "An emoji for the cache registry using Buildkite emoji syntax.",
+			},
+			"color": resource_schema.StringAttribute{
+				Optional:            true,
+				MarkdownDescription: "A color for the cache registry as a hex code, for example `#BADA55`.",
+			},
+			"policy": resource_schema.StringAttribute{
```
petetomasik, line 112: `policy` is Optional+Computed but has no plan modifier, unlike every other computed attribute in this schema (`id`, `uuid`, `cluster_uuid`, `slug`, `created_at` all carry `UseStateForUnknown()`). That means it goes unknown on every update and we send an explicit `policy: null` to the API.

The framework marks a computed attribute unknown based purely on the *config* value being null, with no check on the planned value. See `MarkComputedNilsAsUnknown` in terraform-plugin-framework@v1.19.0, `internal/fwserver/server_planresourcechange.go:465-537` (it logs "marking computed attribute that is null in the config as unknown"), gated at `:200` on the plan differing from prior state.

So the chain is:

1. No `policy` in config, so the config value is null.
2. User changes only `description`. Plan differs from prior state, so the transform runs.
3. `policy` is computed with a null config and no `Default`, so it is marked unknown.
4. Nothing restores it, since there is no plan modifier here.
5. `cacheRegistryPolicyPayload` returns nil for unknown (line 302).
6. `Policy *string` has no `omitempty` (generated.go:5434), so genqlient marshals an explicit null.

I confirmed step 5 and 6 locally:

```
cacheRegistryPolicyPayload(unknown) = <nil>
marshalled update input: {...,"color":null,"policy":null}
```

Then `updateClusterCacheRegistryState` takes the API value unconditionally, because `state.Policy.IsUnknown()` is true at line 295.

Impact depends on how the resolver treats an explicit null. If it means "clear", an unrelated description edit silently wipes the policy. If it is ignored, every update still plans `policy -> (known after apply)`. Either way it contradicts the attribute description on line 116, which says removing a configured policy retains its current value. That only holds when the plan is otherwise a complete no-op, in which case no update runs at all.

buildkite/resource_cluster_cache_registry.go
```diff
@@ -0,0 +1,370 @@
...
+	err := retry.RetryContext(ctx, timeout, func() *retry.RetryError {
+		organizationID, err := r.client.GetOrganizationID()
+		if err == nil {
+			log.Printf("Deleting cache registry with ID %s ...", state.ID.ValueString())
+			_, err = deleteCacheRegistry(ctx, r.client.genqlient, *organizationID, state.ID.ValueString())
+		}
+		if isResourceNotFoundError(err) {
+			return nil
+		}
+		return retryContextError(err)
```
petetomasik, line 257: This conflates a failed org lookup with "the registry is already gone", and the result is that we drop a live resource from state.

`GetOrganizationID()` and the delete mutation share one closure, and `isResourceNotFoundError` is applied to the combined `err`. When the org cannot be resolved, `GetOrganizationID` returns `fmt.Errorf("organization %s not found", slug)` (util.go:138). That is a plain error, not a `gqlerror.List`, so `isResourceNotFoundError` falls through to `resourceNotFoundRegex.MatchString(err.Error())` (util.go:40), and the regex matches on `not\s+found` (util.go:20). The closure returns nil, `Delete` reports success, and Terraform removes the resource from state while the cache registry still exists in Buildkite.

Trigger: `terraform destroy` with a token whose org access has been revoked, or after the org slug changes.

I grepped every Delete in the package, and this is the only one that calls `GetOrganizationID` and `isResourceNotFoundError` in the same closure, so it is not an inherited pattern. `resource_team_registry.go`, `resource_pipeline_schedule.go` and `resource_pipeline.go` all scope the check to the mutation error.

```suggestion
		organizationID, err := r.client.GetOrganizationID()
		if err != nil {
			return retryContextError(err)
		}

		log.Printf("Deleting cache registry with ID %s ...", state.ID.ValueString())
		if _, err := deleteCacheRegistry(ctx, r.client.genqlient, *organizationID, state.ID.ValueString()); err != nil {
			if isResourceNotFoundError(err) {
				return nil
			}
			return retryContextError(err)
		}
		return nil
```

Worth adding a `CheckDestroy` to the acceptance test as well, since that is the check that would have caught this.

petetomasik (changes requested):
Couple of issues that might need some adjustments.

buildkite/resource_cluster_cache_registry.go
blaknite: Thanks for tracing this through the framework planning behavior. I added `UseStateForUnknown()` and acceptance coverage proving an omitted API-default policy survives a description-only update in c5fefaf.

buildkite/resource_cluster_cache_registry.go
```diff
@@ -0,0 +1,377 @@
...
+
+	err := retry.RetryContext(ctx, timeout, func() *retry.RetryError {
+		organizationID, err := r.client.GetOrganizationID()
+		if err != nil {
+			return retryContextError(err)
+		}
+
+		log.Printf("Deleting cache registry with ID %s ...", state.ID.ValueString())
+		if _, err := deleteCacheRegistry(ctx, r.client.genqlient, *organizationID, state.ID.ValueString()); err != nil {
+			if isResourceNotFoundError(err) {
```
petetomasik: Thanks for scoping this to the mutation error in 6a7402b. That was the right move, but it narrows the hole rather than closing it, because the delete mutation itself raises an org-not-found error. So the chain is unchanged from the original report, just relocated:

1. `deleteCacheRegistry` returns a `gqlerror.List` whose message is "No organization found".
2. `isResourceNotFoundError` iterates the list and tests `resourceNotFoundRegex` (util.go:20), which matches on `No\s+\w+(\s+\w+)*\s+found`.
3. "No organization found" matches, so this returns nil.
4. `Delete` reports success and Terraform drops the resource while the registry still exists in Buildkite.

Trigger: `terraform destroy` after the token's access to the org is revoked, or with a stale cached org ID (`GetOrganizationID` memoises in client.go, so a long-lived provider instance can hold one across a permission change).

The message we actually want to accept is the registry's own, `CACHE_REGISTRY_NOT_FOUND_ERROR = RecordNotFound.new("No cache registry found")`. Narrowing to that also drops the raw-string fallback in `isResourceNotFoundError`, which closes the `*HTTPError` body-regex hazard called out in the comment on `restErrorHandler` in client.go:

```go
		log.Printf("Deleting cache registry with ID %s ...", state.ID.ValueString())
		if _, err := deleteCacheRegistry(ctx, r.client.genqlient, *organizationID, state.ID.ValueString()); err != nil {
			if isCacheRegistryNotFoundError(err) {
				return nil
			}
			return retryContextError(err)
		}
		return nil
```

with:

```go
// The shared helper also matches "No organization found", which the delete
// mutation raises when the organization cannot be resolved, so scope this to
// the registry's own message and to structured GraphQL errors only.
var cacheRegistryNotFoundRegex = regexp.MustCompile(`(?i)no\s+cache\s+registry\s+found`)

func isCacheRegistryNotFoundError(err error) bool {
	var errList gqlerror.List
	if !errors.As(err, &errList) {
		return false
	}
	for _, e := range errList {
		if cacheRegistryNotFoundRegex.MatchString(e.Message) {
			return true
		}
	}
	return false
}
```

The `CheckDestroy` added in this revision would not catch this, since it only runs after a successful destroy. A test that stubs the mutation returning "No organization found" and asserts `Delete` errors would.

buildkite/resource_cluster_cache_registry.go
```diff
@@ -0,0 +1,377 @@
...
+	state.ID = types.StringValue(result.Id)
+	state.UUID = types.StringValue(result.Uuid)
+	state.ClusterID = types.StringValue(result.CacheRegistryCluster.Id)
+	state.ClusterUUID = types.StringValue(result.CacheRegistryCluster.Uuid)
+	state.Name = types.StringValue(result.Name)
+	state.Slug = types.StringValue(result.Slug)
+	state.Description = types.StringPointerValue(result.Description)
+	state.Emoji = types.StringPointerValue(result.Emoji)
+	state.Color = types.StringPointerValue(result.Color)
+	if state.Policy.IsNull() || state.Policy.IsUnknown() || result.Policy == nil || !cacheRegistryPoliciesEquivalent(state.Policy.ValueString(), *result.Policy) {
```
petetomasik, line 348: The `else` path here, keeping the configured text when the API's expanded form is semantically equivalent, is the entire reason this is not a plain assignment. Nothing tests it.

I confirmed by mutation: replacing the whole condition with `if true` leaves every test green.

```
$ perl -0pi -e 's/if state\.Policy\.IsNull\(\).*?\{/if true {/s' buildkite/resource_cluster_cache_registry.go
$ go test ./buildkite -run 'TestClusterCacheRegistry|TestUpdateClusterCacheRegistryState|TestCacheRegistryPolicy' -count=1
ok  	github.com/buildkite/terraform-provider-buildkite/buildkite	0.657s
```

That mutation reinstates the perpetual-diff bug this function exists to prevent and undoes the fix for the scalar-`action` blocker from the first review round, and the suite does not notice.

Why the existing tests miss it. `TestUpdateClusterCacheRegistryState` (line 61) passes a zero-value `state`, so `state.Policy.IsNull()` is true and it short-circuits on the first term of the condition every time. `TestCacheRegistryPolicy` (line 149) exercises `cacheRegistryPoliciesEquivalent` in isolation but never through the function that calls it, so the wiring between them is unverified.

The acceptance test does not compensate, because it does not run in CI. From this PR's own passing build, https://buildkite.com/buildkite/terraform-provider-buildkite-main/builds/4802, in the `acceptance tests` job:

```
SKIP buildkite.TestAccBuildkiteClusterCacheRegistryResource (0.00s)

=== SKIP: buildkite TestAccBuildkiteClusterCacheRegistryResource (0.00s)
    resource_cluster_cache_registry_test.go:170: cache registry acceptance tests require the local Buildkite GraphQL endpoint
```

`localPreCheck` (line 169) skips unless `BUILDKITE_GRAPHQL_URL` contains `graphql.buildkite.localhost`. That variable is set nowhere in `.buildkite/pipeline.yml` or `docker-compose.yml` and never appears in the job log; the only other reference to it in the repo is the `make schema` target. The job still exits 0, so the skip is silent and the check reports green.

What did run in that job and pass: `TestClusterCacheRegistrySchema`, `TestUpdateClusterCacheRegistryState`, and `TestCacheRegistryPolicy`. That is the entire CI safety net for this resource, and none of the three can fail if this branch breaks. Everything that would have caught it, the create step, the unrelated-update retention step, the `PlanOnly` reorder step, import, and the out-of-band delete, is behind the skip.

This needs no API access to close. The shapes below are what `CacheRegistry::Policy#to_h` actually returns: `save.scopes`, `restore.scopes` and `rules` are always present, and `Array.wrap` widens a scalar rule action.

```go
func TestUpdateClusterCacheRegistryStatePolicy(t *testing.T) {
	t.Parallel()

	tests := map[string]struct {
		configured, fromAPI, want string
	}{
		"expanded defaults keep the configured text": {
			configured: `{"save":{"scopes":{"branch":true}}}`,
			fromAPI:    `{"save":{"scopes":{"branch":true}},"restore":{"scopes":[]},"rules":[]}`,
			want:       `{"save":{"scopes":{"branch":true}}}`,
		},
		"widened rule actions keep the configured text": {
			configured: `{"rules":[{"effect":"allow","action":"save"}]}`,
			fromAPI:    `{"save":{"scopes":{}},"restore":{"scopes":[]},"rules":[{"effect":"allow","action":["save"]}]}`,
			want:       `{"rules":[{"effect":"allow","action":"save"}]}`,
		},
		"a genuinely different policy takes the API value": {
			configured: `{"rules":[]}`,
			fromAPI:    `{"save":{"scopes":{}},"restore":{"scopes":[]},"rules":[{"effect":"deny","action":["save"]}]}`,
			want:       `{"save":{"scopes":{}},"restore":{"scopes":[]},"rules":[{"effect":"deny","action":["save"]}]}`,
		},
	}

	for name, test := range tests {
		t.Run(name, func(t *testing.T) {
			t.Parallel()
			state := clusterCacheRegistryResourceModel{Policy: jsontypes.NewNormalizedValue(test.configured)}
			updateClusterCacheRegistryState(&state, CacheRegistryValues{Policy: &test.fromAPI})
			if got := state.Policy.ValueString(); got != test.want {
				t.Errorf("policy = %q, want %q", got, test.want)
			}
		})
	}
}
```

Worth adding the documented example's shape from `examples/resources/buildkite_cluster_cache_registry/resource.tf` as a fourth case, since that is the config we are telling people to write and it is the one that exercises both `restore.scopes` entries and scalar actions together.

petetomasik (changes requested):
2 more findings in addition to the 1 blocker from buildsworth.

buildkite/resource_cluster_cache_registry.go
blaknite: Good catch. I scoped not-found handling to the delete mutation so organization lookup failures now propagate, and added a Relay-based `CheckDestroy` in 6a7402b.

docs/resources/cluster_cache_registry.md
```diff
@@ -0,0 +1,82 @@
...
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "buildkite_cluster_cache_registry Resource - terraform-provider-buildkite"
+subcategory: ""
+description: |-
+  A Cache Registry stores cached build data for a Buildkite Cluster.
+---
+
+# buildkite_cluster_cache_registry (Resource)
+
+A Cache Registry stores cached build data for a Buildkite Cluster.
```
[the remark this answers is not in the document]
blaknite: Agreed. I added the private-preview and organization-enablement requirement to the resource description and regenerated the provider docs in 74b8ce7.

buildkite/resource_cluster_cache_registry.go
blaknite: Fixed in 7f06da7. The state-mapping function now has direct coverage for expanded defaults, widened actions, genuine policy drift, and the documented example shape, so replacing the preservation branch with unconditional assignment fails the unit suite.

buildkite/resource_cluster_cache_registry.go
blaknite: Fixed in 7f06da7. Delete idempotency now accepts only structured GraphQL errors matching the registry-specific “No cache registry found” message. Organization-not-found, plain HTTP/string errors, and permission errors propagate. Focused unit coverage exercises those distinctions.

buildkite/resource_cluster_cache_registry.go
```diff
@@ -0,0 +1,377 @@
...
+		result, err = getCacheRegistryByNode(ctx, r.client.genqlient, state.ID.ValueString())
+		return retryContextError(err)
+	})
+	if err != nil {
+		resp.Diagnostics.AddError("Unable to read Cache Registry", fmt.Sprintf("Unable to read Cache Registry: %s", err))
+		return
+	}
+
+	cacheRegistry, ok := result.Node.(*getCacheRegistryByNodeNodeCacheRegistry)
+	if !ok || cacheRegistry == nil {
```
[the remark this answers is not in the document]
blaknite: Fixed in 7f06da7. A null Relay lookup now probes the owning cluster cache-registry connection before removing state. Feature, permission, organization, cluster, and connection failures return diagnostics; state is removed only after a successful connection scan confirms absence.

buildkite/resource_cluster_cache_registry.go
```diff
@@ -0,0 +1,438 @@
...
+func updateClusterCacheRegistryState(state *clusterCacheRegistryResourceModel, result CacheRegistryValues) {
+	state.ID = types.StringValue(result.Id)
+	state.UUID = types.StringValue(result.Uuid)
+	state.ClusterID = types.StringValue(result.CacheRegistryCluster.Id)
+	state.ClusterUUID = types.StringValue(result.CacheRegistryCluster.Uuid)
+	state.Name = types.StringValue(result.Name)
+	state.Slug = types.StringValue(result.Slug)
+	state.Description = types.StringPointerValue(result.Description)
+	state.Emoji = types.StringPointerValue(result.Emoji)
+	state.Color = types.StringPointerValue(result.Color)
```
petetomasik, line 548: These three attributes are `Optional` without `Computed` (L103-114), so the planned value is the config value verbatim and is always known. `updateClusterCacheRegistryState` then writes the API response over it in both Create (L169) and Update (L246). Any server-side normalization makes the applied state diverge from the plan, and Terraform treats that as a hard error rather than a diff:

```
Error: Provider produced inconsistent result after apply
.color: was cty.StringVal(""), but now null
```

This isn't speculative in this repo. `resource_test_suite.go:70-74` documents the server behaviour: "the server coerces blank values (eg color) to null, and `""` is the documented way to clear these attributes from configuration". That resource was hardened for exactly this failure, on exactly `color`: it declares `color` as Optional+Computed (`resource_test_suite.go:379-384`) and routes it through `stateStringValue` on create/update (L170, L451) and `refreshedStringValue` on read (L314).

Two triggers:

1. `color = ""` (equally `description = ""` or `emoji = ""`). `optionalStringPayload` sends `""`, the server returns null, state goes null against a known plan value of `""`, and the apply fails. Terraform still persists the returned state, so the next plan re-proposes `""` and fails again until someone hand-edits state.
2. Hex case folding. Both `examples/resources/buildkite_cluster_cache_registry/resource.tf:11` and the acceptance test use `#BADA55`. If the resolver lowercases it, same failure. Emoji shortcode canonicalisation would do the same.

Read has the mirror of this: with `description = ""` in config, Read sets state to null and you get a perpetual non-empty plan.

Suggested fix: make `description`/`emoji`/`color` Optional+Computed and use the existing `stateStringValue` / `refreshedStringValue` helpers, or follow `resource_cluster.go:138-141` and leave planned values untouched on create/update, taking API values only in Read.

Worth flagging that `resource_registry.go:455-457` has the same unfixed shape, so the repo currently holds two conflicting patterns. `resource_test_suite.go` is the one that was actually fixed for this failure class.

buildkite/resource_cluster_cache_registry.go
```diff
@@ -0,0 +1,438 @@
...
+	if json.Unmarshal([]byte(left), &leftPolicy) != nil || json.Unmarshal([]byte(right), &rightPolicy) != nil {
+		return false
+	}
+
+	normalize := func(policy map[string]any) {
+		for section, scopes := range map[string]any{"save": map[string]any{}, "restore": []any{}} {
+			value, ok := policy[section].(map[string]any)
+			if !ok {
+				value = map[string]any{}
+				policy[section] = value
```
petetomasik: `normalize` writes into `policy` without checking that the map is non-nil, so a JSON `null` on either side panics the provider. Reproduced against the real function:

```go
cacheRegistryPoliciesEquivalent(`{"rules":[]}`, `null`)
// panic: assignment to entry in nil map
//   resource_cluster_cache_registry.go:388

cacheRegistryPoliciesEquivalent(`null`, `{}`)
// panic: assignment to entry in nil map
```

The cause is that `json.Unmarshal([]byte("null"), &m)` returns a **nil map with no error**, so both `!= nil` guards at L379 pass and L388 then assigns into a nil map.

How a `"null"` reaches here: `genqlient.yaml` binds the `JSON` scalar to Go `string`, so `CacheRegistryValues.Policy` is a `*string` holding a serialized JSON document. The guard at L348 (`result.Policy == nil`) catches a GraphQL `null`, but it does not catch the resolver returning the four-character string `"null"` as the encoded scalar, which is what Rails `nil.to_json` produces for an absent policy.

The blast radius is worse than a single failed operation. Once that value lands in state via L349, `state.Policy.ValueString()` is `"null"` on every subsequent Read, so the left-hand side panics too and the user cannot plan their way out. The config side is safe (the validator rejects `"null"` at L430-437), so both panic paths route through the API response.

I could not verify what the resolver actually returns for an absent policy, since the API isn't deployed. One create-without-policy against the local instance settles it.

Suggested fix, independent of that answer:

```go
if leftPolicy == nil || rightPolicy == nil {
    return leftPolicy == nil && rightPolicy == nil
}
```

after the unmarshal at L379, or have `normalize` return early on a nil map. Worth adding table cases for `("{}", "null")`, `("null", "{}")`, `("{oops", "{}")` and `("{}", "[]")` at the same time.

buildkite/resource_cluster_cache_registry.go
```diff
@@ -0,0 +1,438 @@
...
+	state.ClusterID = types.StringValue(result.CacheRegistryCluster.Id)
+	state.ClusterUUID = types.StringValue(result.CacheRegistryCluster.Uuid)
+	state.Name = types.StringValue(result.Name)
+	state.Slug = types.StringValue(result.Slug)
+	state.Description = types.StringPointerValue(result.Description)
+	state.Emoji = types.StringPointerValue(result.Emoji)
+	state.Color = types.StringPointerValue(result.Color)
+	if state.Policy.IsNull() || state.Policy.IsUnknown() || result.Policy == nil || !cacheRegistryPoliciesEquivalent(state.Policy.ValueString(), *result.Policy) {
+		state.Policy = jsontypes.NewNormalizedPointerValue(result.Policy)
+	}
```
petetomasik, line 551: Taking the API value when it isn't equivalent is right for Read, where it surfaces drift. In Create and Update it's wrong by construction, and it makes the resource fragile to any additive change on the server.

`policy` is Optional+Computed with `UseStateForUnknown`, so on Create-with-policy and on every subsequent plan the planned value is known: either the config string, or the prior state value carried forward by the Optional+Computed rule. When `cacheRegistryPoliciesEquivalent` returns false, L349 writes a different value over that known plan. The framework's own semantic equality only reverts to the planned value when the two strings are JSON-equal, which by construction they are not here, so Terraform core fails the apply with its generic inconsistency error and the provider emits no diagnostic explaining why.

The deeper issue is what `normalize` knows. It encodes one exact snapshot of the current server response shape: three known top-level keys, `save.scopes` defaulting to `{}` but `restore.scopes` to `[]` (L384), `rules` to `[]` (L394-396), and `action` widened only when it's a `string` (L403). That list has no link back to the server. Probed against the real function:

| config | API response | equivalent |
|---|---|---|
| `{"save":{"scopes":{"branch":true}}}` | same plus `"restore":{"scopes":[]},"rules":[],"version":1` | **false** |
| `{}` | `{"save":{"scopes":{}},"restore":{"scopes":{}},"rules":[]}` | **false** |

So the day the API adds a top-level key (`version`, `enabled`, `ttl`), or returns `restore.scopes` as an object, every apply with a configured `policy` breaks for every user of the provider. The unit case at `resource_cluster_cache_registry_test.go:128-132` ("expanded defaults keep the configured text") is exactly the assertion that inverts, and nothing in the repo catches it because the new field exists in no fixture.

Two paths reach this today rather than at some future GA:

- `policy: JSON` is nullable in `schema.graphql:2021`, so a null response against a configured policy hits L349 and writes null over a known non-null plan.
- The validator at L425-438 accepts any JSON object, so `policy = jsonencode({ retention = { max_age_days = 30 } })` passes and reaches the resolver. If unrecognised keys are dropped from the normalized policy, the apply hard-fails.

Since this is private preview, the cleanest fix is upstream: have the resolver return a canonical policy so provider-side default injection isn't needed. That window closes at GA. Failing that, stop requiring full-shape agreement here: compare only the keys the configured policy actually sets, or accept the configured value in Create/Update and let Read own drift detection. If the strict comparison stays, please add an explicit diagnostic naming the returned policy, so the failure is legible instead of surfacing as core's inconsistency check.

Either way, a comment at L383 recording that the default table mirrors the server and must be updated when the policy schema changes would give the next person hitting this a pointer.

buildkite/resource_cluster_cache_registry.go
```diff
@@ -0,0 +1,438 @@
...
+		})
+		if err != nil {
+			return false, err
+		}
+		if result.Organization == nil {
+			return false, fmt.Errorf("organization %q was not found or is inaccessible", r.client.organization)
+		}
+		if result.Organization.Cluster == nil {
+			return false, fmt.Errorf("cluster %q was not found or is inaccessible", state.ClusterUUID.ValueString())
+		}
```
petetomasik: This branch turns a recoverable situation into one that blocks every plan until the user reaches for `terraform state rm`.

Trigger: create a `buildkite_cluster` plus a cache registry in it, then delete the cluster in the Buildkite UI, which necessarily destroys its cache registries. On the next `terraform plan`:

1. `getCacheRegistryByNode` returns `node: null`, so the type assertion at L198 fails.
2. The fallback `cacheRegistryExistsInCluster` runs with the state's `cluster_uuid`.
3. `result.Organization.Cluster == nil`, so this line returns `cluster %q was not found or is inaccessible`.
4. Read turns that into `resp.Diagnostics.AddError` at L202, which fails refresh and therefore the whole plan.

There's no way out with normal Terraform operations, even though the correct answer is unambiguous: if the cluster doesn't exist, a registry inside it can't exist either. The same dead end occurs if the Cache private preview flag is turned off for an org that already has resources, via the `CacheRegistries == nil` branch at L301-303.

This is also a deviation from the repo convention. `resource_cluster.go:178-193` removes the resource from state whenever the node assertion fails, full stop, and no other resource here second-guesses a null node.

Suggested fix: treat `Organization.Cluster == nil` as "the registry is gone" and fall through to `resp.State.RemoveResource(ctx)`. Keeping `Organization == nil` as an error is defensible, since that genuinely indicates a credentials or org-slug problem rather than a deleted resource. If any of these stay as errors, the diagnostic should at least tell the user to run `terraform state rm`.

Coverage note: the acceptance test's out-of-band step at L365-369 only deletes the registry, so the cluster-deleted path has no coverage at all.

petetomasik (changes requested):

buildkite/resource_cluster_cache_registry.go
```diff
@@ -0,0 +1,672 @@
...
+	state.Emoji = types.StringPointerValue(result.Emoji)
+	state.Color = types.StringPointerValue(result.Color)
+	if state.Policy.IsNull() || state.Policy.IsUnknown() || result.Policy == nil || !cacheRegistryPoliciesEquivalent(state.Policy.ValueString(), *result.Policy) {
+		state.Policy = jsontypes.NewNormalizedPointerValue(result.Policy)
+	}
+	state.CreatedAt = types.StringValue(result.CreatedAt.Format(time.RFC3339))
+	state.UpdatedAt = types.StringValue(result.UpdatedAt.Format(time.RFC3339))
+	if result.Policy == nil {
+		return fmt.Errorf("cache registry %q returned a null policy; expected a JSON object. The remote identity has been recorded in state. Check the API response and run terraform plan before applying again", result.Id)
+	}
```
petetomasik: A null policy is a supported state upstream, but this branch treats it as an invalid response on every path, Read included. A registry whose policy has been cleared can't be refreshed or imported. So the chain is:

1. Someone clears the policy through the API, another client, or a script.
2. On the next `terraform plan`, Read gets `policy: null` and reaches this branch.
3. Read returns an error diagnostic, so refresh fails and so does the whole plan.
4. The message says to "run terraform plan before applying again", which just hits step 3 again. Read fails before Update can run, so the user can't fix it by setting `policy` in config. The only way out is `terraform state rm`.

`terraform import` of such a registry fails the same way.

The rest of the resource already handles null. L549 stores it as a null state value, and with `policy` omitted from config, `UseStateForUnknown` leaves the plan unknown, so Update resends `policy: null`, which is a no-op upstream. If the user does configure a policy, the next plan shows an in-place update that restores it. So I think this only needs the error dropped:

```suggestion
	if result.Policy == nil {
		return nil
	}
```

These cases assert the current behavior and would need to flip: "API null" in `TestClusterCacheRegistryReadPolicy` and `TestClusterCacheRegistryApplyResponse`, the "null" case in `TestUnitClusterCacheRegistryApplyErrorPreservesIdentity`, and "policy null" in the partial-mutation test. An import-of-a-null-policy-registry case in the fake would be worth adding too.

The planned-vs-returned check in `updateClusterCacheRegistryAppliedState` (L530) still catches a configured policy coming back null on Create/Update, and that's the right place for it.

petetomasik (approved):

PR #1259, opened 2026-09-04
fix(pipeline): no perpetual diff for unset issue/review comment match…

dattallant:
With `provider_settings` configured but the two match modes left out, every plan after a create or update reports

```
Note: Objects have changed outside of Terraform
    - issue_comment_match_mode  = "" -> null
    - review_comment_match_mode = "" -> null
```

and anything referencing `provider_settings` (an output, a module) sees a change on every plan until state is rewritten. The REST create/update responses return `""` for an unset match mode while the GraphQL read returns null; the REST side now reads `""` as null too. Configured values are untouched (the validator never allowed `""`). The "" has been written since the match modes were added in #1156 (v1.35.0), so existing state files carry it; to keep a -refresh=false apply consistent with that state, the two attributes now use a variant of UseNonNullStateForUnknown that treats "" in state like null and leaves the plan unknown when an update is pending.

Added a unit test and an acceptance test (provider_settings without the match modes plus an output, relying on the empty follow-up plan) that fails on main and passes here; ran it locally against a test organization.

## PR checklist:
- [x] tests added
- [ ] an example added to `examples/` <-- bug fix

### DIFF
--- a/buildkite/resource_pipeline.go
+++ b/buildkite/resource_pipeline.go
@@ -1810,10 +1810,10 @@ func updatePipelineResourceExtraInfo(state *pipelineResourceModel, pipeline *Pip
 		UseMergeGroupBaseCommitForGitDiffBase:   types.BoolPointerValue(s.UseMergeGroupBaseCommitForGitDiffBase),
 		BuildIssueCommentCreated:                types.BoolPointerValue(s.BuildIssueCommentCreated),
 		IssueCommentCommandWord:                 types.StringPointerValue(s.IssueCommentCommandWord),
-		IssueCommentMatchMode:                   types.StringPointerValue(s.IssueCommentMatchMode),
+		IssueCommentMatchMode:                   matchModeFromREST(s.IssueCommentMatchMode),
 		BuildPullRequestReviewCommentCreated:    types.BoolPointerValue(s.BuildPullRequestReviewCommentCreated),
 		ReviewCommentCommandWord:                types.StringPointerValue(s.ReviewCommentCommandWord),
-		ReviewCommentMatchMode:                  types.StringPointerValue(s.ReviewCommentMatchMode),
+		ReviewCommentMatchMode:                  matchModeFromREST(s.ReviewCommentMatchMode),
 		BuildPullRequestDequeued:                types.BoolPointerValue(s.BuildPullRequestDequeued),
 		BuildPullRequestReopened:                types.BoolPointerValue(s.BuildPullRequestReopened),
 		BuildCheckRunCompleted:                  types.BoolPointerValue(s.BuildCheckRunCompleted),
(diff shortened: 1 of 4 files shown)


buildkite/resource_pipeline.go
```diff
@@ -1829,6 +1829,14 @@ func updatePipelineResourceExtraInfo(state *pipelineResourceModel, pipeline *Pip
 	}
 }
 
+// matchModeFromREST reads an unset match mode, which the REST API returns as "", as null so it agrees with the GraphQL read
+func matchModeFromREST(m *string) types.String {
+	if m == nil || *m == "" {
+		return types.StringNull()
```
petetomasik, line 1835: This turns the cosmetic drift into a hard apply failure for anyone carrying `""` in state from a released version.

Trigger: state holds `""` for this attribute (any apply on v1.35.0 through v1.39.1 with `provider_settings` set and the match modes omitted), the config still omits it, and the user runs `terraform apply -refresh=false`, or applies a plan file produced with refresh off, with any change that updates the pipeline.

Value trace:

1. Terraform core's ProposedNew carries the prior `""` into the proposed new state.
2. `MarkComputedNilsAsUnknown` (framework v1.19.0, `internal/fwserver/server_planresourcechange.go:414`) rewrites it to unknown. It is named for the *config* value being nil, not the planned value: its only guards are config-null, Computed, and no `Default`, all of which these attributes satisfy. It never checks whether the value it overwrites is null.
3. `UseNonNullStateForUnknown` then sees an unknown plan value and a non-null prior state, so it copies the `""` back. The planned value is now a known `""` rather than unknown.
4. Update (`resource_pipeline.go:1471`) runs, this function maps the REST `""` to null, and `resp.State.Set` writes null.

Planned known `""` against applied null gives `Error: Provider produced inconsistent result after apply ... .provider_settings.issue_comment_match_mode: was cty.StringVal(""), but now null`. Framework providers get no legacy type-system downgrade, so this fails the apply rather than logging a warning.

On `main`, step 4 returns `""`, so the plan and the apply agree and nothing errors; this change is what makes them disagree. Under the default refreshing workflow it cannot fire, because refresh writes null into prior state first, step 3 then bails out on a null state value, the plan stays unknown, and unknown accepts null. That is also why the new acceptance test cannot catch it: the test framework always refreshes.

The precondition is real rather than hypothetical. `resource_pipeline_test.go:807` asserts `""` in state from this same code path and passes in CI, and the GraphQL read that produces the null side shipped in #1156 (`7f2a102`), which is contained in v1.35.0 through v1.39.1. Related: the description says this was introduced in #1236 and is not yet released, but #1236 (`c72d9d6`) only added the `build_issues` bool, so that attribution does not hold, and it is what makes this precondition look impossible.

Cheapest fix is a plan modifier on both attributes that forces the plan back to unknown when the prior state is `""` and the config is null. It has to be appended after `UseNonNullStateForUnknown`, since modifiers run in order and that one will re-copy the `""` otherwise. A schema version bump plus an upgrader rewriting `""` to null also works and is more thorough. Accepting it is defensible given the narrow trigger, but then the release note should say that a refresh is required after upgrading and that one `-refresh=false` apply can fail.

petetomasik (changes requested):
Thanks for the contribution @dattallant ! Found a scenario where a previous provider version's state was already configured would cause a provider error.

buildkite/resource_pipeline.go
dattallant: Thanks for the trace. The "" has been written since #1156 (v1.35.0), not #1236; I've corrected the description. e7576b8 swaps UseNonNullStateForUnknown on the two match modes for a small variant that also treats "" in state like null, so with legacy state and -refresh=false a pending update plans them as unknown and the null from the apply is accepted, while a plan with no other change still copies nothing and stays empty. I reproduced your failure with a saved non-refreshed plan against state seeded by v1.39, and the same plan applies cleanly with this commit. The modifier is unit-tested directly since the acceptance framework always refreshes.

Happy to do the schema version plus upgrader route instead if you'd rather have the "" rewritten in state.

petetomasik (approved):

PR #1263, opened 2026-09-07
Document what happens when an apply fails partway

petetomasik:
Several resources take more than one API call. Nothing user-facing said so, or explained that a create which fails partway leaves the resource tainted and that terraform untaint is how you finish the configuration in place rather than replacing a resource that already works.

The section needs a provider template, which the repo did not have, so `docs/index.md`was coming from the built-in one.


petetomasik:
@buildsworth-bk review this PR and approved up to L2

PR #1268, opened 2026-09-10
fix(team_member): replace the membership when team_id or user_id changes

dattallant:
Changing `team_id` or `user_id` on a `buildkite_team_member` planned an in-place update that then failed, and kept failing on every run:

```
~ team_id = "VGVhbS0tLWM1…" -> "VGVhbS0tLWI3…"
Error: Provider produced inconsistent result after apply … .team_id: was cty.StringVal("VGVhbS0tLWI3…"), but now cty.StringVal("VGVhbS0tLWM1…"
```

`TeamMemberUpdateInput` only carries the role, so the update can't move a membership; both attributes now require replacement, like `buildkite_pipeline_team` and `buildkite_cluster_maintainer` already do. The sibling `buildkite_test_suite_team` had the same gap for `test_suite_id`/`team_id` (its update only carries the access level) and gets the same treatment. The team member example also used a `TeamMember` ID where a `User` ID belongs.

Added an acceptance test per resource that moves the membership to another team and expects a replace, and ran them locally against a test organization.

## PR checklist:
- [x] `docs/` updated
- [x] tests added
- [x] an example added to `examples/` <-- example corrected


scadu (approved):
Looks good! Thanks for contributing @dattallant!

PR #1278, opened 2026-09-21
feat(organization): manage user memberships

### LINKED ISSUE #462: Feature request for user organization membership
Hello,

I see that this provider supports managing Buildkite Teams and Buildkite user memberships to those teams.  ie:

https://github.com/buildkite/terraform-provider-buildkite/blob/main/docs/resources/team_member.md

However, I don't see a resource for managing user membership to the Buildkite organization? This would be analogous to this resource in the github provider

https://registry.terraform.io/providers/integrations/github/latest/docs/resources/membership

Would this be possible? It would greatly simplify our onboarding/offboarding of users.

dahtey-bk:
## Why

The provider can manage teams and team memberships, but not users' membership of the organization itself. Customers need that missing piece to manage onboarding and offboarding through Terraform.

## What

Adds `buildkite_organization_membership` using the REST API. It adopts existing memberships, optionally invites absent users, tracks pending invitations through acceptance, and manages organization role and SSO mode.

```hcl
resource "buildkite_organization_membership" "jane" {
  email           = "jane@example.com"
  role            = "MEMBER"
  sso_mode        = "REQUIRED"
  send_invitation = true
}
```

Destroy revokes pending invitations or removes active members by default. The opt-in `downgrade_on_destroy` flag instead retains active users as members, leaving their SSO mode unchanged.

Once active, the resource exposes `user_id` for `buildkite_team_member`. Team management stays separate, invitation resend is out of scope, and the documentation warns against SCIM and Terraform both owning the same membership lifecycle. Terraform does not wait for invitation acceptance; a later refresh discovers it.

Local tests, race checks, vet, lint, and build passed. Terraform lifecycle tests ran against a local REST fixture; live Buildkite validation has been completed using my own BK sandbox org.

## Context

- Closes #462
- [Linear: SUP-1677](https://linear.app/buildkite/issue/SUP-1677/feature-request-for-user-organization-membership)
- [Internal Slack scope discussion](https://buildkite-corp.slack.com/archives/C064NER3L21/p1787653045042339)

## PR checklist

- [x] `docs/` updated
- [x] Tests added
- [x] Examples added to `examples/`
- [ ] `CHANGELOG.md` updated with pending release information — not included in this PR


buildkite/resource_organization_membership.go
```diff
@@ -0,0 +1,430 @@
...
+		`),
+		Attributes: map[string]schema.Attribute{
+			"id": schema.StringAttribute{
+				Computed: true, MarkdownDescription: "Stable resource identifier, initially the member's user UUID or invitation UUID. Does not change on acceptance.",
+				PlanModifiers: []planmodifier.String{stringplanmodifier.UseStateForUnknown()},
+			},
+			"email": schema.StringAttribute{
+				Optional: true, Computed: true, MarkdownDescription: "Email address used to adopt or invite. Specify email or uuid. Required to invite an absent user. Retained after adoption even if the user's primary email changes. Changing a configured email replaces the resource.",
+				Validators:    []validator.String{stringvalidator.LengthAtLeast(1)},
+				PlanModifiers: []planmodifier.String{stringplanmodifier.RequiresReplaceIfConfigured()},
```
petetomasik: Importing an existing membership can produce a destructive replacement solely because the configured email differs from the API email. ImportState records the user UUID, then Read fills email from the primary email; this modifier consequently plans destroy-before-create for both case-only differences and an invited-address/primary-address mismatch. I reproduced both cases with Terraform. Pending-invitation imports have the same case-only problem.

With the default destroy behavior, applying that plan removes the existing member (risking loss of team memberships and tokens), or revokes the pending invitation. If send_invitation is false, recreation then fails.

Could we reconcile imported email differences without replacing the membership/invitation, and add import-followed-by-plan regression tests for these cases? Skipping replacement alone may still leave an in-place diff, so the tests should establish the intended reconciliation behavior rather than assume the first plan will be empty.

buildkite/resource_organization_membership.go
```diff
@@ -0,0 +1,430 @@
...
+	if id := state.InvitationID.ValueString(); id != "" {
+		invitation, err := r.client.getOrganizationMembershipInvitation(ctx, id)
+		if err != nil && !errors.Is(err, errOrganizationMembershipNotFound) {
+			return nil, nil, err
+		}
+		if invitation != nil && invitation.AcceptedBy != nil && invitation.AcceptedBy.ID != "" {
+			member, err := r.client.getOrganizationMembershipMember(ctx, invitation.AcceptedBy.ID)
+			return member, nil, err
+		}
+		member, err := r.client.findOrganizationMembershipMember(ctx, state.Email.ValueString())
```
petetomasik: Could we return a confirmed pending invitation before falling back to the member lookup here? Every refresh currently scans the entire organization membership when the invitee is still absent, even though the invitation GET already tells us it is pending.

For 50 pending invitations in a 5,000-member organization, one refresh makes 2,600 REST requests: 50 invitation GETs plus 51 member-list pages per resource (including the final empty page). Update and destroy repeat the lookup when invoked. This is a substantial rate-limit risk for normal onboarding workloads.

Please keep the direct accepted_by UUID lookup, return immediately for pending invitations, and reserve the email fallback for the other states or a missing invitation. A regression test should assert zero member-list requests for a pending refresh while retaining acceptance-race coverage. If scanning a pending invitation protects a reachable SSO/SCIM lifecycle case, that behavior needs verification and a focused test before choosing a different approach.

petetomasik (changes requested):

